PR #294, opened 2014-04-28
replace extend with underscore 

161chihuahuas:
use underscore's version of `extend()` from within PeerManager


nerdfiles:
Considering the critical nature and significance of this project: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

PR #884, opened 2015-01-02
Fixed deterministicK() to match python-ecdsa

bip32JP:
I added a similar badrs function to python-ecdsa and compared the results.

The 1 badrs (aka forcing it to loop once) gave me a different value. It turns out you missed one of the `v = hmac_k(v)` steps during the loop.

Adding one extra `v = hmac_k(v)` in each loop makes it match up with python-ecdsa perfectly (I even tried up to badrs = 30 and it was fine.

https://tools.ietf.org/html/rfc6979#section-3.2

If you follow Step h from after the `k = bits2int(T)` downward, then back up to the beginning of Step h to loop.

```
K = HMAC_K(V || 0x00)
V = HMAC_K(V)
Empty T
V = HMAC_K(V)
T = V    (since we know tlen == qlen)
k = bits2int(T)
```

As you can see, the original code in Bitcore was missing one `V = HMAC_K(V)` whereas python-ecdsa includes it in the loop.

Including this commit will make the behavior be 100% rfc 6979 compliant.

(The only way this could be a security issue is if someone had a transaction + privkey pair that just happened to produce an out of bounds k or bad r or s... AND they decided to sign that transaction using both bitcore AND Electrum (uses python-ecdsa) and then for some reason posted those two signed transactions in the public space somehow.
I understand there are infinitecimally close to 0 issues in security by missing one hmac in the error loop, but just so we're all on the same page.)

### DIFF
--- a/lib/crypto/ecdsa.js
+++ b/lib/crypto/ecdsa.js
@@ -100,6 +100,7 @@ ECDSA.prototype.deterministicK = function(badrs) {
   for (var i = 0; i < badrs || !(T.lt(N) && T.gt(0)); i++) {
     k = Hash.sha256hmac(Buffer.concat([v, new Buffer([0x00])]), k);
     v = Hash.sha256hmac(v, k);
+    v = Hash.sha256hmac(v, k);
     T = BN.fromBuffer(v);
   }
 
--- a/test/crypto/ecdsa.js
+++ b/test/crypto/ecdsa.js
@@ -103,7 +103,7 @@ describe('ECDSA', function() {
       ecdsa.k.toBuffer().toString('hex')
         .should.not.equal('fcce1de7a9bcd6b2d3defade6afa1913fb9229e3b7ddf4749b55c4848b2a196e');
       ecdsa.k.toBuffer().toString('hex')
-        .should.equal('6f4dcca6fa7a137ae9d110311905013b3c053c732ad18611ec2752bb3dcef9d8');
+        .should.equal('727fbcb59eb48b1d7d46f95a04991fc512eb9dbf9105628e3aec87428df28fd8');
     });
     it('should compute this test vector correctly', function() {
       // test fixture from bitcoinjs


bip32JP:
I also made some test vectors using python-ecdsa:
k_bad00 is badrs = 0
k_bad01 is badrs = 1
k_bad15 is badrs = 15

```
[
	{
		'message': 'test data',
		'privkey': 'fee0a1f7afebf9d2a5a80c0c98a31c709681cce195cbcd06342b517970c0be1e',
		'k_bad00': 'fcce1de7a9bcd6b2d3defade6afa1913fb9229e3b7ddf4749b55c4848b2a196e',
		'k_bad01': '727fbcb59eb48b1d7d46f95a04991fc512eb9dbf9105628e3aec87428df28fd8',
		'k_bad15': '398f0e2c9f79728f7b3d84d447ac3a86d8b2083c8f234a0ffa9c4043d68bd258'
	},
	{
		'message': 'Everything should be made as simple as possible, but not simpler.',
		'privkey': '0000000000000000000000000000000000000000000000000000000000000001',
		'k_bad00': 'ec633bd56a5774a0940cb97e27a9e4e51dc94af737596a0c5cbb3d30332d92a5',
		'k_bad01': 'df55b6d1b5c48184622b0ead41a0e02bfa5ac3ebdb4c34701454e80aabf36f56',
		'k_bad15': 'def007a9a3c2f7c769c75da9d47f2af84075af95cadd1407393dc1e26086ef87'
	},
	{
		'message': 'Satoshi Nakamoto',
		'privkey': '0000000000000000000000000000000000000000000000000000000000000002',
		'k_bad00': 'd3edc1b8224e953f6ee05c8bbf7ae228f461030e47caf97cde91430b4607405e',
		'k_bad01': 'f86d8e43c09a6a83953f0ab6d0af59fb7446b4660119902e9967067596b58374',
		'k_bad15': '241d1f57d6cfd2f73b1ada7907b199951f95ef5ad362b13aed84009656e0254a'
	},
	{
		'message': 'Diffie Hellman',
		'privkey': '7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f7f',
		'k_bad00': 'c378a41cb17dce12340788dd3503635f54f894c306d52f6e9bc4b8f18d27afcc',
		'k_bad01': '90756c96fef41152ac9abe08819c4e95f16da2af472880192c69a2b7bac29114',
		'k_bad15': '7b3f53300ab0ccd0f698f4d67db87c44cf3e9e513d9df61137256652b2e94e7c'
	},
	{
		'message': 'Japan',
		'privkey': '8080808080808080808080808080808080808080808080808080808080808080',
		'k_bad00': 'f471e61b51d2d8db78f3dae19d973616f57cdc54caaa81c269394b8c34edcf59',
		'k_bad01': '6819d85b9730acc876fdf59e162bf309e9f63dd35550edf20869d23c2f3e6d17',
		'k_bad15': 'd8e8bae3ee330a198d1f5e00ad7c5f9ed7c24c357c0a004322abca5d9cd17847'
	},
	{
		'message': 'Bitcoin',
		'privkey': 'fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140',
		'k_bad00': '36c848ffb2cbecc5422c33a994955b807665317c1ce2a0f59c689321aaa631cc',
		'k_bad01': '4ed8de1ec952a4f5b3bd79d1ff96446bcd45cabb00fc6ca127183e14671bcb85',
		'k_bad15': '56b6f47babc1662c011d3b1f93aa51a6e9b5f6512e9f2e16821a238d450a31f8'
	},
	{
		'message': 'i2FLPP8WEus5WPjpoHwheXOMSobUJVaZM1JPMQZq',
		'privkey': 'fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140',
		'k_bad00': '6e9b434fcc6bbb081a0463c094356b47d62d7efae7da9c518ed7bac23f4e2ed6',
		'k_bad01': 'ae5323ae338d6117ce8520a43b92eacd2ea1312ae514d53d8e34010154c593bb',
		'k_bad15': '3eaa1b61d1b8ab2f1ca71219c399f2b8b3defa624719f1e96fe3957628c2c4ea'
	},
	{
		'message': 'lEE55EJNP7aLrMtjkeJKKux4Yg0E8E1SAJnWTCEh',
		'privkey': '3881e5286abc580bb6139fe8e83d7c8271c6fe5e5c2d640c1f0ed0e1ee37edc9',
		'k_bad00': '5b606665a16da29cc1c5411d744ab554640479dd8abd3c04ff23bd6b302e7034',
		'k_bad01': 'f8b25263152c042807c992eacd2ac2cc5790d1e9957c394f77ea368e3d9923bd',
		'k_bad15': 'ea624578f7e7964ac1d84adb5b5087dd14f0ee78b49072aa19051cc15dab6f33'
	},
	{
		'message': '2SaVPvhxkAPrayIVKcsoQO5DKA8Uv5X/esZFlf+y',
		'privkey': '7259dff07922de7f9c4c5720d68c9745e230b32508c497dd24cb95ef18856631',
		'k_bad00': '3ab6c19ab5d3aea6aa0c6da37516b1d6e28e3985019b3adb388714e8f536686b',
		'k_bad01': '19af21b05004b0ce9cdca82458a371a9d2cf0dc35a813108c557b551c08eb52e',
		'k_bad15': '117a32665fca1b7137a91c4739ac5719fec0cf2e146f40f8e7c21b45a07ebc6a'
	},
	{
		'message': '00A0OwO2THi7j5Z/jp0FmN6nn7N/DQd6eBnCS+/b',
		'privkey': '0d6ea45d62b334777d6995052965c795a4f8506044b4fd7dc59c15656a28f7aa',
		'k_bad00': '79487de0c8799158294d94c0eb92ee4b567e4dc7ca18addc86e49d31ce1d2db6',
		'k_bad01': '9561d2401164a48a8f600882753b3105ebdd35e2358f4f808c4f549c91490009',
		'k_bad15': 'b0d273634129ff4dbdf0df317d4062a1dbc58818f88878ffdb4ec511c77976c0'
	}
]
```

eordano:
LGTM, somebody else please review and merge @maraoz @yemel @braydonf

eordano:
Also, ping @ryanxcharles for backporting into `fullnode`: https://github.com/ryanxcharles/fullnode/blob/master/lib/ecdsa.js#L122

yemel:
Good catch! Thanks @bip32JP for your contribution.

ryanxcharles:
Thanks! Will include the change in fullnode.

ryanxcharles:
I'm not sure this change is correct. The extra HMAC occurs under every computation of deterministic k, not just when "badrs" is triggered. Are we absolutely sure that python-ecdsa had the correct implementation of RFC 6979? Perhaps python-ecdsa was erroneously computing an extra HMAC.

braydonf:
Still looking into it here, but not 100% certain. Would be good to have more tests around this that illustrate the point.

ryanxcharles:
Part of the problem is that it is difficult to test the case badrs > 0, since it only occurs in rare circumstances. We need a test case for that condition. This is probably why bitcoinjs-lib throws an exception there rather than looping - they probably didn't have a test case and therefore couldn't be sure it was correct.

ryanxcharles:
See my above referenced PR to fullnode. I think that may be the correct way to solve this. After re-reading RFC 6979, I think @bip32JP is probably correct - there should be an extra HMAC computation. But the extra HMAC computation should only happen with badrs > 0. Unfortunately I don't have a test case for this. How do we trigger badrs > 0? i.e., what we need to do is find a value of k that produces a bad signature.

bip32JP:
It is very clear in the RFC.

I will break down the entire steps to their basic pseudo code.

https://tools.ietf.org/html/rfc6979#section-3.2
```
h1 = H(m)

V = 0x01 0x01 0x01 ... 0x01

K = 0x00 0x00 0x00 ... 0x00

K = HMAC_K(V || 0x00 || int2octets(x) || bits2octets(h1))

V = HMAC_K(V)

K = HMAC_K(V || 0x01 || int2octets(x) || bits2octets(h1))

V = HMAC_K(V)

       // Step H starts:
       // Notice H-1 says "Set T to empty sequence"... this is "generate new T" RFC speaks of.
       // this RFC wasn't necessarily made for SHA256 + SECP256k1
       // (where hlen and qlen are always equal)

1.  Set T to the empty sequence.  The length of T (in bits) is
    denoted tlen; thus, at that point, tlen = 0.

2.  While tlen < qlen, do the following:

    V = HMAC_K(V)

    T = T || V

    k = bits2int(T)

    If that value of k is within the [1,q-1] range, and is
    suitable for DSA or ECDSA (i.e., it results in an r value
    that is not 0; see Section 3.4), then the generation of k is
    finished.  The obtained value of k is used in DSA or ECDSA.
    Otherwise, compute:

    K = HMAC_K(V || 0x00)

    V = HMAC_K(V)

    and loop (try to generate a new T, and so on).
    // "and loop (try to generate a new T)" means that AFTER performing the K and V steps after "Otherwise, compute:" you must return back and start over generating a new T again using the: empty T -> HMAC_K(V) -> T = T || V etc. (though in our case it will only be one HMAC_K(V) because hlen is qlen)
```

To to reiterate what happens if there is a bad r or s, OR an out of bounds k...

```
    K = HMAC_K(V || 0x00)

    V = HMAC_K(V)

    T = []  // tlen now 0 again

    V = HMAC_K(V)

    T = T || V  // tlen = qlen, as sha256 will always be 256 bits and curve order (qlen) is 256 bits. So effectively same as T = V

    k = bits2int(T) // we're back to where we looped again, this is next k to check.
```

bip32JP:
> But the extra HMAC computation should only happen with badrs > 0.

RFC 6979 states

> If that **value of k is within the [1,q-1] range**, and is
> suitable for DSA or ECDSA (i.e., it **results in an r value
> that is not 0**; see Section 3.4), then the generation of k is
> finished.  The obtained value of k is used in DSA or ECDSA.
> **Otherwise**, compute:
> ...
> and loop...

So both 1. out of bounds k AND 2. bad r or s should follow the same loop of:
```
    K = HMAC_K(V || 0x00)

    V = HMAC_K(V)

    V = HMAC_K(V)

    T = V

    k = bits2int(T)

    Check k
```

ryanxcharles:
> So both 1. out of bounds k AND 2. bad r or s should follow the same loop of

Thanks @bip32JP . I think the loop you modifies does exactly that. Here's the conditional:

```
i < badrs || !(T.lt(N) && T.gt(0))
```

The extra HMAC you've added is performed even when r, s and k are all valid, no? The test-case you modified was a normal signature that does not trigger badrs or out of bounds k. Did you run that same test-case through python-ecdsa?

bip32JP:
```
!(T.lt(N) && T.gt(0))
this will only trigger if T is NOT (less than N and greater than 0) aka this will only trigger if T is out of bounds.

i = 0; i < badrs
This will only trigger if badrs is greater than 0. if it is 0 (default) it will not trigger.
```
So if badrs is 0 and T is in bounds, none of the code inside that for loop will be executed.

Try it for yourself, place a `console.log( 'TEST' );` inside there and sign something.

bip32JP:
> The test-case you modified was a normal signature that does not trigger badrs or out of bounds k

Incorrect. the test case I modified was with badrs = 1.

ryanxcharles:
You're right, my bad. I assumed there was no test case for badrs > 0 - however, there is a test case for that. What there is no test case for is actually computing a signature that triggers badrs. The loop is actually triggered only in the case where there is a bad rs or out of bounds k, so your PR is correct.

PR #888, opened 2015-01-02
Use defaultNetwork in Script.toAddress

### LINKED ISSUE #883: Script.prototype.toAddress and Networks.defaultNetwork
Expected behavior:
```
var address = script.toAddress();
// uses the current default network
```
Current behavior:
```
var address = script.toAddress(Networks.defaultNetwork);
// requires passing the default network as an argument
```

yemel:
Closes #883

### DIFF
--- a/lib/script/script.js
+++ b/lib/script/script.js
@@ -604,6 +604,7 @@ Script.buildPublicKeyHashOut = function(to) {
     .add(to.hashBuffer)
     .add(Opcode.OP_EQUALVERIFY)
     .add(Opcode.OP_CHECKSIG);
+  s._network = to.network;
   return s;
 };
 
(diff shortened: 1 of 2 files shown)


braydonf:
If a 'testnet' address was used to instantiate a script, and the default network was 'livenet', would `script.toAddress()` return the same address?

eordano:
No, it wouldn't. I think it's fine, we can only go so far on babysitting the user?

@braydonf please merge if you agree, otherwise let's discuss alternatives

braydonf:
I think this should be true, in all cases:
```
var x = script.toAddress(script.fromAddress(y));
assert.equal(x, y);
```

yemel:
`.toAddress(...)` receives a `network` object not a `script`. Do you mean:

```
assert.equal(address, script.fromAddress(address).toAddress())
```

eordano:
That would need us to add a "network" to each script. I'd rather not :(

braydonf:
@yemel. Yep, edited to be correct.

braydonf:
Doesn't `Address` accept Script as input? It should be possible to then do:
```
var address = new Address(script);
```
And script.toAddress isn't needed?

yemel:
I would like to keep the `toAddress` method on Script, mostly for convenience.

maraoz:
ACK

braydonf:
ACK

PR #889, opened 2015-01-02
Add method for validating a derivation path

### LINKED ISSUE #878: Add HDPrivateKey.isValidPath()
The only way to check if a path is valid is trying to derive the key and catch the error. HDPrivateKey and/or HDPublicKey should offer a helper for this.

yemel:
Closes #878

### DIFF
--- a/lib/hdprivatekey.js
+++ b/lib/hdprivatekey.js
@@ -63,6 +63,22 @@ function HDPrivateKey(arg) {
 }
 
 /**
+ * Verifies that a given path is valid.
+ *
+ * @param {string|number} arg
+ * @param {boolean?} hardened
+ * @return {boolean}
+ */
+HDPrivateKey.prototype.isValidPath = function(arg, hardened) {
+  try {
+    this.derive(arg, hardened);
+    return true;
+  } catch (err) {
+    return false;
+  }
+};
+
+/**
  * Get a derivated child based on a string or number.
  *
  * If the first argument is a string, it's parsed as the full path of
@@ -101,9 +117,15 @@ HDPrivateKey.prototype._deriveWithNumber = function(index, hardened) {
   if (index >= HDPrivateKey.Hardened) {
     hardened = true;
   }
+
   if (index < HDPrivateKey.Hardened && hardened) {
     index += HDPrivateKey.Hardened;
   }
+
+  if (index < 0 || index >= HDPrivateKey.MaxIndex) {
+    throw new hdErrors.InvalidPath(index);
+  }
+
   var cached = HDKeyCache.get(this.xprivkey, index, hardened);
   if (cached) {
     return cached;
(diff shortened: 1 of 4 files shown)


test/hdprivatekey.js
```diff
@@ -190,6 +190,24 @@ describe('HDPrivate key interface', function() {
...
+  it('validates correct paths', function() {
+    var privateKey = new HDPrivateKey(xprivkey);
+    var valid = privateKey.isValidPath('m/0\'/1/2\'');
+    valid.should.equal(true);
+
+    var valid = privateKey.isValidPath(123, true);
+    valid.should.equal(true);
+  });
+
+  it('validates illigal paths', function() {
```
braydonf: illegal

test/hdpublickey.js
```diff
@@ -216,10 +216,28 @@ describe('HDPublicKey interface', function() {
...
+  it('validates correct paths', function() {
+    var publicKey = new HDPublicKey(xpubkey);
+    var valid = publicKey.isValidPath('m/123/12');
+    valid.should.equal(true);
+
+    var valid = publicKey.isValidPath(123, true);
+    valid.should.equal(true);
+  });
+
+  it('validates illigal paths', function() {
```
braydonf: another typo

lib/hdpublickey.js
```diff
@@ -65,6 +65,22 @@ function HDPublicKey(arg) {
...
 }
 
+
+/**
+ * Verifies that a given path is valid.
+ *
+ * @param {string|number} arg
+ * @return {boolean}
+ */
+HDPublicKey.prototype.isValidPath = function(arg) {
```
braydonf: I wonder if instead of a try/catch here we could use a regex to validate the path?

yemel: I thought about that, the issue is being as flexible as the `derive` method (that accepts number/boolean or strings). However another Pro of avoiding the try/catch is we can expose the method as a static function.

I didn't worry too much about the calculation because the class uses a cache, and if you check for a path validity you most probably want to derive it.

Do you think the style improvement worth the change?

braydonf: The cache is a good point. I think it's fine for now, however we may want to revisit it later. And have a method that is used to validate the path that is the same that is used for validation within the derive method.

eordano:
LGTM after typo fix

yemel:
Updated to be static an static function. Reused the code on HDPublicKey as well.

lib/hdprivatekey.js
```diff
@@ -63,6 +63,58 @@ function HDPrivateKey(arg) {
...
+    return arg >= 0 && arg < HDPrivateKey.MaxIndex;
+  }
+
+  return false;
+};
+
+/**
+ * Internal function that splits a string path into a derivation index array.
+ * It will return null if the string path is malformed.
+ * It does not validates if a indexes are in bounds.
```
braydonf: It does not validate if indexes are in bounds.

braydonf:
LGTM after the typo fix commented above

lib/hdprivatekey.js
eordano: So, this looks like it just splits the paths. The validation seems to be in the `isValidPath` function, it checks in the line 82 for `arg >= 0 && arg < HDPrivateKey.MaxIndex;` As it's a internal function it looks OK to me.

eordano:
LGTM - don't think any modifications are needed to approve

lib/hdprivatekey.js
maraoz: @braydonf is commenting on the comment's grammar: "It does not validates if a indexes are in bounds." should read "It does not validate if indexes are in bounds."

eordano: Oh sorry

PR #892, opened 2015-01-02
Change PrivateKey.toString from WIF to hex format

### LINKED ISSUE #876: Add PrivateKey.toHex() method
Trivial change, alias for:
`privateKey.toBuffer().toString('hex');`

yemel:
This way is both a PublicKey and PrivateKey implements their `toString` method as a `this.toBuffer().toString('hex');`. IMO it's clearer and removes the need for a `.toHex()` method.

Closes #876

### DIFF
--- a/lib/privatekey.js
+++ b/lib/privatekey.js
@@ -284,11 +284,20 @@ PrivateKey.isValid = function(data, network){
 };
 
 /**
+ * Will output the PublicKey encoded as hex string
+ *
+ * @returns {String}
+ */
+PrivateKey.prototype.toString = function() {
+  return this.toBuffer().toString('hex');
+}
+
+/**
  * Will output the PrivateKey to a WIF string
  *
  * @returns {String} A WIP representation of the private key
  */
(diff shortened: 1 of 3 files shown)


lib/privatekey.js
```diff
@@ -284,11 +284,20 @@ PrivateKey.isValid = function(data, network){
 };
 
 /**
+ * Will output the PublicKey encoded as hex string
```
braydonf: typo: PrivateKey

test/privatekey.js
```diff
@@ -316,7 +316,7 @@ describe('PrivateKey', function() {
 
     it('should parse this uncompressed livenet address correctly', function() {
       var privkey = PrivateKey.fromString(wifLivenetUncompressed);
-      privkey.toString().should.equal(wifLivenetUncompressed);
+      privkey.toString().should.equal("96c132224121b509b7d0a16245e957d9192609c5637c6228311287b1be21627a");
```
braydonf: The test here is checking that the input and output match, perhaps `toWIF` should be used here?

PR #894, opened 2015-01-03
Fixed public key calculation in rare situations that would produce an invalid point

braydonf:


ryanxcharles:
Wow... why is this? I guess there is something about elliptic's internal structure of G that isn't being copied with ``Point(ec.curve.g.getX(), ec.curve.g.getY())`` ?

Another way to fix it is:

```
var g = new Point();
ec.curve.g.copy(g);
```

...in case you would rather copy than reference elliptic's value of g.

Also confirmed that this does not fix it:

```
var g = Point.fromX(ec.curve.g.isOdd, ec.curve.g.getX());
```

It's disturbing that this happens. If the x and y values are not enough to specify elliptic's internal representation of a point, then the toString/fromString and toJSON/fromJSON methods are also wrong.

ryanxcharles:
After glancing at elliptic's "g" for secp256k1, stored in curves.js, I see this:

```
  g: [
    '79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798',
    '483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8',
    {
      'doubles': {
        'step': 4,
        'points': [
          [
            'e60fce93b59e9ec53011aabc21c23e97b2a31369b87a5ae9c44ee89e2a6dec0a',
            'f7e3507399e595929db99f34f57937101296891e44d23f0be1f32cce69616821'
          ],
          [
            '8282263212c609d9ea2a6e3e172de238d8c39cabd5ac1ca10646e23fd5f51508',
            '11f8a8098557dfe45e8256e830b60ace62d613ac2f7b17bed31b6eaff6e26caf'
          ],
...
```

Perhaps elliptic expects these precomputed values to be there when doing point multiplication, and when not, it sometimes silently fails and derives the wrong value?

ryanxcharles:
I think that's what it is. When I run the "precompute" function on G, and then do the point multiplication, things work.

Changing this:
```
  var p = Point(ec.curve.g.getX(), ec.curve.g.getY());
  return p;
```

Into this:
```
  var p = Point(ec.curve.g.getX(), ec.curve.g.getY());
  p.precompute(Point.getN().bitLength() + 1);
  return p;
```

fixes the problem. (although it makes things run much slower because, of course, the precomputed values are there for speed). So elliptic is sometimes silently failing during point multiplication when precomputed values aren't available. This is a bug in elliptic somewhere.

ryanxcharles:
Rigging a new "mul" function to always precompute values also fixes the problem:
```
var _mul = Point.prototype.mul;
Point.prototype.mul = function(bn) {
  if (!this.precomputed)
    this.precompute(Point.getN().bitLength() + 1);  
  return _mul.call(this, bn);
};
```
However, this slows things down dramatically. I think the real answer is to figure out why elliptic is silently failing and fix that.

ryanxcharles:
...also, one more addendum. This doesn't mean toString/fromString or toJSON/fromJSON are wrong. What's wrong is elliptic's "mul" function. That is where the bug lies.

braydonf:
I found another case where commit bc4e62ecd4f1aa8e3f72afc3403ae27c9e151bf3 didn't fix the issue, and indeed, looks like there is a bug in the "mul" function:

```javascript
// elliptic.js

var EC = require('elliptic').ec
var ec = new EC('secp256k1')

var point = ec.g.mul('f2cc9d2b008927db94b89e04e2f6e70c180e547b3e5e564b06b8215d1c264b53')
// <EC Point x: 495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba y: 41a2d4539b40a914e755a3a24367ce731195c354480dad4f2ce9e9c8f8003bce>

var point1 = e.curve.pointFromX(true, '495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba')
// <EC Point x: 495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba y: 3dc9318f0a7ae59bafbccef98692931dc6a8c28208333e88de5c031766f70777>

var point2 = e.curve.pointFromX(false, '495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba')
// <EC Point x: 495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba y: c236ce70f5851a6450433106796d6ce239573d7df7ccc17721a3fce79908f4b8>

var keypair = e.keyPair('f2cc9d2b008927db94b89e04e2f6e70c180e547b3e5e564b06b8215d1c264b53')
// <Key priv: f2cc9d2b008927db94b89e04e2f6e70c180e547b3e5e564b06b8215d1c264b53 pub: null >

var pub = keypair.getPublic(true, 'hex')
// incorrect pub: '02495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba'


// bitcoinjs

var bitcoinjs = require('bitcoinjs-lib');

var eckey = bitcoinjs.ECKey.fromWIF('L5MgSwNB2R76xBGorofRSTuQFd1bm3hQMFVf3u2CneFom8u1Yt7G')
var priv = eckey.d.toHex()
// 'f2cc9d2b008927db94b89e04e2f6e70c180e547b3e5e564b06b8215d1c264b53'

var pub = eckey.pub.toHex()
// currect pub: '03e275faa35bd1e88f5df6e8f9f6edb93bdf1d65f4915efc79fd7a726ec0c21700'


// lookup the associated address of the different results

var bitcore = require('bitcore');
var PublicKey = bitcore.PublicKey;
var Address = bitcore.Address;

var address = new Address.fromPublicKey(new PublicKey('03e275faa35bd1e88f5df6e8f9f6edb93bdf1d65f4915efc79fd7a726ec0c21700'));
// <Address: 19FKem9YKbszEwsXRLw9WnkggMQrdLbAhE, type: pubkeyhash, network: livenet>

var address1 = new Address.fromPublicKey(new PublicKey('02495f99aa21d7e60ff3f065f2a14e67612b29b1cb5af7b7fc4738cede6f2d9eba'));
// <Address: 1279WAo1PmBoYmu12UeQdsGN2ZqPqPuCzb, type: pubkeyhash, network: livenet>


// bitcoin-qt debug window

importprivkey "L5MgSwNB2R76xBGorofRSTuQFd1bm3hQMFVf3u2CneFom8u1Yt7G"
dumpprivkey "19FKem9YKbszEwsXRLw9WnkggMQrdLbAhE"
// outputs: L5MgSwNB2R76xBGorofRSTuQFd1bm3hQMFVf3u2CneFom8u1Yt7G
dumpprivkey "1279WAo1PmBoYmu12UeQdsGN2ZqPqPuCzb"
// outputs "Private key for address 1279WAo1PmBoYmu12UeQdsGN2ZqPqPuCzb is not known (code -4)"
```

braydonf:
Ready for review

ryanxcharles:
ACK. I also think this reminds us that a full audit of bn.js and elliptic would be valuable. There could be other gotchas with these libraries that could lead to lost funds. One way to do that is to rewrite both of them, create test vectors that test all edge cases (which you will thoroughly understand if you rewrite them), then make sure the same test vectors pass on all libraries.

eordano:
LGTM

yemel:
ACK

PR #896, opened 2015-01-05
add deterministicK test vectors

ryanxcharles:
These vectors were produced by @bip32JP and tested against a similar
implementation of the "badrs" approach to RFC 6979 in python-ecdsa as described
in this related bitcoinjs-lib PR:

https://github.com/bitcoinjs/bitcoinjs-lib/pull/337


eordano:
LGTM

braydonf:
The likelihood of triggering a badrs > 0 may be possible to consider completely improbable.

PR #903, opened 2015-01-05
Add `change` when serializing transaction

eordano:


lib/address.js
```diff
@@ -112,6 +112,8 @@ Address.prototype._classifyArguments = function(data, network, type) {
     return Address._transformScript(data, network);
   } else if (typeof(data) === 'string') {
     return Address._transformString(data, network, type);
+  } else if (_.isObject(data) && data.network && data.hash && data.type) {
```
braydonf: Rather than checking for the properties here, could we use preconditions to check that they are set?

lib/address.js
```diff
@@ -140,6 +142,20 @@ Address._transformHash = function(hash){
...
 
 /**
+ * Deserializes an address serialized through `Address#toObject()`
+ * @param {Object} data
+ * @param {string} data.hash - the hash that this address encodes
+ * @param {string} data.type - either 'pubkeyhash' or 'scripthash'
+ * @param {network} data.network - the name of the network associated
+ * @return {Address}
+ */
+Address._transformObject = function(data) {
```
braydonf: Does type need to be set here? Its an argument, but looks like it isn't used.

braydonf:
Does _transformObject have test coverage?

lib/address.js
eordano: I'd rather only take this path if the properties are set. Maybe in the future there are other structures for objects that we can accept.

lib/address.js
eordano: It is used, it's needed when returned (I'll add a precondition)

eordano:
Yep, it has coverage: https://coveralls.io/files/412568217#L152

I'll improve slightly on this and get back to you

lib/address.js
braydonf: I was thinking it would be better to take the path, and then receive precondition error, than to spend time trying to figure out why why the object wasn't being accepted an as input because the property was slightly incorrect.

yemel:
ACK

Issue #784, opened 2014-12-17
check that constructor.name works in all browsers


braydonf:

braydonf:
Fixed in #904

PR #904, opened 2015-01-05
Fixed bugs in Safari and IE. Closes #837 and #784

### LINKED ISSUE #837: Browser Compatibility
Run tests in IE (latest versions), and various other operating systems to check that all tests are passing.

IE11: https://gist.github.com/braydonf/61459334f5a86dc5f7be

braydonf:


eordano:
LGTM

test/util/preconditions.js
```diff
@@ -52,7 +52,8 @@ describe('preconditions', function() {
       $.checkArgumentType(1, PrivateKey);
     } catch (e) {
       error = e;
-      e.message.should.equal('Invalid Argument for (unknown name), expected PrivateKey but got number');
```
maraoz: what is `e.message` in other browsers? I'm worried this assertion is failing there

braydonf: The name of the function usually appears, in this case "PrivateKey"

maraoz: "PrivateKey" instead of "(unknown name)"? in that case, ACK to this PR

eordano: Yep, an extra argument should be added to the error creation

braydonf: The message stays the same for browsers that support `constructor.name`, and likewise a toString() method.

PR #905, opened 2015-01-06
Fixed tests to handle a `defaultNetwork` change. Closes #890

### LINKED ISSUE #890: If Networks.defaultNetwork is set to testnet, 14 tests fail.
This means that either we have some code or tests that depend on the default network being livenet. The first one is dangerous, so we should fix all tests to be defaultNetwork-proof so we are sure that our code is strong against changes to defaultNetwork (which probably is)

braydonf:
- Updated tests to work for both 'testnet' and 'livenet' as the default network.
- Fixed a bug in Address where the network property was being set as a string.
- Fixed a bug in HDKeys to handle serialized keys when the defaultNetwork changed.

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -244,7 +244,7 @@ Address._transformScript = function(script, network){
     info.hashBuffer = Hash.sha256ripemd160(script.toBuffer());
     info.type = Address.PayToScriptHash;
   }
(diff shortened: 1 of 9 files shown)


eordano:
You rock dude

Issue #908, opened 2015-01-07
add version member to bitcore object


maraoz:
I think this would be useful:

```
var bitcore = require('bitcore');
console.log(bitcore.version);
// prints 'v0.8.6'
```

thofmann:
Implemented and merged: #915

braydonf:
Thanks!

PR #915, opened 2015-01-08
Add version member to bitcore object

thofmann:
Resolves issue #908

var bitcore = require('bitcore');
console.log(bitcore.version);
// prints 'v0.8.6'


maraoz:
LGTM

PR #924, opened 2015-01-09
Added getID to PublicKey

braydonf:
- Tests in BloomFilter in [bitcoin core](https://github.com/bitcoin/bitcoin/blob/master/src/test/bloom_tests.cpp#L85) use the CPubKey.GetID as input into the BloomFilter
- The hash is calculated when generating an Address however the hash itself is not currently available

### DIFF
--- a/lib/publickey.js
+++ b/lib/publickey.js
@@ -385,12 +385,12 @@ PublicKey.prototype.toBuffer = PublicKey.prototype.toDER = function() {
   }
 };
 
-/*
+/**
  * Will return a sha256 + ripemd160 hash of the serialized public key
  * @see https://github.com/bitcoin/bitcoin/blob/master/src/pubkey.h#L141
  * @returns {Buffer}
  */
-PublicKey.prototype.getID = function getID() {
+PublicKey.prototype._getID = function _getID() {
   return Hash.sha256ripemd160(this.toBuffer());
 };
 
--- a/lib/publickey.js
+++ b/lib/publickey.js
@@ -3,6 +3,7 @@
 var Address = require('./address');
 var BN = require('./crypto/bn');
 var Point = require('./crypto/point');
+var Hash = require('./crypto/hash');
 var JSUtil = require('./util/js');
 var Network = require('./networks');
 var _ = require('lodash');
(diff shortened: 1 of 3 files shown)


lib/publickey.js
```diff
@@ -384,6 +385,15 @@ PublicKey.prototype.toBuffer = PublicKey.prototype.toDER = function() {
...
 };
 
+/*
+ * Will return a sha256 + ripemd160 hash of the serialized public key
+ * @see https://github.com/bitcoin/bitcoin/blob/master/src/pubkey.h#L141
+ * @returns {Buffer}
+ */
+PublicKey.prototype.getID = function getID() {
+  return Hash.sha256ripemd160(this.toBuffer())
+}
```
eordano: missing `;`

lib/publickey.js
```diff
@@ -384,6 +385,15 @@ PublicKey.prototype.toBuffer = PublicKey.prototype.toDER = function() {
   }
 };
 
+/*
+ * Will return a sha256 + ripemd160 hash of the serialized public key
+ * @see https://github.com/bitcoin/bitcoin/blob/master/src/pubkey.h#L141
+ * @returns {Buffer}
+ */
+PublicKey.prototype.getID = function getID() {
```
eordano: Can we rename this to `fingerprint`, `hash`, or `hash160` and/or make this start with `_`? I feel strange about "the ID of a public key"

braydonf: Hash has an entirely different meaning of a double sha256 hash, see: https://github.com/bitcoin/bitcoin/blob/master/src/pubkey.h#L148

lib/publickey.js
```diff
@@ -384,6 +385,15 @@ PublicKey.prototype.toBuffer = PublicKey.prototype.toDER = function() {
   }
 };
 
+/*
```
eordano: Missing extra `*`

lib/publickey.js
eordano: I see. Still hate the name, but it's not that bad as it's the name in the ref implementation.

eordano:
LGTM after missing `*` on jsdoc

Issue #882, opened 2015-01-02
Output Inspect Method


braydonf:
The inspect method isn't defined on Output, and is displaying in a confusing manner that only shows the internal values `_scriptBuffer` and `_satoshis`.

PR #935, opened 2015-01-09
Add inspect method to output

### LINKED ISSUE #881: Confusing use of output._satoshis vs output.satoshis
`output._satoshis` is a BN, and `output.satoshis` is a Number. One appears to be used for internal usage and hence the underscore, however a number should be sufficient since the maximum number of satoshis is 2100000000000000 which can accurately be represented without BN.

eordano:
Closes #881 #882


maraoz:
ACK

Issue #937, opened 2015-01-09
Rename internal use of `_satoshis` that is a BN to `_satoshisBN`


braydonf:
To help distinguish between the properties `*.satoshi` and `*._satoshi`, we can rename `._satoshis` to `_satoshisBN`. This is recommended because the confusion was the cause of a bug that was fixed: https://github.com/braydonf/bitcore/commit/bce28cd227d6e6c2a746ff655909c5e9e567ddad#diff-3ff19b4e31a68503f7ddd89615e606d7L659

PR #942, opened 2015-01-09
Use satoshisBN internally

eordano:
Closes https://github.com/bitpay/bitcore/issues/937


braydonf:
Can we remove `_satoshis` in favor of `_satoshisBN`? There is now `.satoshis`, `._satoshis`, `._satoshisBN`

eordano:
@braydonf let's do that on another issue: https://github.com/bitpay/bitcore/issues/950

PR #945, opened 2015-01-10
Add alternative to create an Address

eordano:
When (by error or any chance) one has two bitcore instances, an address can't be instantiated from the other bitcore instance's address. This fixes that


maraoz:
LGTM

PR #947, opened 2015-01-11
Added mempool request

elichai:
I added `GetMempool` function that sends an 'mempool' request.
the answer will come in 'inv' message.

### DIFF
--- a/lib/transport/messages.js
+++ b/lib/transport/messages.js
@@ -584,6 +584,17 @@ function GetHeaders(starts, stop) {
 util.inherits(GetHeaders, GetBlocks);
 module.exports.GetHeaders = Message.COMMANDS.getheaders = GetHeaders;
 
+/**
+ * GetMempool Message
+ *
+ * @name Transport.Message.GetMempool
+ */
+function GetMempool() {
+  this.command = 'mempool';
+}
+
+util.inherits(GetMempool, Message);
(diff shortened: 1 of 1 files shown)


eordano:
Looks good, even though untested. I'll merge in this changes into `bitcore-p2p` (https://github.com/bitpay/bitcore-p2p) as this is going to go away with https://github.com/bitpay/bitcore/pull/952

elichai:
So wait. to switch to bitcore-p2p for p2p requests?

elichai:
btw, I tested it (couple of time with couple of nodes) and it worked just fine.

braydonf:
@elichai A mocha test in the `test` directory, so that the functionality can be added to the test coverage. You can see the `pool` tests to see how to fake a message from a peer, the `addr` messages for example.

elichai:
ok, i'll work on it :)

eordano:
Hmm, those tests need some love. They are actually quite dumb... I may change them

eordano:
(Sorry about my last message, seemed quite harsh)

Hey @elichai, don't worry about it. These tests don't look like they are actually testing much, not doing another thing than adding test coverage. I'll update the test code to automatically run on each message.

Just compare a good test with a nice description to a test that adds nothing, like the ones for messages:

screen shot 2015-01-12 at 21 04 12
screen shot 2015-01-12 at 21 03 43

PR #955, opened 2015-01-13
new error system with no code generation

maraoz:
This removes the error code generator, and just builds the error classes in memory. 

It also adds an error extension function for submodules. See example use here: https://github.com/bitpay/bitcore-ecies/pull/10/files#diff-11c8ed6891535db0efdc438909405929R19


maraoz:
updated and ready for review!

PR #956, opened 2015-01-13
Add more networks in networks.js

abstream:
I can see only these networks:

['bitcoin', 'litecoin', 'dogecoin', 'darkcoin'];

would it be appropriate to add other bitcoin derived networks such as namecoin for example?
how about other bitcoin derived/similar networks?

thanks.


eordano:
Hi there,

We have been working on a rewrite of this library. Take a look at the current code, as it's fundamentally different from what we had before (we're calling these versions 0.8 and 0.1 for the old branch).

It's not in our scope to add more networks to this library. Feel free to fork it to add other networks. It's also troublesome because of different logic rules: it's probably not the case with litecoin or dogecoin, but the transactions on namecoin require a different handling than those on bitcoin.

Thanks!

abstream:
Hello,

I installed the master and was able to sign and broadcast litecoin transaction from the first try!
It is really hard to find network info and make it work with networks.js....can you point me to relevant networks that will be easy to explore - I am looking into adding any number of compatible networks in my app....for example i didn't know that there will be problems with namecoin...

thanks.

maraoz:
@abstream look here: https://github.com/bitpay/bitcore/blob/master/lib/networks.js

PR #960, opened 2015-01-14
Drop files that go into bitcore-explorers

eordano:


maraoz:
ACK. waiting for travis

PR #976, opened 2015-01-18
Added JavaScript syntax highlighting

himynameismartin:
Hi, just a quick question regarding documentation formatting.

Every file at https://github.com/bitpay/bitcore/tree/master/docs/guide starts with two lines:
**title: Address**
**description: A simple interface to generate and validate a bitcoin address.**

followed by a line containing **---**, which in Markdown speech means &lt;h2&gt;ing the previous line.

Was it meant to introduce a horizontal rule instead (an empty line before the **---** is missing then)?

If it should be so, I can easily fix it, just didn't want to waste the time in case I'm wrong.

Best, Martin

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -5,24 +5,27 @@ description: Sample code for the most common task in any bitcoin application.
 
 ## Create a Private Key
 
-```
+```javascript
 var privKey = new bitcore.PrivateKey();
 ```
 
 ## Create an Address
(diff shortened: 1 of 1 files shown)


braydonf:
Thanks for the fixes.

The title and description are yaml intended for the hexo, when the docs are generated into html. There may be a better way to handle this, however it is the best solution currently.

PR #974, opened 2015-01-16
Removed devDependencies that can be defined in bitcore-build.

braydonf:

PR #977, opened 2015-01-19
Fixed the build by specifying the exact version of bitcore-build.

braydonf:


eordano:
Shouldn't we merge https://github.com/bitpay/bitcore/pull/974 instead of this?

PR #998, opened 2015-01-23
Networks: Added the ability to define a custom network. Closes #997

### LINKED ISSUE #997: Custom Networks
In some situations it may be useful to be able to send a custom network to a class, for example Address so that other types of prefixes can be used. Exposing the Network class in Networks, and then defining an additional network should be possible, and then having `getNetwork` retrieve from this list for network validation.

braydonf:


eordano:
LGTM!

PR #1004, opened 2015-01-27
Transaction: expose the sighash namespace

eordano:


maraoz:
ACK

eordano:
Changed the tests to use the exposed sighash object. Will review similar cases on other tests.

yemel:
ACK

PR #1006, opened 2015-01-27
Add fromBuffer and toBuffer to PrivateKey

### LINKED ISSUE #739: PrivateKey `fromBuffer` G7
- Add `fromBuffer` static method to PrivateKey - [G7](https://github.com/bitpay/bitcore/blob/v0.8/CONTRIBUTING.md#g7---general-standard-methods)

eordano:
Closes #739

### DIFF
--- a/lib/privatekey.js
+++ b/lib/privatekey.js
@@ -152,12 +152,8 @@ PrivateKey._transformBuffer = function(buf, network) {
 
   var info = {};
 
-  if (buf.length === 1 + 32 + 1 && buf[1 + 32 + 1 - 1] === 1) {
-    info.compressed = true;
-  } else if (buf.length === 1 + 32) {
-    info.compressed = false;
-  } else {
-    throw new Error('Length of buffer must be 33 (uncompressed) or 34 (compressed)');
+  if (buf.length === 32) {
+    return PrivateKey._transformBNBuffer(buf, network);
   }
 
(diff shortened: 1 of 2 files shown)


braydonf:
In what situation is a BN is a buffer? Usually most BN's are an instance of BN or a hexa.

eordano:
Is it ok with you if I change the current `fromBuffer` (which takes a DER encoded key) to something like `fromDERBuffer` instead of a `fromBNBuffer`? The use case I'm thinking of is `new PrivateKey(get256RandomBitsAsBuffer())`.

I'd like to keep as much versatility in the constructor as possible, because otherwise a new user needs to read through the doc to remember what was the particular method's name that he needed to use.

braydonf:
The current `toBuffer` is the BN buffer (I forgot about that). So `fromBuffer` should accept the output from that.

lib/privatekey.js
```diff
@@ -282,7 +282,17 @@ PrivateKey.isValid = function(data, network){
...
-}
+};
+
+/**
+ * Returns a private key generated by a big number serialized into a buffer
+ *
+ * @param {Buffer} arg
+ * @returns {PrivateKey}
+ */
+PrivateKey.fromBuffer = function(buffer, arg0, arg1) {
```
braydonf: `arg0` I believe is network, and arg1 wouldn't be used.

maraoz:
ACK

PR #1009, opened 2015-01-28
Add option to retrieve an address from an output

eordano:


lib/transaction/output.js
```diff
@@ -32,6 +33,18 @@ Object.defineProperty(Output.prototype, 'script', {
   }
 });
 
+Object.defineProperty(Output.prototype, 'address', {
+  configurable: false,
+  writeable: false,
+  enumerable: true,
+  get: function() {
+    if (!this._address) {
+      this._address = new Address(this.script);
```
yemel: If no address associated to the script the accessor will throw an error.

I think returning null is a good solution for this scenario, that provides a clean way to check if the output has an address without a try/catch block. The other option is to make it a function and let the error bubble up.

maraoz:
I would do it via a method, not a property.

eordano:
Closing, the `toAddress` method already exists in the `script` property of the output.

PR #1010, opened 2015-01-28
Allow satoshis to be a string on output creation

eordano:


maraoz:
ACK. waiting for travis

PR #1016, opened 2015-01-28
Fix hd string path parsing

### LINKED ISSUE #1014: Bug in bitcore.HDPrivateKey.isValidPath
```
var isValid = bitcore.HDPrivateKey.isValidPath('m/12asd'); //  returns true;
```

And you can also derive it
```
var xkey = new bitcore.HDPrivateKey();
var derived1 = xkey.derive('m/12asd');
var derived2 = xkey.derive('m/2147483660');

assert.equals(derived1, derived2);
```

yemel:
Closes #1014

### DIFF
--- a/lib/hdprivatekey.js
+++ b/lib/hdprivatekey.js
@@ -106,8 +106,13 @@ HDPrivateKey._getDerivationIndexes = function(path) {
   }
 
   var indexes = steps.slice(1).map(function(step) {
-    var index = parseInt(step);
-    index += step != index.toString() ? HDPrivateKey.Hardened : 0;
+    var index = step ? +step : NaN;
+
+    var isHardened = isNaN(index) && step[step.length-1] == "'";
+    if (isHardened) {
+      index = (+(step.slice(0, -1))) + HDPrivateKey.Hardened;
+    }
+
     return index;
   });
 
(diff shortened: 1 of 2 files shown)


eordano:
LGTM

lib/hdprivatekey.js
```diff
@@ -106,8 +106,13 @@ HDPrivateKey._getDerivationIndexes = function(path) {
...
-    index += step != index.toString() ? HDPrivateKey.Hardened : 0;
+    var index = step ? +step : NaN;
+
+    var isHardened = isNaN(index) && step[step.length-1] == "'";
+    if (isHardened) {
+      index = (+(step.slice(0, -1))) + HDPrivateKey.Hardened;
+    }
+
     return index;
   });
```
maraoz: This code is way too complicated for what it does. I don't want to be a PITA, but what can you consider rewriting it in a simpler way? 
I'm talking about the whole `_getDerivationIndexes` function

maraoz: or maybe we can do it in another PR if you want

eordano:
```javascript
  var privateValidPath = function(path) {
    return !!(/^[mM][']?(\/[0-9]+[']?)*[/]?$/.exec(path));
  };
  var publicValidPath = function(path) {
    return !!(/^[mM](\/[0-9]+)*[/]?$/.exec(path));
  };
```

PR #1027, opened 2015-02-04
Add documentation for instantiating P2SH multisig

thofmann:
There is documentation available at
http://bitcore.io/api/#Create_a_Multisig_Address but I think it would
also be useful in the address.md document.

### DIFF
--- a/docs/address.md
+++ b/docs/address.md
@@ -38,7 +38,7 @@ A pay-to-script-hash multisignature Address can be instantiated from an array of
 
 ```javascript
 // a 2-of-3 address from public keys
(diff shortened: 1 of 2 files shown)


braydonf:
LGTM

thofmann:
Thanks. I agree with lowerCamelCase, but was mostly following the example from [here](http://bitcore.io/api/#Create_a_Multisig_Address) (also found [here](https://github.com/bitpay/bitcore/blob/970bf12df882e274c28a57ec1af5cf5850ec6efd/docs/index.md#create-a-multisig-address)).

Should this example be changed as well?

PR #1028, opened 2015-02-04
Improvements to Transaction

eordano:
This commit contains a series of modifications that where detected from @matiu's feedback:

* The `addInput` method was not being used when `from` was called with a non-p2sh output
* The change output had no actual reason not to be updated on each modification, as there is no private key cache that could be signing more often than what we wanted for performance issues
* Add a `fee` field to the `toObject` serialization


maraoz:
LGTM

maraoz:
please check test coverage: ` coverage/coveralls — Coverage decreased (-0.02%) to 96.89%1`

PR #1041, opened 2015-02-05
UnspentOutput: Add tests to check for zero amount of satoshis

eordano:


test/transaction/unspentoutput.js
```diff
@@ -53,6 +53,30 @@ describe('UnspentOutput', function() {
...
+    var zero = {
+      'txId': 'a477af6b2667c29670467e4e0728b685ee07b240235771862318e29ddbe58458',
+      'outputIndex': 0,
+      'script': 'OP_DUP OP_HASH160 20 0x88d9931ea73d60eaf7e5671efc0552b912911f2a OP_EQUALVERIFY OP_CHECKSIG',
+      'amount': 0
+    };
+    it('fails when no amount is defined', function() {
+      expect(function() {
+        return new UnspentOutput(notDefined);
+      }).to.throw();
```
maraoz: please add more specificity, i.e:
```
.to.throw('Must provide an amount for the output');
```
This helps some bugs from remaining hidden (we had this problem in the far past)

eordano: Updated

PR #1048, opened 2015-02-06
Add some useful Transaction methods to handle nLockTime

maraoz:
This PR also:
- fixes 0 fee handling
- fixes change address logic
- adds Transaction#getChangeOutput

### DIFF
--- a/lib/transaction/transaction.js
+++ b/lib/transaction/transaction.js
@@ -226,33 +226,46 @@ Transaction.prototype.toObject = function toObject() {
   this.outputs.forEach(function(output) {
     outputs.push(output.toObject());
   });
-  return {
-    changeScript: this._changeScript ? this._changeScript.toString() : undefined,
-    changeIndex: !_.isUndefined(this._changeIndex) ? this._changeIndex : undefined,
-    fee: this._fee ? this._fee : undefined,
+  var obj = {
     version: this.version,
     inputs: inputs,
     outputs: outputs,
     nLockTime: this.nLockTime
   };
+  if (this._changeScript) {
+    obj.changeScript = this._changeScript.toString();
+  }
+  if (!_.isUndefined(this._changeIndex)) {
+    obj.changeIndex = this._changeIndex;
+  }
+  if (!_.isUndefined(this._fee)) {
+    obj.fee = this._fee;
+  }
+  return obj;
 };
 
 Transaction.prototype.fromObject = function(transaction) {
   var self = this;
   _.each(transaction.inputs, function(input) {
-    if (input.output && input.output.script) {
-      input.output.script = new Script(input.output.script);
-      if (input.output.script.isPublicKeyHashOut()) {
-        self.addInput(new Input.PublicKeyHash(input));
-        return;
-      } else if (input.output.script.isScriptHashOut() && input.publicKeys && input.threshold) {
-        self.addInput(new Input.MultiSigScriptHash(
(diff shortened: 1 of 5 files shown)


maraoz:
updated

eordano:
We should add a method to check if the transaction is valid (based only on the nLockTime and only if it's a timestamp). We *could* add an estimated (height -> timestamp) conversion based on historical data and expected difficulty growth?

eordano:
Coverage decreased (-0.01%) to 97.1%

maraoz:
fixed coverage, ready for review :)

lib/transaction/transaction.js
```diff
@@ -269,14 +257,84 @@ Transaction.prototype.fromObject = function(transaction) {
...
+};
+
+Transaction.prototype._checkConsistency = function() {
+  if (!_.isUndefined(this._changeIndex)) {
+    $.checkState(this._changeScript);
+    $.checkState(this.outputs[this._changeIndex]);
+    $.checkState(this.outputs[this._changeIndex].script.toString() ===
+      this._changeScript.toString());
+  }
+  // TODO: add other checks
```
braydonf: What other checks? An issue might be better, and more descriptive.

maraoz: yes, was planning to open an issue, thanks for the reminder :)

test/transaction/transaction.js
```diff
@@ -120,6 +120,8 @@ describe('Transaction', function() {
     script: Script.buildPublicKeyHashOut(fromAddress).toString(),
     satoshis: 100000
   };
+  var weirdUtxoWith100000Satoshis = JSON.parse(JSON.stringify(simpleUtxoWith100000Satoshis));
+  weirdUtxoWith100000Satoshis.script = new Script().add('OP_TRUE');
```
braydonf: `var anyoneCanSpendUtxo = `?

test/transaction/transaction.js
```diff
@@ -407,6 +436,66 @@ describe('Transaction', function() {
...
+        return new Transaction().lockUntilDate(499999999);
+      }).to.throw(errors.Transaction.LockTimeTooEarly);
+    });
+    it('fails if the block height is negative', function() {
+      expect(function() {
+        return new Transaction().lockUntilBlockHeight(-1);
+      }).to.throw(errors.Transaction.NLockTimeOutOfRange);
+    });
+  });
+  it('handles weird output', function() {
```
braydonf: Same as above, "handles an anyone-can-spend output" could be more descriptive.

lib/transaction/transaction.js
```diff
@@ -269,14 +257,84 @@ Transaction.prototype.fromObject = function(transaction) {
...
+};
+
+/**
+ * Sets nLockTime so that transaction is not valid until the desired date(a
+ * timestamp in seconds since UNIX epoch is also accepted)
+ *
+ * @param {Date | Number} time
+ * @return {Transaction} this
```
braydonf: These may need to be `@returns`

braydonf:
LGTM

eordano:
LGTM

lib/transaction/transaction.js
maraoz: apparently they're synonyms http://usejsdoc.org/tags-returns.html

maraoz:
updated with @braydonf comments

PR #1057, opened 2015-02-09
Update transaction.js

nathanph:
This is just confusing. =P


maraoz:
Thanks! Sorry about that :)

Issue #912, opened 2015-01-08
Error Message Enhancements


braydonf:
Error messages are repetitive, too long, and the main message of the error is half the way into the string.
```js
var a = new Address();
// bitcore.Error.InvalidArgument: Invalid Argument: TypeError: First argument is required, please include address data.
```
These could be readable, examples:
```js
// Bitcore Error: Invalid Argument: First argument is required, please include address data.
```
```js
// TypeError: First argument is required, please include address data.
```
```js
// First argument is required, please include address data.
```

braydonf:
Also, it may be helpful to add a link in the error message, such as: http://bitcore.io/api/address.html

yemel:
I like the option 2. I really like the idea of adding the documentations link on the error.

PR #1058, opened 2015-02-10
Added docs link to InvalidArgument message for preconditions.

braydonf:
- Error messages will appear as: "Invalid Argument: First argument is required, please include address data. Documentation: http://bitcore.io/guide/address.html"


braydonf:
Should partially close: https://github.com/bitpay/bitcore/issues/912 (Would need to be added to more of the standard messages)

PR #1068, opened 2015-02-12
check that there is no inconsistency between docs and bitcore version

maraoz:


eordano:
Is this WIP?

maraoz:
nope, why?

eordano:
Oh, saw the build was failing and I thought it was still in progress.

Merged

Issue #1054, opened 2015-02-08
Add an 'inspect' method for opcode.js


thofmann:
Would this make sense?

```javascript
/**
 * Will return a string formatted for the console
 *
 * @returns {String} Script opcode
 */
Opcode.prototype.inspect = function() {
  return '<Opcode: ' + this.toString() + ', hex: '+this.toHex()+', decimal: '+this.num+'>';
};
```

maraoz:
+1!

thofmann:
Closed by #1071

PR #1071, opened 2015-02-12
Add inspect method to Opcode.

thofmann:
Also added test for it, which is passing locally for me with mocha.


maraoz:
nice!
closes https://github.com/bitpay/bitcore/issues/1054

maraoz:
ACK

PR #1074, opened 2015-02-14
Cleanup/requirements

thofmann:
These are nonfunctional changes. I only changed some of the requirement variable names for consistency.

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -2,7 +2,7 @@
 
 var _ = require('lodash');
 var $ = require('./util/preconditions');
-var base58check = require('./encoding/base58check');
+var Base58Check = require('./encoding/base58check');
 var Networks = require('./networks');
 var Hash = require('./crypto/hash');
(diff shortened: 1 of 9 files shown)


braydonf:
The style guidelines look like they are conflicting on this one between [G0](https://github.com/bitpay/bitcore/blob/master/CONTRIBUTING.md#g0---default-to-felixges-style-guide) and [G6](https://github.com/bitpay/bitcore/blob/master/CONTRIBUTING.md#g6---naming-utility-namespaces).

eordano:
You're right @braydonf 

As G6 is quite specific, should we add a note stating that this is an exception to G0?

eordano:
@braydonf actually, felix's style guide doesn't mention anything about Namespaces (which is a more appropriate name for these utils). I think there's no conflict in that case

eordano:
Overall ACK; I'll merge because the coverall decrease doesn't make sense.

PR #1078, opened 2015-02-16
Add check for small fees

eordano:


maraoz:
LGTM. 
All these extra checks we're adding to `serialize` are really useful, but I'm afraid users will be confused when trying to build test transactions. Maybe we should add `"Use Transaction#uncheckedSerialize if you want to skip security checks"`. Spawning a separate issue for discussion: https://github.com/bitpay/bitcore/issues/1086

Issue #1080, opened 2015-02-16
createMultisign with empty publickeys should return error


matiu:
```
>> bitcore.Address.createMultisig([],3,'testnet');
<Address: 2N96Zk9c46EdiDq53uPgkttYJZTSph9fY3W, type: scripthash, network: testnet>
```

thofmann:
Wouldn't it make sense to return an error whenever `publicKeys.length` is less than `threshold` (which includes your given case)?

PR #1083, opened 2015-02-18
Transaction: Check that threshold is less than amount of signatures

eordano:


eordano:
Closes #1080

lib/transaction/transaction.js
```diff
@@ -447,6 +447,8 @@ Transaction.prototype._fromNonP2SH = function(utxo) {
 };
 
 Transaction.prototype._fromMultisigUtxo = function(utxo, pubkeys, threshold) {
+  $.checkArgument(threshold <= pubkeys.length,
```
braydonf: Shouldn't this be "Threshold must be less than or equal to the number of publickeys"? If there are 5 pubkeys, the threshold should be 5 and lower, not greater (e.g. 1-of-5 signatures required or 5-of-5) This check may need to also be at: https://github.com/bitpay/bitcore/blob/master/lib/address.js#L278

maraoz: Agreed with @braydonf on error message, and about the need of changing `address.js` to solve issue #1080.
To be sure, I'd add a test with @matiu's example, asserting it fails (see https://github.com/bitpay/bitcore/issues/1080)

test/transaction/transaction.js
```diff
@@ -401,6 +401,17 @@ describe('Transaction', function() {
         transaction.addInput(new Transaction.Input(), Script.empty(), 0);
       }).to.not.throw();
     });
+    it('does not allow a threshold number greater than the amount of public keys', function() {
+      expect(function() {
+        transaction = new Transaction();
+        return transaction.from({
+          txId: '0000000000000000000000000000000000000000000000000000000000000000',
+          outputIndex: 0,
+          script: Script.buildMultisigOut([public1, public2], 2).toScriptHashOut(),
```
maraoz: shouldn't this work? 2 public keys, threshold of 2, isn't this a normal 2-of-2?

test/transaction/transaction.js
```diff
@@ -401,6 +401,17 @@ describe('Transaction', function() {
...
+          satoshis: 10000
+        }, [], 1);
+      }).to.throw(errors.InvalidArgument);
```
maraoz: try to specify an error message to make this test more robust. i.e:
```
.to.throw('Threshold must be less than or equal to the number of publickeys');
```

test/transaction/transaction.js
eordano: This should be a different test (the call to `from` is expected to fail, not the creation of the multisig)

lib/transaction/transaction.js
eordano: Fixed

test/transaction/transaction.js
eordano: I changed that for an empty script to make it clearer

test/transaction/transaction.js
eordano: Done

test/transaction/transaction.js
maraoz: aaah, gotcha

maraoz:
ACK, waiting for travis

Issue #1079, opened 2015-02-16
script.isDataOut should return true for data with lengths from 0 to 80


braydonf:
See: https://github.com/bitcoin/bitcoin/pull/5286

Currently anything longer than 40 returns false.

PR #1084, opened 2015-02-18
Script: Update max length for standard opreturn to 80

eordano:


eordano:
Closes #1079

braydonf:
LGTM

lib/script/script.js
```diff
@@ -46,6 +46,8 @@ var Script = function Script(from) {
   }
 };
 
+Script.OP_RETURN_STANDARD_SIZE = 80;
```
braydonf: Should this go near the other constants that are defined on Script?

eordano: Updated

PR #1100, opened 2015-02-20
WIP: Add test for zero fee bug in transaction fee calculation.

braydonf:


braydonf:
Looks like this is only possible when using `toObject` and `fromObject`.

maraoz:
yeah, serialization removes input amount info :(
I don't have any ideas on how to make this work without complicating the API or querying a blockchain API

Issue #1082, opened 2015-02-17
Add MerkleBlock constructor


braydonf:
As part of BIP37, add a MerkleBlock constructor as is needed for a bitcore-p2p "merkleblock" message type. We can add it in bitcore-p2p, though it may make sense to have it in Bitcore along with Block and BlockHeader.

maraoz:
+1, maybe put this inside a `lib/block/` folder?

throughnothing:
It seems like just `lib/merkleblock.js` would be fine since there is `lib/blockheader.js` and `lib/block.js`, but either way works.

PR #1102, opened 2015-02-20
Add MerkleBlock

throughnothing:
Relates to #1082


throughnothing:
Still WIP

maraoz:
added "WIP" to title, please remove it when ready.
btw, great work! :)

throughnothing:
Thanks, will do from now on.

test/merkleblock.js
```diff
@@ -0,0 +1,134 @@
+'use strict';
+
+var bitcore = require('..'),
```
braydonf: One variable per var statement: https://github.com/felixge/node-style-guide#declare-one-variable-per-var-statement

throughnothing: Oops, thanks.

throughnothing:
I'm thinking this may work for the interface to check for a transaction in a `MerkleBlock`.  Thoughts?

```javascript
/**
 * @param {Transaction|String} - Transaction or hex string hash of transaction
 * @returns {Bool} - True/False whether this MerkleBlock contains the given TX
 */
MerkleBlock.prototype.containsTx = function containsTx() {
  ...
};
```

braydonf:
`MerkleBlock.prototype.hasTransaction` ?

braydonf:
If merkleblock.js is going into a block directory, I think block.js and blockheader.js should as well. This should be possible without an API change, though we may want to bump to 0.11 in case files are loaded directly.

throughnothing:
Hmmm, now I'm actually thinking the interface should maybe be:

````javascript
// This method needs no args, b/c the block can be self-verified.
// Relevant Transaction ID's needed to validate are already in this.hashes[]
MerkleRoot.prototype.isValid = function isValid() {
  ...
}

MerkleRoot.prototype.hasTransaction = function hasTransaction(transactionId) {
    return this.hashes.indexOf[transactionId] !== -1;
}
```

A consumer of this would do something like:

``` javascript
var mb = MerkleBlock.fromBuffer(...);
if(mb.valid() && mb.hasTransaction(myTransaction)) {
    // Do stuff
}
```

I think this is better because you often want to check that a `MerkleBlock` is valid, but without giving it a transaction ID, etc..  You also may want to check the `MerkleBlock` to see if it has a specific transaction by ID.  These are 2 separate actions, imo, and should not be combined into 1 via the `hasTransaction()` method doing both.

In my example above, the `hasTransaction(transactionId)` method could also itself call `isValid()` such as:

```javascript
MerkleRoot.prototype.hasTransaction = function hasTransaction(transactionId) {
  return this.isValid() && this.hashes.indexOf[transactionId] !== -1;
}
```

I don't think that is as good, because it's combining 2 functionalities into one, but it does make the `hasTransaction()` call "safer" for clients, if they forget to check that it's valid, and assume the `MerkleBlock` is valid just because it has the transaction they asked about.

Thoughts?

throughnothing:
Possibly in line with the `validProofOfWork()` method name on `BlockHeader`, the `isValid()` method here for a `MerkleBlock` could be named something like `validMerkleTree()`, but I think `isValid()` is pretty short, clear, and makes total sense.

braydonf:
`isValid` is often used in Bitcore with arguments, and not on the prototype (e.g. https://github.com/bitpay/bitcore/blob/master/lib/address.js#L433 https://github.com/bitpay/bitcore/blob/master/lib/publickey.js#L338), so I don't think it should be used for validating an instance. However `valid*` is used for validation checks on the instance.  (e.g. `validProofOfWork`, `validMerkleRoot`, `validTimestamp`) 

`hasTransaction` should only check that the transaction exists in the block, there are a number of other validations that need to be done. Even `validProofOfWork` isn't complete since the bits need to be compared with the previous block in the chain, where it's recalculated every 2016 blocks. So `valid` really can't be implemented without the context of a chain.

index.js
```diff
@@ -29,8 +29,9 @@ bitcore.errors = require('./lib/errors');
 
 // main bitcoin library
 bitcore.Address = require('./lib/address');
-bitcore.Block = require('./lib/block');
-bitcore.BlockHeader = require('./lib/blockheader');
+bitcore.Block = require('./lib/block/block');
```
braydonf: What about?

/lib/block/index.js (Block)
/lib/block/header.js (BlockHeader)
/lib/block/merkle.js (MerkleBlock)

throughnothing: Yup I thought of that when making the change but wasn't sure which would be preferable.  I'll make that change.  Thanks.

throughnothing:
@braydonf thanks for the thoughtful response.

I'm unclear what the final resolution is though.  Should I go with `validMerkleTree()`, and `hasTransactiom(tx)` as the remaining methods needed?

lib/block/merkleblock.js
```diff
@@ -0,0 +1,262 @@
...
+ */
+MerkleBlock.prototype.toJSON = function toJSON() {
+  return JSON.stringify(this.toObject());
+};
+
+/**
+ * Verify that the MerkleBlock is valid
+ * @returns {Bool} - True/False whether this MerkleBlock is Valid
+ */
+MerkleBlock.prototype.validMerkleTree = function validMerkleTree() {
```
throughnothing: Would love some feedback on the name of this method, plus the style of the code (with the named functions inside, etc.).  Not sure if that's the style we want to follow as I couldn't find other good examples of recursive functions being used in this codebase.  I can move those to top-level private prototype functions if necessary (i'm talking about `traverse`, and `calcTreeWidth`).

lib/block/merkleblock.js
```diff
@@ -0,0 +1,285 @@
...
+MerkleBlock.prototype.hasTransaction = function hasTransaction(tx) {
+  $.checkArgument(!_.isUndefined(tx), 'No transaction given');
+  $.checkArgument(tx instanceof Transaction
+    || typeof tx === 'string', 'No transaction given');
+
+  var hash = tx;
+  if(tx instanceof Transaction) {
+    hash = tx.id;
+  }
+  var revHash = BufferUtil.reverse(new Buffer(hash,'hex')).toString('hex');
```
throughnothing: Wasn't sure if I should check normal + reversed version of the hash.  The reason is, the hashes in the `MerkleBlock` will be reversed, so I could default to just reversing what this method gets in always, but it may be nice to be able to pass them in either way?

lib/block/merkleblock.js
```diff
@@ -0,0 +1,280 @@
...
+       */
+      hashes: arg.hashes,
+      /**
+       * @name MerkleBlock#flags
+       * @type {Number[]}
+       */
+      flags: arg.flags
+    };
+  } else {
+    throw new TypeError('Unrecognized argument for Block');
```
maraoz: Shouldn't this be `'Unrecognized argument for MerkleBlock'`?

lib/block/merkleblock.js
```diff
@@ -0,0 +1,280 @@
...
+  return JSON.stringify(this.toObject());
+};
+
+/**
+ * Verify that the MerkleBlock is valid
+ * @returns {Bool} - True/False whether this MerkleBlock is Valid
+ */
+MerkleBlock.prototype.validMerkleTree = function validMerkleTree() {
+  $.checkState(this.flags instanceof Array, 'MerkleBlock flags is not an array');
+  $.checkState(this.hashes instanceof Array, 'MerkleBlock flags is not an array');
```
maraoz: - Please use `_.isArray(this.flags)`
- Second error message should be `'MerkleBlock hashes is not an array`

lib/block/merkleblock.js
```diff
@@ -0,0 +1,280 @@
...
+  var self = this;
+  if(this._validMerkleTree === true) {
+    return true;
+  } else if (this._validMerkleTree === false) {
+    return false;
+  }
```
maraoz: I'd refactor this to 
```
// if result is cached, return it
if (_.isBoolean(this._validMerkleTree)) {
  return this._validMerkleTree;
}
```
for clarity (took me some time to understand this was caching for a boolean)

lib/block/merkleblock.js
```diff
@@ -0,0 +1,280 @@
...
+    height++;
+  }
+
+  var flagBitsUsed = 0;
+  var hashesUsed = 0;
+
+  // Modeled after Bitcoin Core merkleblock.h CalcTreeWidth()
+  function calcTreeWidth(height) {
+    return (self.numTransactions + (1 << height) - 1) >> height;
+  }
```
maraoz: please extract this function outside of `MerkleBlock.prototype.validMerkleTree` so it's not redefined on each call.

lib/block/merkleblock.js
```diff
@@ -0,0 +1,280 @@
...
+      return Hash.sha256sha256(new Buffer.concat([left, right]));
+    }
+  }
+
+  var root = traverse(height, 0);
+  if(hashesUsed !== this.hashes.length) {
+    return this._setValidMerkleTree(false);
+  }
+  return this._setValidMerkleTree(BufferUtil.equals(root, this.header.merkleRoot));
+}
```
maraoz: I don't understand the use of `flagBitsUsed`, given it apparently is always 0. Same as above, can you extract this outside `MerkleBlock.prototype.validMerkleTree`?

lib/block/merkleblock.js
maraoz: I'm fine with accepting only the MerkleBlock version.

lib/block/merkleblock.js
```diff
@@ -0,0 +1,280 @@
...
+ */
+MerkleBlock.prototype.hasTransaction = function hasTransaction(tx) {
+  $.checkArgument(!_.isUndefined(tx), 'No transaction given');
+  $.checkArgument(tx instanceof Transaction
+    || typeof tx === 'string', 'No transaction given');
+
+  var hash = tx;
+  if(tx instanceof Transaction) {
+    hash = tx.id;
+  }
```
maraoz: given the preconditions above, this is equivalent to:
```
var hash = tx.id || tx;
```

lib/block/merkleblock.js
```diff
@@ -0,0 +1,280 @@
...
+
+/**
+ * @param {Transaction|String} - Transaction or Transaction ID Hash
+ * @returns {Bool} - return true/false if this MerkleBlock has the TX or not
+ * @private
+ */
+MerkleBlock.prototype.hasTransaction = function hasTransaction(tx) {
+  $.checkArgument(!_.isUndefined(tx), 'No transaction given');
+  $.checkArgument(tx instanceof Transaction
+    || typeof tx === 'string', 'No transaction given');
```
maraoz: Please use different error messages for different errors. This allows for better testing. I'd use `'Wrong transaction type given, use "Transaction" or "string"`

maraoz:
Some comments addded. Really good contribution. The code style, design decisions, and implementation are all excellent. Thanks a lot @throughnothing !!! :+1: :+1:

lib/block/merkleblock.js
braydonf: We'll also want to check that the merkleRoot matches that of the block header, or is that a separate call?

lib/block/merkleblock.js
braydonf: Considering that the hashes may change, should the value be cached?

lib/block/merkleblock.js
braydonf: It may not always be zero: https://github.com/bitcoin/bips/blob/master/bip-0037.mediawiki#Constructing_a_partial_merkle_tree_object

maraoz:
fixes https://github.com/bitpay/bitcore/issues/1082

lib/block/merkleblock.js
throughnothing: Yes, `flagBitsUsed` is definitely not always zero, note how it is `++`'ed on almost ever `traversal` iteration.

throughnothing:
Thanks @maraoz, lots of great comments/finds.  Will Fix this up shortly.

lib/block/merkleblock.js
throughnothing: @braydonf I don't think the hashes would change in a single instance.  Why/how would they?

lib/block/merkleblock.js
throughnothing: It's incremented on line 182

lib/block/merkleblock.js
throughnothing: Should I do the same with the traverse function below?

lib/block/merkleblock.js
throughnothing: Hmm, I think I'm doing that on line 176:

```javascript
  return this._setValidMerkleTree(BufferUtil.equals(root, this.header.merkleRoot));
```

throughnothing: In the `MerkleBlock`, you won't know if the client has access to the full block or not.  It would be up to the client to be verifying that this `MerkleBlock` fits into its chain somewhere.  You could come across a valid `MerkleBlock` for a valid chain that is no longer part of the longest chain I suppose, and it would still validate.

maraoz:
LGTM

throughnothing:
I'm not sure about that 1 Codacy failure.  It says unexpected use of `&` on `merkleblock.js` line 186, but I think the use of & there is valid and even necessary...

throughnothing:
It says:

```
Bitwise operators are very rare in JavaScript programs and quite often &amp; is simply a mistyped &amp;&amp;.
```

Heh, it's not very rare in what we're doing here :)  Maybe we can turn that check off/down in Codacy?

braydonf:
Bitwise should be fine: https://github.com/bitpay/bitcore/blob/master/.jshintrc#L2

throughnothing:
¯\_(ツ)_/¯

test/merkleblock.js
```diff
@@ -0,0 +1,192 @@
...
+
+    it('should validate good merkleblocks', function() {
+      data.JSON.forEach(function(json) {
+        var b = MerkleBlock(JSON.stringify(json));
+        b.validMerkleTree().should.equal(true);
+        b._validMerkleTree.should.equal(true);
+      });
+    });
+
+    it('should respect _validMerkleTrees', function() {
```
braydonf: The fact that this can return "true" when it could actually be "false" because of the cache value, seems a bit concerning.

lib/block/merkleblock.js
```diff
@@ -0,0 +1,284 @@
...
+ */
+MerkleBlock.prototype.hasTransaction = function hasTransaction(tx) {
+  $.checkArgument(!_.isUndefined(tx), 'tx cannot be undefined');
+  $.checkArgument(tx instanceof Transaction || typeof tx === 'string',
+      'Invalid tx given, tx must be a "string" or "Transaction"');
+
+  var hash = tx.id || tx;
+  var revHash = BufferUtil.reverse(new Buffer(hash,'hex')).toString('hex');
+
+  return (this.hashes.indexOf(hash) !== -1
```
braydonf: Hashes also include merkle nodes from the partial merkle tree: https://bitcoin.org/en/developer-reference#merkleblock

throughnothing: I know.  Are you worried about collisions with txIDs ?

test/merkleblock.js
throughnothing: After thinking through it, I think you've convinced me that this is an over-optimization for now.  Better and safer to not do this until we understand more about how it's used.  Will remove.

lib/block/merkleblock.js
throughnothing: You have to parse the MerkleTree to get the ones that are TXID's.  The only good solution I can think of is that we'd have to parse the tree in the constructor, and have a `this.transactions[]` array.  But then, would the constructor fail if the MerkleBlock doesn't validate?  Do users not have to call the `validMerkleTree()` method, and can just assume any `MerkleBlock` object that gets built is valid?

Or we could re-walk the tree every time `hasTransaction()` is called, and then only search amongst the ones that are transaction IDs.  We could cache the transactions array for future `hasTransaction()` calls, but it sounds like we might not like that.

I'm sure there is a better way?

braydonf: Yep. Also looking into expected byte order of the hashes.

throughnothing: Sweet, let me know your thoughts.

braydonf: That sounds like it should work, e.g.:
```javascript
var hashes = this.getTransactionHashes();
var exists = (hashes.indexOf(hash) !== -1);
return exists;
```

Not sure about caching, we can address that if it's an issue though. Also I think it makes sense for the validaty checks to be independent, so verifying that a transaction after verifying that the block fits in the chain (validate proof-of-work, timestamp, merkletree, etc) would only need to be: `if (merkleblock.hasTransaction(tx)) { ... };`

throughnothing: Well, the thing about that is that getting the Transaction hashes and verifying the tree are basically the same operation.  If the tree is invalid, I'm not even sure you could continue finding out which ones are transactions (what if the bit flag is wrong or doesn't make sense?).

braydonf: I see, that makes sense then.

braydonf: So it would be more similar to:

```javascript
var hashes = this.getTransactionHashes();
if (hashes) {
  return (hashes.indexOf(hash) !== -1);
}
return false;
```

lib/block/merkleblock.js
```diff
@@ -0,0 +1,269 @@
...
+    false;
+  }
+  return BufferUtil.equals(root, this.header.merkleRoot);
+}
+
+/** Traverse a the tree in this MerkleBlock, validating it along the way
+ *  Modeled after Bitcoin Core merkleblock.cpp TraverseAndExtract()
+ * @param {Number} - Current height
+ * @param {Number} - Current position in the tree
+ * @returns {Buffer|null} - Buffer containing the Merkle Hash far that height
```
braydonf: typo? "for"

lib/block/merkleblock.js
braydonf: Perhaps there is a "getTransactionHashesAndRoot" method, and getTransactionHashes and validMerkleTree both use, but return different values?

throughnothing: Huh, I think I like that.

lib/block/merkleblock.js
throughnothing: good catch

throughnothing:
@braydonf thanks for the help on this.  I think I came up with something a little better even than your last suggestion, and I also made `_traverseMerkleTree()` less dependent on object state utilizing pass by reference instead.

The only thing I maybe don't like is initializing the `opts` object for cases where callers don't set the initialization (`flagBitsUsed` and `hashBitsUsed` *have* to initialize to 0, though, and `txs` should be optional, as it returns the found txs if you want).

Thoughts?

maraoz:
@throughnothing Don't worry about the Codacy fail: we need to deactivate that check, it's not helping.

Regarding the `opts` initialization, I think it's fine. That's a common issue with recursive functions, and it's not really a problem. A usual way to deal with this is to create another function that calls the recursive function, with the correct initialization opts. i.e:
```
MerkleBlock.prototype._getMerkleTreeRoot = function() {
  var height = this._calcTreeHeight();
  var opts = { hashesUsed: 0, flagBitsUsed: 0 };
  var root = this._traverseMerkleTree(height, 0, opts);
  if(opts.hashesUsed !== this.hashes.length) {
    return null;
  }
  return root;
};
```
and in https://github.com/bitpay/bitcore/pull/1102/files#diff-28653da22e62345088cc538f5e234609R159
```
var root = this._getMerkleTreeRoot();
```

maraoz:
In any case, I'm fine to merge this current PR. You have my ACK (If anyone else agrees, he can merge it)

throughnothing:
Thanks @maraoz !

throughnothing:
For the Codacy checks, maybe you can leave them on, but not have them fully fail a PR.  Some of it's stuff is useful to see.

braydonf:
Incoming PR (with additional test): https://github.com/throughnothing/bitcore/pull/1/files

braydonf:
ACK

maraoz:
amazing work @throughnothing !

throughnothing:
Thanks @maraoz and thanks for all the help @braydonf !

PR #1103, opened 2015-02-21
Fix JS doc for publickey.js

dskloet:


maraoz:
LGTM

dskloet:
This is the first time I make a pull request (for any project). Is there anything else I need to do?

braydonf:
Seems like we may have a mix between Google Closure Compiler syntax optional params and the JSDoc syntax: http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

We may want to pick one and use that consistantly.

dskloet:
I also noticed that you were using String instead of string while the primitive JS type is called string (lower case). But I didn't touch that since it at least seemed to be consistent.

eordano:
Congratz on your first PR!

eordano:
@braydonf: I opened https://github.com/bitpay/bitcore/issues/1108 to discuss that

dskloet:
Thanks :)

Issue #1095, opened 2015-02-19
Addresses from Testnet HDPrivateKey are Livenet


braydonf:
```javascript
var hdprivatekey = bitcore.HDPrivateKey.fromString('tprv...');
var address = hdprivatekey.derive(0).publicKey.toAddress();
```
The above will generate an livenet address, even though the hdprivatekey was given a testnet extended key.

PR #1104, opened 2015-02-23
Fix bug on HDKeys not returning the correct network when requesting a PrivateKey/PublicKey from the HDKey

eordano:
See #1095


test/hdpublickey.js
```diff
@@ -31,7 +33,7 @@ describe('HDPublicKey interface', function() {
       got = e instanceof errorType;
     }
     if (!error instanceof errorType) {
-      console.log('Adsasd', typeof error);
+      console.log('Error', typeof error);
```
maraoz: nice catch hehe

maraoz:
fixes #1095

PR #1107, opened 2015-02-23
Fix an error message on fee calculation

eordano:


lib/transaction/transaction.js
```diff
@@ -145,7 +145,7 @@ Transaction.prototype._validateFees = function() {
     return 'Fee is more than ' + Transaction.FEE_SECURITY_MARGIN + ' times the suggested amount';
   }
   if (this._getUnspentValue() < this._estimateFee() / Transaction.FEE_SECURITY_MARGIN) {
-    return 'Fee is less than ' + Transaction.FEE_SECURITY_MARGIN + ' times the suggested amount';
+    return 'Fee is ' + Transaction.FEE_SECURITY_MARGIN + ' times less than the suggested amount';
```
maraoz: Still confusing. Why don't we return something like `'Fee too low: expected X but found Y'`?

braydonf: "Fee to low: expected more than X but found Y" ? In the case that the fee is too high: "Fee is too high: expected less than X but found Y" ?

maraoz:
ACK

PR #1124, opened 2015-02-27
increase coverage for Input and Output, fix some bugs

maraoz:


lib/transaction/input/input.js
```diff
@@ -57,21 +64,19 @@ Input.prototype.toObject = function toObject() {
...
-  }
-  return new Input({
-    output: json.output ? new Output(json.output) : undefined,
-    prevTxId: json.prevTxId || json.txidbuf,
-    outputIndex: _.isUndefined(json.outputIndex) ? json.txoutnum : json.outputIndex,
-    sequenceNumber: json.sequenceNumber || json.seqnum,
-    scriptBuffer: new Script(json.script, 'hex')
-  });
+  $.checkArgument(JSUtil.isValidJSON(json), 'Invalid JSON provided to Input.fromJSON');
+  return Input.fromObject(JSON.parse(json));
```
maraoz: best part of the PR

test/transaction/output.js
```diff
@@ -47,8 +56,30 @@ describe('Output', function() {
...
+      'cabd5ac1ca10646e23fd5f51508 21 0x038282263212c609d9ea2a6e3e172de23' +
+      '8d8c39cabd5ac1ca10646e23fd5f51508 OP_2 OP_CHECKMULTISIG OP_EQUAL')
+  });
+  it('toBufferWriter', function() {
+    output2.toBufferWriter().toBuffer().toString('hex')
+      .should.equal('00ab904100000000485215038282263212c609d9ea2a6e3e172de2' +
+        '38d8c39cabd5ac1ca10646e23fd5f5150815038282263212c609d9ea2a6e3e172d' +
+        'e238d8c39cabd5ac1ca10646e23fd5f5150852ae87');
+  });
+  it('to/from JSON', function() {
```
eordano: Please don't name tests plainly with the name of the method you're testing, so that the tests become more semantic. I'd suggest this test for example to be called `to/from JSON roundtrip works correctly`, with is what the test is actually looking for.

test/transaction/input/input.js
```diff
@@ -41,6 +49,39 @@ describe('Transaction.Input', function() {
...
+      var input = Input.fromJSON(coinbaseJSON);
+      var obj = input.toObject();
+      obj.script = new Buffer(obj.script, 'hex');
+      Input.fromObject(obj).should.deep.equal(input);
+      obj.script = 42;
+      Input.fromObject.bind(null, obj).should.throw('Invalid argument type: script');
+    });
+  });
+
+  it('_estimateSize', function() {
```
eordano: Shouldn't this fail with a `NonImplementedError` or something?

Same as https://github.com/bitpay/bitcore/pull/1124/files#r25592034, please use a more descriptive name.

If this were the test for `PublicKeyHashInput`, I'd use something like `_estimateSize returns the maximum size a scriptSig for a P2PKH`

maraoz:
updated with comments

Issue #1147, opened 2015-03-14
Performance issue at HD Private Key


matiu:
Precalculating HDPublicKey on `.derive` that HDPrivate key
(https://github.com/bitpay/bitcore/blob/master/lib/hdprivatekey.js#L426)
makes the derivation ~60% slower (10s vs. 24s in my systems for 1000 priv key derivations).

Maybe those fields could be calculated on request (there are not always needed)

Here is the test code:
``` js
#!/usr/bin/env node

var b = require('./index');

var i = 0,
  n = 1000;

console.log('Starting'); //TODO
console.time('s');
var h = new b.HDPrivateKey();
while (i++ < n) {
  h.derive('m/' + i);
  if (!(i % 50)) console.log(i);
};

console.timeEnd('s');
```

maraoz:
https://github.com/bitpay/bitcore/pull/1149 fixes this and improves performance from 10 seconds to 4 seconds with the given code

PR #1149, opened 2015-03-16
lazy calc for xpubkey in HDPrivateKey

maraoz:
fixes https://github.com/bitpay/bitcore/issues/1147

### DIFF
--- a/test/hdkeys.js
+++ b/test/hdkeys.js
@@ -76,6 +76,11 @@ describe('BIP32 compliance', function() {
     publicKey.toString().should.equal(publicKey.xpubkey);
   });
 
+  it('cache for xpubkey works for test vector 1', function() {
+    var pk = HDPrivateKey(vector1_m_private);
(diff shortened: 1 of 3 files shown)


eordano:
Looks great, but is the test specific enough? Maybe it would be better to check for the internal property...

test/hdkeys.js
```diff
@@ -76,6 +76,11 @@ describe('BIP32 compliance', function() {
     publicKey.toString().should.equal(publicKey.xpubkey);
   });
 
+  it('cache for xpubkey works for test vector 1', function() {
+    var pk = HDPrivateKey(vector1_m_private);
+    pk.xpubkey.should.equal(pk.xpubkey);
```
eordano: Add this line before please: `expect(pk._hdPublicKey).to.be.undefined`; so this test fails before the commit

maraoz: done

PR #1151, opened 2015-03-18
refactor Address<->Script relation

maraoz:
- adds `Address.payingTo(script)`, which returns a p2sh address paying to that script
- Address.fromScript now works for p2sh in, p2sh out, p2pkh in, and p2pkh out, and returns the Address that can be found inside the script.
- 100% coverage on Address
- fix bug in `Script.isScriptHashIn` I found

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -244,28 +244,11 @@ Address._transformPublicKey = function(pubkey) {
  * @private
  */
 Address._transformScript = function(script, network) {
-  var info = {};
-  if (!(script instanceof Script)) {
-    throw new TypeError('script must be an instance of Script.');
-  }
-  if (script.isScriptHashOut()) {
-    info.hashBuffer = script.getData();
-    info.type = Address.PayToScriptHash;
-  } else if (script.isPublicKeyHashOut()) {
-    info.hashBuffer = script.getData();
-    info.type = Address.PayToPublicKeyHash;
-  } else if (script.isPublicKeyHashIn()) {
-    // hash the publickey found in the scriptSig
-    info.hashBuffer = Hash.sha256ripemd160(script.chunks[1].buf);
-    info.type = Address.PayToPublicKeyHash;
-  } else if (script.isScriptHashIn()) {
-    // hash the redeemscript found at the end of the scriptSig
-    info.hashBuffer = Hash.sha256ripemd160(script.chunks[script.chunks.length - 1].buf);
-    info.type = Address.PayToScriptHash;
-  } else {
+  $.checkArgument(script instanceof Script, 'script must be a Script instance');
+  var info = script.getAddressInfo(network);
+  if (!info) {
     throw new errors.Script.CantDeriveAddress(script);
   }
-  info.network = Networks.get(network) || Networks.defaultNetwork;
   return info;
 };
 
@@ -297,7 +280,7 @@ Address.createMultisig = function(publicKeys, threshold, network) {
  */
 Address._transformString = function(data, network, type) {
   if (typeof(data) !== 'string') {
-    throw new TypeError('Address supplied is not a string.');
+    throw new TypeError('data parameter supplied is not a string.');
   }
   var addressBuffer = Base58Check.decode(data);
   var info = Address._transformBuffer(addressBuffer, network, type);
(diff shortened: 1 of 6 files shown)


lib/address.js
```diff
@@ -255,9 +254,16 @@ Address._transformScript = function(script, network){
...
     info.type = Address.PayToScriptHash;
+  } else if (script.isPublicKeyHashIn()) {
+    // hash the publickey found in the scriptSig
+    info.hashBuffer = Hash.sha256ripemd160(script.chunks[1].buf);
+    info.type = Address.PayToPublicKeyHash;
+  } else {
+    throw new errors.Script.CantDeriveAddress(script);
   }
   info.network = Networks.get(network) || Networks.defaultNetwork;
   return info;
```
maraoz: changes to `_transformScript` are the biggest change in the PR

eordano:
LGTM. I give up on trying to remove the coupling between PublicKey <-> Address <-> Script (I'd like this to be a one-way thing). Nice PR :D

braydonf:
I think it would be better to put more of the script logic (hashing chunks, etc) in `Script.toAddress`: https://github.com/bitpay/bitcore/blob/master/lib/script/script.js#L718

maraoz:
I found a bug testing this PR with network transactions, will move to WIP

lib/script/script.js
```diff
@@ -713,14 +716,19 @@ Script.fromAddress = function(address) {
...
-  if (this.isPublicKeyHashOut() || this.isScriptHashOut()) {
+  var canConvertToAddress = this.isPublicKeyHashOut() ||
+    this.isScriptHashOut() ||
+    this.isPublicKeyHashIn() ||
+    this.isScriptHashIn();
+  if (canConvertToAddress) {
     return new Address(this, network);
   }
-  throw new Error('The script type needs to be PayToPublicKeyHash or PayToScriptHash');
+  return false;
```
braydonf: Returning a boolean sounds better, that way checking isScriptHashOut, etc isn't needed prior, and checking then checking for false instead. May be good to bump to v0.12.0.

maraoz: +1, this will require major version change, give I've changed the `Address.fromScript` contract

maraoz:
Just fixed some nasty bugs I uncovered with this PR, will consider @braydonf's suggestions now

eordano:
Note that this breaks the API. Next release must be a minor

maraoz:
ready again for review

lib/script/script.js
```diff
@@ -713,14 +727,44 @@ Script.fromAddress = function(address) {
...
+  var info = {};
+  if (this.isScriptHashOut()) {
+    info.hashBuffer = this.getData();
+    info.type = Address.PayToScriptHash;
+  } else if (this.isPublicKeyHashOut()) {
+    info.hashBuffer = this.getData();
+    info.type = Address.PayToPublicKeyHash;
+  } else if (this.isPublicKeyHashIn()) {
+    // hash the publickey found in the scriptSig
+    info.hashBuffer = Hash.sha256ripemd160(this.chunks[1].buf);
```
eordano: Should `this.chunks[1].buf` be `this.getData()?`

lib/script/script.js
```diff
@@ -713,14 +727,44 @@ Script.fromAddress = function(address) {
...
+    info.type = Address.PayToPublicKeyHash;
+  } else if (this.isScriptHashIn()) {
+    // hash the redeemscript found at the end of the scriptSig
+    info.hashBuffer = Hash.sha256ripemd160(this.chunks[this.chunks.length - 1].buf);
```
eordano: ditto?

lib/script/script.js
maraoz: Thought about it, but I think it would be weird to return the hash of the public key, given the input script in p2pkh case contains a public key AND a signature... I mean, why would getData only return one of those 2, and hashed?

lib/script/script.js
maraoz: same here... a p2sh input script has a LOT more data than the hash of the redeem script... why would `getData` return that?

braydonf:
LGTM

lib/script/script.js
eordano: Sounds good!

maraoz: Thanks for the merge!

PR #1152, opened 2015-03-20
WIP add genesis block hashes

maraoz:
Add genesis block hashes

### DIFF
--- a/lib/networks.js
+++ b/lib/networks.js
@@ -71,7 +71,8 @@ function addNetwork(data) {
     xprivkey: data.xprivkey,
     networkMagic: BufferUtil.integerAsBuffer(data.networkMagic),
     port: data.port,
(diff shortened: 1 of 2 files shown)


lib/networks.js
```diff
@@ -103,7 +105,17 @@ addNetwork({
     'seed.bitcoinstats.com',
     'seed.bitnodes.io',
     'bitseed.xf2.org'
-  ]
+  ],
+  genesis: new Buffer('010000000000000000000000000000000000000000000000000000000000' +
```
eordano: This is the whole block, but for testnet you used only the hash. Please pick one (I'd pick the hash)

maraoz: I'm using this as a support branch for a PR to `bitcore-node`. I'll add WIP to the title

PR #1154, opened 2015-03-24
Fix bug if no value is supplied.

martindale:
This fixes a minor bug in some corner cases where multiple network changes may take place over time, or specifically, if the network is undefined.

### DIFF
--- a/lib/networks.js
+++ b/lib/networks.js
@@ -75,7 +75,7 @@ function addNetwork(data) {
   });
 
   _.each(_.values(network), function(value) {
-    if (value && !_.isObject(value)) {
+    if (!_.isUndefined(value) && !_.isObject(value)) {
       networkMaps[value] = network;
     }
   });
--- a/lib/networks.js
+++ b/lib/networks.js
@@ -75,7 +75,7 @@ function addNetwork(data) {
   });
 
   _.each(_.values(network), function(value) {
-    if (!_.isObject(value)) {
+    if (value && !_.isObject(value)) {
       networkMaps[value] = network;
     }
   });
--- a/test/networks.js
+++ b/test/networks.js
@@ -20,7 +20,7 @@ describe('Networks', function() {
       pubkeyhash: 0x10,
       privatekey: 0x90,
       scripthash: 0x08,
-      xpubkey:  0x0278b20e,
+      xpubkey: 0x0278b20e,
       xprivkey: 0x0278ade4,
       networkMagic: 0xe7beb4d4,
       port: 20001,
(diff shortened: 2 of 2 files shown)


test/networks.js
```diff
@@ -40,6 +40,25 @@ describe('Networks', function() {
...
+      xprivkey: 0x0278ade5,
+      networkMagic: 0xe7beb4d5,
+      port: 20008,
+      dnsSeeds: [
+        'somenet.localhost'
+      ]
+    };
+    networks.add(custom);
+    var network = networks.get(undefined);
+    should.not.exist( network );
```
braydonf: whitespace `should.not.exist(network)`

test/networks.js
```diff
@@ -40,6 +40,25 @@ describe('Networks', function() {
...
     }
   });
+  
+  it('should not set a network map for an undefined value', function() {
+    var custom = {
+      name: 'somenet',
+      pubkeyhash: 0x13,
+      privatekey: 0x93,
+      scripthash: 0x11,
+      xpubkey:  0x0278b20f,
```
braydonf: extra space here

lib/networks.js
```diff
@@ -75,7 +75,7 @@ function addNetwork(data) {
   });
 
   _.each(_.values(network), function(value) {
-    if (!_.isObject(value)) {
+    if (value && !_.isObject(value)) {
```
braydonf: `if (!_.isUndefined(value) && !_.isObject(value)) {`

to handle the case between `0` and `undefined`

maraoz:
@martindale tests seem to be broken. Please check. I've restarted travis just in case it's gone mad

braydonf:
ACK

test/networks.js
```diff
@@ -40,6 +40,25 @@ describe('Networks', function() {
...
+      xprivkey: 0x0278ade5,
+      networkMagic: 0xe7beb4d5,
+      port: 20008,
+      dnsSeeds: [
+        'somenet.localhost'
+      ]
+    };
+    networks.add(custom);
+    var network = networks.get(undefined);
+    should.not.exist(network);
```
maraoz: shouldn't `networks.get(undefined)` throw an exception?

braydonf: I don't think so currently, since there is code that is written like this:
`var network = Networks.get(argument) || Networks.defaultNetwork;` 
If argument is `undefined`, `undefined` is returned and then is set to the default.

maraoz: I see. I'll merge then

Issue #1157, opened 2015-03-26
Add docs for `transaction.getFee()` to mention that the method is not available once a transaction is serialized.


braydonf:
The `getFee()` method will only work when building a transaction in Bitcore. Once the transaction has been serialized, the unspent outputs for the inputs are no longer available and the fee can not be calculated. Documentation that mentions this could be useful here: https://github.com/bitpay/bitcore/blob/master/lib/transaction/transaction.js#L744

PR #1158, opened 2015-03-26
Update transaction.getFee() JSDoc

carnesen:
Attempt at resolving https://github.com/bitpay/bitcore/issues/1157, "Add docs for `transaction.getFee()` to mention that the method is not available once a transaction is serialized."


eordano:
Looks good! Thanks!

Issue #1164, opened 2015-03-30
bitcore-mnemonic (probably also other submodules)


fanatid:
```js
$ node
> Mnemonic = require('bitcore-mnemonic')
> bitcore = require('bitcore')
> pubKey = Mnemonic().toHDPrivateKey().publicKey
> address = bitcore.Address.fromPublicKey(pubKey)
TypeError: Address must be an instance of PublicKey.
> address = new bitcore.Address(pubKey)
Invalid Argument: Must provide a `hash` or `hashBuffer` property
```
If second error yet looks normal, first very confusing
`bitcore.Address` not recognized `PublicKey` because `bitcore-mnemonic` used own `bitcore` module...

Probably it's unsolvable problem and I think better way will be add information about this behaviour to `bitcore-mnemonic`

braydonf:
Something along the lines of "unexpected bitcore version detected" may be a better error if multiple bitcore versions are used.  This should be fixed with: https://github.com/bitpay/bitcore-mnemonic/pull/21 (as the versions of bitcore should match).

maraoz:
@fanatid thanks for reporting this in detail. Many developers found this problem (even internally at bitpay) and it's really annoying. We're working on a solution to better report this situation and try to minimize it's occurrence.

maraoz:
should be good to go if you use latest npm versions of both modules (released today). Error messages were improved too. Let me know if the problem persists.

fanatid:
Yes, now all good. Thanks!

PR #1165, opened 2015-03-30
Add guard to prevent more than one instance of bitcore

eordano:


eordano:
See:
* https://github.com/bitpay/bitcore/issues/1164
* https://github.com/bitpay/bitcore-message/issues/15

index.js
```diff
@@ -1,5 +1,14 @@
 var bitcore = module.exports;
 
+var inBrowser = typeof process === 'undefined' || typeof process.versions === 'undefined';
+if ((inBrowser && window._bitcore) || (!inBrowser && global._bitcore)) {
+  throw new Error('More than one instance of bitcore found');
+}
+if (inBrowser) {
+  window._bitcore = true;
```
maraoz: I'd save the bitcore version, so that the error can be more detailed

eordano: Addressed

fanatid:
But how this solve the problem? As I see, if I import `bitcore-mnemonic` first and then trying importing `bitcore` -- error will be raised.. I'm wrong?

eordano:
@fanatid we'll release an updated version of `bitcore-mnemonic` with this patch: https://github.com/bitpay/bitcore-message/commit/a328b3491744d38416614971beeee495e8ffb18f

maraoz:
ACK

olalonde:
This won't raise any error if bitcore is required first though right? (e.g.: `require ('bitcore');require('bitcore-message')`.

fanatid:
@olalonde as I understand, with new version of bitcore and other submodules, we should get
```bash
$ npm ls | grep bitcore
├─┬ bitcore@a.b.c
├── bitcore-mnemonic@a.b.c
```
instead
```bash
$ npm ls | grep bitcore
├─┬ bitcore@a.b.c
├─┬ bitcore-mnemonic@a.b.c1
│ └─┬ bitcore@a.b.c2
```

in first case only one bitcore will be used, so everything (I hope) should be fine, but we must be careful with versions, otherwise error will be raised

PR #1166, opened 2015-03-31
drop cached value for inputAmount and outputAmount

fanatid:
Approximate code for getting a bug:
```js
var bitcore = require('bitcore')
var tx = new bitcore.Transaction().change('...').from({...}).to('...', xxx).serialize()
```


maraoz:
ACK. many thanks for the contribution!

Issue #1171, opened 2015-04-04
network immutable


fanatid:
Probably [defineImmutable](https://github.com/bitpay/bitcore/blob/0938eadab5466b73de12bb020fb6a942934cef8f/lib/util/js.js#L62) will better than [_.extend](https://lodash.com/docs#assign) in [addNetwork](https://github.com/bitpay/bitcore/blob/4f4b2b93fbcbd71882bdc3ebd9f6278c487d1894/lib/networks.js#L67) ?

maraoz:
+1

PR #1185, opened 2015-04-11
Immutable network Object

fanatid:
Related with #1171 
I don't know what make with `networkMagic` (Buffer) and with `dnsSeeds` (Array), may be use [Object.freeze](http://mdn.io/object.freeze)? But may be in this case also use *Object.freeze* for network instead `JSUtil.defineImmutable`?
Also tests not looks strong... I should check all properties or name will be enough?

### DIFF
--- a/lib/networks.js
+++ b/lib/networks.js
@@ -2,6 +2,7 @@
 var _ = require('lodash');
 
 var BufferUtil = require('./util/buffer');
+var JSUtil = require('./util/js');
 var networks = [];
 var networkMaps = {};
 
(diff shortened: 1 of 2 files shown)


lib/util/js.js
```diff
@@ -60,11 +60,11 @@ module.exports = {
    * @return {Object} The target object
    */
   defineImmutable: function defineImmutable(target, values) {
-    Object.keys(values).forEach(function(key){
+    _.each(values, function(value, key) {
```
braydonf: lodash "each" will iterate over anything with a .length property, including strings. "values" here is expected to be an object, not an array or string.

fanatid:
thanks for remark, pr updated

maraoz:
LGTM

PR #1186, opened 2015-04-11
allowing non-standard P2SH/multisig

quartzjer:
First, I don't expect that this (trivial) patch is fit to be merged, but I thought a pull request might be a good place to have the discussion around an alternative interface to enable this usage pattern.

The challenge is, as of [0.10 now non-standard P2SH opcodes are valid](https://github.com/bitcoin/bitcoin/blob/v0.10.0/doc/release-notes.md#standard-script-rules-relaxed-for-p2sh-addresses), and there's not an obvious way to work with them in the bitcore transaction utilities.  An example of this is [here](https://github.com/quartzjer/pennybank/blob/master/test/bitcore_p2cm.test.js#L64) where I'm testing microtransaction-enabling scripts using a multisig with added conditions (hence this patch to the multisig input).


maraoz:
An interesting discussion to have, true.
I wouldn't do it the way you did, because you're adding an extra script to a multisig output builder method, which is somewhat confusing.
This is part of a bigger, long standing discussion about using custom scripts as inputs. 
Did you look at `transaction.addInput(input)`? We may need to think about a better interface for this, though. If I understand correctly, you used `_fromMultisigUtxo` because you wanted to have support for signing that input via the `Transaction#sign` method?

quartzjer:
Yep, how I did it was just a minimum to make it work, not ideal.

I think addInput may have changed since I last tried it moons ago so I need to check that path out again... and yes, the flow is identical to multisig except with some additional op's in the script :)

PR #1196, opened 2015-04-21
increase script test coverage

maraoz:
- small fix in documentation
- small fix in `Script._addByType`
- increase test coverage

### DIFF
--- a/lib/script/script.js
+++ b/lib/script/script.js
@@ -182,32 +182,40 @@ Script.fromString = function(str) {
   return script;
 };
 
+Script.prototype._chunkToString = function(chunk) {
+  var opcodenum = chunk.opcodenum;
+  var str = '';
+  if (!chunk.buf) {
+    // no data chunk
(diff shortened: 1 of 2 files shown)


lib/script/script.js
```diff
@@ -583,7 +591,7 @@ Script.buildMultisigOut = function(publicKeys, threshold, opts) {
  *
  * @param {PublicKey[]} pubkeys list of all public keys controlling the output
  * @param {number} threshold amount of required signatures to spend the output
- * @param {Array} signatures signatures to append to the script
+ * @param {Buffer[]} signatures and array of signature buffers to append to the script
```
braydonf: "Array" I think is accurate here, jsdoc won't parse this correctly with our documentation ([also see](https://github.com/bitpay/bitcore-mnemonic/commit/e530ff8b4d118fbbf033934f10466d804707f392#diff-da7cf14a5d0febb8b6aa96c9e244ce54R32))

maraoz: fixed

PR #1200, opened 2015-04-27
sighash: remove dead branch

dcousens:
Best I can tell, this branch would never be reached as:

`a > b - 1` is the same as `a >= b`.

Quick verification for those unsure: 

``` javascript
b = 5
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9].forEach(function(a) { console.log(a > b - 1, a >= b) })
```


maraoz:
ACK. nice catch.
Would love another developer's ACK before merging

braydonf:
ACK

braydonf:
For reference, relevant bitcoin core code:
- https://github.com/bitcoin/bitcoin/blob/f425050546644a36b0b8e0eb2f6934a3e0f6f80f/src/script/interpreter.cpp#L1036
- https://github.com/bitcoin/bitcoin/blob/164d7b674b49b3c062d0331f28d21ad4f6bf5207/src/test/sighash_tests.cpp#L26

Issue #1180, opened 2015-04-07
Block.toObject() has prevHash and merkleRoot reversed of what they should be


pnagurny:
For example, Block 1 on testnet: https://test-insight.bitpay.com/block/00000000b873e79784647a6c82962c70d228557d24a747ea4d1b8bbe878e1206

The prevHash should be 000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943.
The merkleRoot should be 4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b.

Instead it is coming out as:
prevHash: 43497fd7f826957108f4a30fd9cec3aeba79972084e90ead01ea330900000000
merkleRoot: bac8b0fa927c0ac8234287e33c5f74d38d354820e24756ad709d7038fc5f31f0

The reason it's doing this is because prevHash and merkleRoot are serialized in bitcoin's "hash byte order" which writes the bytes in reverse.

What Braydon and I are doing in ChainDB is instead of keeping prevHash and merkleRoot as Buffers, we convert them to strings in fromBuffer(). Then we convert them to buffers again in toBuffer().

maraoz:
ACK

PR #1202, opened 2015-04-28
fix Block#toObject API

maraoz:
This PR fixes https://github.com/bitpay/bitcore/issues/1180
and standardizes a little more the classes `Block` and `BlockHeader`

### DIFF
--- a/lib/block/block.js
+++ b/lib/block/block.js
@@ -43,18 +43,7 @@ Block._from = function _from(arg) {
   } else if (JSUtil.isValidJSON(arg)) {
     info = Block._fromJSON(arg);
   } else if (_.isObject(arg)) {
-    info = {
-      /**
-       * @name Block#header
-       * @type {BlockHeader}
-       */
-      header: arg.header,
-      /**
-       * @name Block#transactions
-       * @type {Transaction[]}
-       */
-      transactions: arg.transactions
-    };
+    info = Block._fromObject(arg);
   } else {
     throw new TypeError('Unrecognized argument for Block');
   }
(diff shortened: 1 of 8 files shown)


braydonf:
That should fix the issue, one thing that we may want to do, and may need to be a part of another PR, is to use strings for the hashes and merkleRoot everywhere except in `toBuffer` and `fromBuffer`. That way the reversing only needs to be dealt with during serialization.

pnagurny:
I agree with Braydon. This will fix the issue here and at some point you may want to consider just using strings everywhere except where serialized.

maraoz:
I agree with your idea, but let's do it in another PR :)

lib/transaction/transaction.js
```diff
@@ -313,6 +313,7 @@ Transaction.prototype.toObject = function toObject() {
     obj.changeIndex = this._changeIndex;
   }
   if (!_.isUndefined(this._fee)) {
+    console.log(this._fee);
```
braydonf: leftover `console.log`

braydonf:
Bump to 0.12 for this?

lib/transaction/transaction.js
maraoz: @braydonf the sniper. thanks

lib/block/block.js
```diff
@@ -62,27 +51,35 @@ Block._from = function _from(arg) {
 };
 
 /**
- * @param {String|Object} - A JSON string or object
+ * @param {String} - A JSON string
  * @returns {Object} - An object representing block data
  * @private
  */
 Block._fromJSON = function _fromJSON(data) {
```
braydonf: I think many fromJSON methods currently handle both an object and string, we should make sure these are all consistent.

maraoz: Reading the code I realized many classes don't have a `fromObject` method. I removed the Object option on those because I added a `fromObject` method. IMO we should have separate `fromJSON`/`fromObject` methods in each class. `fromJSON` most commonly will be equivalent to `Foo.fromObject(JSON.parse(input))`

maraoz:
Regarding bumping minor version: agreed. I'll release this Wednesday.

lib/block/block.js
braydonf: I'd say we should make this a part of another PR that we can go through add those methods where necessary, so that there there are round trips for `fromJSON/toJSON` and `fromObject/toObject`. However if either of those handle both a string and object, we only need one pair.

PR #1205, opened 2015-04-29
Fix block parsing 

maraoz:


martindale:
Reviewing – fix those tests, tho'.

test/script/script.js
```diff
@@ -755,4 +753,5 @@ describe('Script', function() {
     });
   });
 
+
```
martindale: nit: extra space

test/script/interpreter.js
```diff
@@ -36,6 +36,7 @@ Script.fromBitcoindString = function(str) {
       var tstr = token.slice(1, token.length - 1);
       var cbuf = new Buffer(tstr);
       tbuf = Script().add(cbuf).toBuffer();
+      //bw.writeUInt8(tstr.length);
```
martindale: nit: commented line

lib/transaction/input/input.js
```diff
@@ -105,14 +114,19 @@ Input.prototype.toBufferWriter = function(writer) {
 };
 
 Input.prototype.setScript = function(script) {
+  this._script = null;
   if (script instanceof Script) {
     this._script = script;
     this._scriptBuffer = script.toBuffer();
+  } else if (JSUtil.isHexa(script)) {
+    // hex string script
+    this._scriptBuffer = new Buffer(script, 'hex');
```
braydonf: Should `this._script` should be set here?

maraoz: No, it's only set where it's already there. In this case, it will be lazy loaded if requested

braydonf:
LGTM

PR #1208, opened 2015-04-30
Update bn.js to 2.0.4

braydonf:
Patches a bug when a number greater than or equal to `2 ^ 52` (and less than `2 ^ 53 - 1`) was used to initialize an instance of BN. The number would be represented incorrectly and lead to unexpected, and incorrect, output serialization. Note: `2 ^ 52` is greater than `21000000 * 1e8` (max number of satoshis).

See relevant commit: https://github.com/indutny/bn.js/commit/64d710d7babeb55f5e491976964653751282e23c
 ( https://github.com/indutny/bn.js/pull/46 )


maraoz:
LGTM. Could you add a test-case for 0 in the `JSUtil.isNaturalNumber` tests? Just to make clear what the expected behavior is.

PR #1213, opened 2015-05-09
Fixed script.isPublicKeyHashIn bug with variable length signatures.

braydonf:
- DER encoded signatures have a max length of 72 bytes, however a minimum size isn't defined.
- R and S values can have arbitrary lengths
- For more information: https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#der-encoding


maraoz:
ACK. We should maybe add something like `Signature.isValid` in the future

martindale:
We should add some fixtures for this particular case; #1221 for tracking.  Otherwise, nice work.

Issue #1174, opened 2015-04-05
Rounding error and incorrect fee


dskloet:
I tried to create a transaction with an output of 81.99 mBTC and a fee of 10 bits. But because 81.99 * 100000 becomes 8198999.999999999 I actually created a transaction with a non-integer amount of satoshis. This should simply have failed (or at most rounded to 81.99000 satoshis) but instead bitcore seems to have truncated it down to 8198999 satoshis because the resulting transaction was:

https://test-insight.bitpay.com/tx/8ed23b2513a8960e78602d2fb5b9e12eb7ba922f63ada6ddc4a75faf44f50e31

Since I didn't specify a change address, this also affected the fee, which became 1001 even though I explicitly specified it to be 1000.

So I see 2 bugs here:
1. A non-integer amount of satoshis was truncated without error.
2. A transaction was serialized with a fee different from the explicitly specified fee.

maraoz:
Can you show the part of the code where you created the transaction please? We'll investigate

dskloet:
I did something like
```javascript
var amount = 81.99 * 100000;
var fee = 1000;
var tx = new bitcore.Transaction().
    tx.from(cachedOutputs).
    to(toAddress, amount).
    fee(fee);
```
Do you need a full working example?

maraoz:
Please note that JavaScript has some problems when dealing with different bitcoin units. That's why we built the `Unit` class (http://bitcore.io/guide/unit.html)
Look:
```
$ node
> 81.99 * 100000
8198999.999999999
> var bitcore = require('bitcore');
undefined
> var Unit = bitcore.Unit;
undefined
> Unit.fromMilis(81.99).toSatoshis()
8199000
```

could this be the problem?

dskloet:
Thanks, good to know about `Unit`. But yes, I already knew the floating point imprecision was the problem. Did you see my points 1 and 2 at the bottom?

1. Do you agree that it should be an error to specify a non-integer number of satoshis and that it shouldn't just quietly truncate the amount?
2. And do you agree that if a fee is specified, it should never create a transaction with a different fee without error?

maraoz:
aaah, I missunderstood your two points, sorry. Yes, I agree with your two ideas. I'll move this issue to "ready". Feel free to implement this fixes in a PR, or we'll look at it in the future

dskloet:
I may try next weekend but it's a bit of a hassle since my employer needs to approve before I can contribute code to an open source project. So don't hold your breath.

[...]

PR #1214, opened 2015-05-09
Don't ignore the fee when it's explicitly specified.

dskloet:
Make sure a specified transaction fee and outputs add up to the sum of the inputs.

### DIFF
--- a/lib/errors/spec.js
+++ b/lib/errors/spec.js
@@ -85,8 +85,14 @@ module.exports = [{
     name: 'InvalidSatoshis',
     message: 'Output satoshis are invalid',
   }, {
-    name: 'FeeError',
-    message: 'Fees are not correctly set {0}',
+    name: 'SmallFeeError',
+    message: 'Fee is too small: {0}',
+  }, {
+    name: 'LargeFeeError',
+    message: 'Fee is too large: {0}',
+  }, {
+    name: 'DifferentFeeError',
+    message: 'Unspent value is different from specified fee: {0}',
   }, {
     name: 'ChangeAddressMissing',
     message: 'Change address is missing'
--- a/lib/errors/spec.js
+++ b/lib/errors/spec.js
@@ -85,16 +85,16 @@ module.exports = [{
     name: 'InvalidSatoshis',
     message: 'Output satoshis are invalid',
   }, {
-    name: 'Fee',
+    name: 'FeeError',
     message: 'Internal Error on Fee {0}',
     errors: [{
-      name: 'TooSmallError',
+      name: 'TooSmall',
       message: 'Fee is too small: {0}',
     }, {
-      name: 'TooLargeError',
+      name: 'TooLarge',
       message: 'Fee is too large: {0}',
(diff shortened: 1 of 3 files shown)


dskloet:
I'm looking at how the coverage could have decreased.

maraoz:
LGTM

dskloet:
I found a problem when investigating the coverage. I'll write a test and fix it. Is there a way to amend the pull request or do I have to make a new pull request?

martindale:
@dskloet just push new commits to the same branch and it'll automatically associate them with this pull request!  :)

dskloet:
Thanks!
The problem wasn't caught by the tests because a test that expected a FeeError passed because it got a FeeError for a different reason than intended. I'm fixing it by introducing some different kinds of fee errors to distinguish the different cases.

I've pushed the fix.

dskloet:
FYI: This pull requests addresses part of issue #1174.

lib/errors/spec.js
```diff
@@ -85,8 +85,14 @@ module.exports = [{
...
-    message: 'Fees are not correctly set {0}',
+    name: 'SmallFeeError',
+    message: 'Fee is too small: {0}',
+  }, {
+    name: 'LargeFeeError',
+    message: 'Fee is too large: {0}',
+  }, {
+    name: 'DifferentFeeError',
+    message: 'Unspent value is different from specified fee: {0}',
   }, {
```
braydonf: Perhaps we could make these additional errors part of a general `FeeError`:

```js
if (err instanceof errors.FeeError) {
  //...
}
```
And if interested in a more specific exception:
```js
if (err instanceof errors.FeeError.TooSmall) {
  //...
}
```

braydonf:
Is there a use case for disabling the check if fees are different? 

If so, the related [Transaction serialization docs](https://github.com/bitpay/bitcore/blob/master/docs/transaction.md#serialization-checks) need to be updated with the additional serialization option.

dskloet:
The check wasn't there before so I'm mostly just being defensive. Would you prefer I remove the ability to turn it off or add it to the docs?

dskloet:
I realized that if it can be disabled, I also need to add a test that it can be disabled.

lib/errors/spec.js
```diff
@@ -85,8 +85,18 @@ module.exports = [{
     name: 'InvalidSatoshis',
     message: 'Output satoshis are invalid',
   }, {
-    name: 'FeeError',
-    message: 'Fees are not correctly set {0}',
+    name: 'Fee',
```
braydonf: For backwards compatibility, we can keep this as `FeeError`

braydonf:
If the fee is specified and the actual fee would differ, I can't think of a use case to disable the check, as by not specifying an exact fee would be the means to do such. (I'll verify though).

dskloet:
Makes sense. Let me know when you've verified it and I'll remove it.

braydonf:
Yep, works as expected.

PR #1217, opened 2015-05-10
fix typo: defineProperty.writeable

fanatid:
writable instead writeable, see [defineProperty on developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty)

### DIFF
--- a/lib/block/block.js
+++ b/lib/block/block.js
@@ -277,7 +277,7 @@ Block.prototype._getHash = function() {
 
 var idProperty = {
   configurable: false,
-  writeable: false,
+  enumerable: true,
   /**
    * @returns {string} - The big endian hash buffer of the header
    */
--- a/lib/block/blockheader.js
+++ b/lib/block/blockheader.js
@@ -233,7 +233,6 @@ BlockHeader.prototype._getHash = function hash() {
 
 var idProperty = {
   configurable: false,
-  writeable: false,
   enumerable: true,
   /**
    * @returns {string} - The big endian hash buffer of the header
--- a/lib/transaction/input/input.js
+++ b/lib/transaction/input/input.js
@@ -27,7 +27,6 @@ Input.DEFAULT_SEQNUMBER = DEFAULT_SEQNUMBER;
 
 Object.defineProperty(Input.prototype, 'script', {
   configurable: false,
-  writeable: false,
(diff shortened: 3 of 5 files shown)


fanatid:
PR updated..
details about [data and accessor property](http://www.ecma-international.org/ecma-262/5.1/#sec-8.6.1)
tests passed before, because writable had a typo

maraoz:
Don't these changes make those properties writable? They should be immutable

fanatid:
Named accessor property doesn't have writable property, only: get, set, enumerable and confirgurable... and because this you can't use key writable or value...
for example:
```js
var x = {}
Object.defineProperty(x, 'x', {
  writable: true,
  get: function () { return this._x }
})
```
throws:
```js
TypeError: Invalid property.  A property cannot both have accessors and be writable or have a value, #<Object>
```
you don't have this error before only because instead writable you have writeable

braydonf:
ACK

Using `defineProperty` with `value`, `get`, etc. makes the property effectively immutable, for Block and BlockHeader the set is using  `noop`, however those may not be necessary.

lib/block/block.js
```diff
@@ -277,7 +277,7 @@ Block.prototype._getHash = function() {
 
 var idProperty = {
   configurable: false,
-  writeable: false,
+  enumerable: true,
```
braydonf: This is used for both `id`, and `hash` we may not want these both to be enumerable.

fanatid:
set is equal to _.noop has zero influence for property... should I delete this line?
@braydonf way you want make `id` and `hash` invisible?

fanatid:
btw, I also see in [unit.js](https://github.com/bitpay/bitcore/blob/698625cc47da94f977ad0346a3370d2d5411c58a/lib/unit.js#L57) doesn't use configurable in defineProperty, are you agree about adding this property?
also [Unit.key](https://github.com/bitpay/bitcore/blob/698625cc47da94f977ad0346a3370d2d5411c58a/lib/unit.js#L68) define as mutable object, defineImmutable will be better?

eordano:
`id` and `hash` take some time to be calculated, but it may not be bad to have them enumerated as long as the hash calculation doesn't get triggered...

PR #1218, opened 2015-05-11
improve PublicKey docs and fix #1215

### LINKED ISSUE #1215: PublicKey `toJSON()` returns invalid hex?
The `toJSON()` method from PublicKey seems to lose leading zeroes in the point coordinates, causing it to return invalid hexadecimal (odd length).

From the Bitcore Playground:

    >> publicKey.point
      <EC Point
        x: 0589ee559348bd6a7325994f9c8eff12bd5d73cc683142bd0dd1a17abc99b0dc
        y: 8b3f9a2d8b0d343aa021b9c4c57207acbfa9d95fcd7cacbc248f1a936c0422a5
      >

    >> publicKey.toJSON()
      {
        "x":"589ee559348bd6a7325994f9c8eff12bd5d73cc683142bd0dd1a17abc99b0dc",
        "y":"8b3f9a2d8b0d343aa021b9c4c57207acbfa9d95fcd7cacbc248f1a936c0422a5",
        "compressed":true
      }
  
Note that `toJSON().x` has one less character, the leading zero.

If this is actually a bug, the culprit may be `toObject()`, which uses the numeric coordinates without padding them. The output should still be valid hex, though, so I don't know.

    PublicKey.prototype.toObject = function toObject() {
      return {
        x: this.point.getX().toString('hex'),
        y: this.point.getY().toString('hex'),
        compressed: this.compressed
      };
    };

maraoz:
Fixes https://github.com/bitpay/bitcore/issues/1215

### DIFF
--- a/docs/publickey.md
+++ b/docs/publickey.md
@@ -9,7 +9,7 @@ description: A simple interface for handling private keys.
 Represents a bitcoin public key and is needed to be able to receive bitcoin, as is usually represented as a bitcoin [Address](address.md).
 See the official [Bitcoin Wiki](https://en.bitcoin.it/wiki/Technical_background_of_version_1_Bitcoin_addresses).
 
-A PublicKey in Bitcore is an immutable object and can be instantiated from a [Point](crypto.md), string, [PrivateKey](privatekey.md), Buffer and a [BN](crypto.md).
+A PublicKey in Bitcore is an immutable object and can be instantiated from a [Point](crypto.md), string, [PrivateKey](privatekey.md), Buffer or a [BN](crypto.md).
 
 ## Instantiate a Public Key
 
@@ -39,7 +39,7 @@ if (PublicKey.isValid('02a1633cafcc01ebfb6d78e39f687a1f0995c62fc95f51ead10a02ee0
 
 ## Handling compressed and uncompressed public keys
 
-It's important to note that there are two possible ways to represent public key.
+It's important to note that there are two possible ways to represent a public key.
 The standard is *compressed* and includes the X value and parity (as represented above in the documentation).
 There is also a longer version that is *uncompressed* which includes both X and Y values. Using this encoding will generate a different bitcoin address, so be careful when selecting the encoding.
 Uncompressed public keys start with 0x04; compressed public keys begin with 0x03 or 0x02 depending on whether they're greater or less than the midpoint of the curve. These prefix bytes are all used in official secp256k1 documentation.
--- a/docs/publickey.md
+++ b/docs/publickey.md
@@ -6,7 +6,10 @@ description: A simple interface for handling private keys.
 
 ## Description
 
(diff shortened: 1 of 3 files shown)


docs/publickey.md
```diff
@@ -6,7 +6,10 @@ description: A simple interface for handling private keys.
 
 ## Description
 
-Represents a bitcoin public key and is needed to be able to receive bitcoin, as is usually represented as a bitcoin [Address](address.md), see the official [Bitcoin Wiki](https://en.bitcoin.it/wiki/Technical_background_of_version_1_Bitcoin_addresses). A PublicKey in Bitcore is an immutable object and can be instantiated from a [Point](crypto.md), string, [PrivateKey](privatekey.md), Buffer and a [BN](crypto.md).
+Represents a bitcoin public key and is needed to be able to receive bitcoin, as is usually represented as a bitcoin [Address](address.md).
+See the official [Bitcoin Wiki](https://en.bitcoin.it/wiki/Technical_background_of_version_1_Bitcoin_addresses).
+
+A PublicKey in Bitcore is an immutable object and can be instantiated from a [Point](crypto.md), string, [PrivateKey](privatekey.md), Buffer and a [BN](crypto.md).
```
eordano: `...or a [BN]...` instead of `...and a [BN]...`

docs/publickey.md
```diff
@@ -34,4 +37,33 @@ if (PublicKey.isValid('02a1633cafcc01ebfb6d78e39f687a1f0995c62fc95f51ead10a02ee0
 }
 ```
 
-Note: It's important to note that there are two possible ways to represent public key, the standard is *compressed* and includes the x value and parity (as represented above in the documentation). There is also a longer version that is *uncompressed* which includes both x and y values, and using this can generate a different bitcoin address, so it's important to note this possibility, however it's discouraged to be used.
+## Handling compressed and uncompressed public keys
+
+It's important to note that there are two possible ways to represent public key.
```
eordano: `represent` (a) `public key`

docs/publickey.md
```diff
@@ -34,4 +37,33 @@ if (PublicKey.isValid('02a1633cafcc01ebfb6d78e39f687a1f0995c62fc95f51ead10a02ee0
 }
 ```
 
-Note: It's important to note that there are two possible ways to represent public key, the standard is *compressed* and includes the x value and parity (as represented above in the documentation). There is also a longer version that is *uncompressed* which includes both x and y values, and using this can generate a different bitcoin address, so it's important to note this possibility, however it's discouraged to be used.
+## Handling compressed and uncompressed public keys
```
braydonf: Uppercase for titles please :)

braydonf:
LGTM

Do we need to go into great detail about uncompressed public keys with full examples?

braydonf:
Also some of the examples line width is quite wide and do not format well in the docs rendering (requires side-scrolling).

docs/publickey.md
braydonf: Also, we could shorten to "Compressed vs Uncompressed"

braydonf:
ACK

braydonf:
Thanks!

PR #1226, opened 2015-05-12
Simplify transaction.getSerializationError()

dskloet:
This takes care of some lint errors about number of statements and cyclomatic complexity.

### DIFF
--- a/lib/transaction/transaction.js
+++ b/lib/transaction/transaction.js
@@ -184,7 +184,8 @@ Transaction.prototype.invalidSatoshis = function() {
 };
 
 /**
- * Retrieve a possible error that could appear when trying to serialize and broadcast this transaction
+ * Retrieve a possible error that could appear when trying to serialize and
+ * broadcast this transaction.
  *
  * @param {Object} opts allows to skip certain tests. {@see Transaction#serialize}
  * @return {bitcore.Error}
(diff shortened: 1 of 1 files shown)


dskloet:
I tried rebasing and pushing again to get the increased coverage from #1229 but I can't seem to push anymore after rebasing.

dskloet:
In my local branch the coverage is 97.79%.

dskloet:
I noticed that my newer pull request is already merged and no comments here yet. Is this change controversial?

dskloet:
@braydonf what does the "in progress" label mean? Do you expect more before this is ready?

lib/transaction/transaction.js
```diff
@@ -231,24 +217,39 @@ Transaction.prototype._isFeeDifferent = function() {
...
 
-Transaction.prototype._isFeeTooLarge = function() {
+Transaction.prototype._isFeeTooLarge = function(opts) {
+  if (opts.disableLargeFees) {
+    return;
+  }
   var fee = this._getUnspentValue();
-  var maximumFee = Math.floor(Transaction.FEE_SECURITY_MARGIN * this._estimateFee());
+  var maximumFee = Math.floor(
+      Transaction.FEE_SECURITY_MARGIN * this._estimateFee());
```
braydonf: This line can fit on one line, and be under 120.

lib/transaction/transaction.js
```diff
@@ -231,24 +217,39 @@ Transaction.prototype._isFeeDifferent = function() {
...
 };
 
-Transaction.prototype._isFeeTooSmall = function() {
+Transaction.prototype._isFeeTooSmall = function(opts) {
+  if (opts.disableSmallFees) {
+    return;
+  }
   var fee = this._getUnspentValue();
-  var minimumFee = Math.ceil(this._estimateFee() / Transaction.FEE_SECURITY_MARGIN);
+  var minimumFee = Math.ceil(
```
braydonf: Similar here

lib/transaction/transaction.js
```diff
@@ -192,38 +193,23 @@ Transaction.prototype.invalidSatoshis = function() {
...
-    return new errors.Transaction.DustOutputs();
-  }
-  if (!opts.disableIsFullySigned && !isFullySigned) {
-    return new errors.Transaction.MissingSignatures();
-  }
-  if (!opts.disableMoreOutputThanInput && this._getUnspentValue() < 0) {
-    return new errors.Transaction.InvalidOutputAmountSum();
-  }
+Transaction.prototype._hasFeeError = function(opts) {
+  return this._isFeeDifferent() ||
```
braydonf: Can also go on one line.

lib/transaction/transaction.js
```diff
@@ -256,15 +257,36 @@ Transaction.prototype._missingChange = function() {
...
+Transaction.prototype._hasDustOutputs = function(opts) {
+  if (opts.disableDustOutputs) {
+    return;
+  }
   var index, output;
   for (index in this.outputs) {
     output = this.outputs[index];
-    if (output.satoshis < Transaction.DUST_AMOUNT && !output.script.isDataOut()) {
-      return true;
+    if (output.satoshis < Transaction.DUST_AMOUNT &&
```
braydonf: Similar, one line

dskloet:
You use 120 character lines?
The contribution document mentions not adding gulp lint warnings and it warns about lines over 80.

braydonf:
As per the jshint, yes: https://github.com/bitpay/bitcore/blob/master/.jshintrc

braydonf:
Also the order of these checks may need to be changed, see: https://github.com/bitpay/bitcore/pull/1226

dskloet:
This is purely a refactoring. If the logic needs to change it should be a different change with new tests.

lib/transaction/transaction.js
dskloet: I think it's more readable like this, don't you?

dskloet:
I've merge 3 of the 4 lines back to be long.
But in one case I broke it up because I think it's more readable, not because of the length.
Do you want me to merge that one into one line as well?

lib/transaction/transaction.js
```diff
@@ -231,24 +217,37 @@ Transaction.prototype._isFeeDifferent = function() {
     var fee = this._fee;
     var unspent = this._getUnspentValue();
     if (fee !== unspent) {
-      return 'Unspent value is ' + unspent + ' but specified fee is ' + fee;
+      return new errors.Transaction.FeeError.Different(
+          'Unspent value is ' + unspent + ' but specified fee is ' + fee);
```
eordano: nit: This would be more consistent if the end parenthesis is on the next line and the indentation is only two spaces (like `{` and `}` on functions):
```javascript
return new errors.Transaction.FeeError.Different(
  'Unspent value is ' + unspent + ' but specified fee is ' + fee
);
```

See for example https://github.com/bitpay/bitcore/blob/master/lib/transaction/transaction.js#L363-L365

dskloet: Wow, that looks super weird to me :). But this is not my project so I will adapt. But do you think you can make the linter say these things? That would make it easier to adopt the correct coding style.

Actually, this fits in 120 characters so I just put it all on one line.

lib/transaction/transaction.js
maraoz: +1 to multiple lines

braydonf:
LGTM

maraoz:
LGTM

maraoz:
Great work @dskloet

dskloet:
Thanks, it's a pity it was reverted in #1248.

PR #1227, opened 2015-05-13
remove inBrowser

fanatid:
In browserify `global` referrers to `window` -- [see docs](https://github.com/substack/browserify-handbook#global)
and check on browser detection was wrong, module `process` exists in browserify bundle

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,24 +3,19 @@ var bitcore = module.exports;
 // module information
 bitcore.version = 'v' + require('./package.json').version;
 
-var inBrowser = typeof process === 'undefined' || typeof process.versions === 'undefined';
-if ((inBrowser && window._bitcore) || (!inBrowser && global._bitcore)) {
-  var versions = bitcore.version + ' and ' + (inBrowser ? window._bitcore : global._bitcore);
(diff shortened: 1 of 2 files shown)


braydonf:
We should also add unit tests around the check (unfortunately missing currently), and the message can be inaccurate, e.g. the versions can be the same.

fanatid:
understand, I will try to do it in nearest days

fanatid:
PR updated

maraoz:
LGTM

fanatid:
but.. may be better will be use `process.browser` instead `typeof window`, thoughts?

fanatid:
... and I'll happy move guard to package, because I want use something like this in [coloredcoinjs-lib](http://github.com/chromaway/coloredcoinjs-lib)

braydonf:
@fanatid: Submitted a PR, should be able to run a unit test and check that global._bitcore is defined in Node.js and browsers, and that the guard will throw an error: https://github.com/fanatid/bitcore/pull/1

fanatid:
@braydonf thanks!

pnagurny:
ACK

PR #1231, opened 2015-05-13
Handling Invalid Output Scripts. Closes #1230

### LINKED ISSUE #1230: Invalid Script Handling for Output.toObject()
output.toObject() with an invalid script case isn't currently handled, as script would be `null`. We may need to do something similar to:

```js
Output.prototype.toObject = function toObject() {
  var obj = {
    satoshis: this.satoshis
  };
  if (this.script) {
    obj.script = this.script.toString()
  } else {
    obj.script = this._scriptBuffer.toString('hex')
  }
};
```

braydonf:
- Changed `toObject` serialization to always use a hexa string for a script
- Updated inspect method to handle a null script
- Roundtrip `toObject` with an invalid script
- Additional test coverage for Output

### DIFF
--- a/test/transaction/output.js
+++ b/test/transaction/output.js
@@ -22,6 +22,12 @@ describe('Output', function() {
     script: Script.empty()
   });
 
+  it('throws error with unrecognized argument', function() {
+    (function() {
+      var out = new Output(12345);
+    }).should.throw(TypeError);
+  });
+
   it('can be assigned a satoshi amount in big number', function() {
     var newOutput = new Output({
       satoshis: new BN(100),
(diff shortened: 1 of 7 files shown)


test/transaction/output.js
```diff
@@ -134,22 +162,18 @@ describe('Output', function() {
...
-    });
-    output._scriptBuffer = 'bad';
-
     (function() {
-      var result = output.script;
+      var output = new Output({
+        satoshis: 1000,
+        script: 'bad'
+      });
     }).should.throw('Invalid hex string');
```
maraoz: Would it be better to throw a more specific error? (i.e: `errors.Transaction.Output.InvalidScript`)?

braydonf: This is an error from Buffer that will take the else path, so we can handle an `InvalidScript` but not other types of errors.

pnagurny:
Looks good to me

PR #1253, opened 2015-05-27
add base to BN.fromString

maraoz:
I encountered a problem with `BN.fromString`. 

A quick workaround is to use BN.fromBuffer instead, but this should be fixed

### DIFF
--- a/lib/crypto/bn.js
+++ b/lib/crypto/bn.js
@@ -21,9 +21,9 @@ BN.fromNumber = function(n) {
   return new BN(n);
 };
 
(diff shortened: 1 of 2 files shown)


maraoz:
if someone wants to tackle this, they can create PRs to my branch. I can't fix this right now

braydonf:
May be an issue with not passing a base option to BN.

braydonf:
Still WIP?

maraoz:
updated title

erc155:
e2542866d1134c11920190b686be69434c092dbf

PR #1258, opened 2015-06-01
Removed double errors in publickey. Fixes #1256

### LINKED ISSUE #1256: Remove Double Error Creation
There are a few places where an error is being passed as the message string for preconditions, such as here: https://github.com/bitpay/bitcore/blob/master/lib/publickey.js#L157

We can replace:
```js
$.checkArgument(PublicKey._isBuffer(buf), new TypeError('Must be a hex buffer of DER encoded public key'));
```
with:
```js
$.checkArgument(PublicKey._isBuffer(buf), 'Must be a hex buffer of DER encoded public key');
```

braydonf:


maraoz:
ACK

matiu:
Code ACK. Tests pass.

PR #1262, opened 2015-06-03
fixed PrivateKey initialization from JSON

mpolci:
I discovered this error after calling toWIF() method on privateKey initialized from a json text like this:
{"bn":"ce6...fec","compressed":true,"network":"testnet"}
The network was not initialized with a Network object but toWIF() needs to read network.privatekey.


braydonf:
I think we should add a test that checks instantiation from JSON if it doesn't already exist.

mpolci:
I wrote the test

pnagurny:
LGTM

PR #1264, opened 2015-06-05
fix Transaction.sortOutputs

### LINKED ISSUE #1261: InvalidSorting raised if there is no change
If there is no change then calling `.shuffleOutputs` or `.sortOutputs` on a transaction will raise `errors.Transaction.InvalidSorting()` - "The sorting function provided did not return the change output as one of the array elements".

This happens even if `.change` is set appropriately in the case when no change is required. It might be strange to sort the outputs when no change address is required, but i believe the code should handle it more gracefully.

The issue appears to lie in `_newOutputOrder` checking `this.outputs[this._changeIndex]` when `_changeIndex` is undefined.

fanatid:
Fix #1261

### DIFF
--- a/lib/transaction/transaction.js
+++ b/lib/transaction/transaction.js
@@ -899,16 +899,18 @@ Transaction.prototype.sortOutputs = function(sortingFunction) {
 };
 
 Transaction.prototype._newOutputOrder = function(newOutputs) {
-  var changeIndex = 0;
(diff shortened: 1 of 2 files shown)


braydonf:
LGTM

matiu:
ACK.

braydonf:
Included in 0.12.10. Thanks!

PR #1267, opened 2015-06-09
Modularize privkey network version check

brandonrobertz:
This is a fix for the Bitcoin (mainnet & testnet) specific hardcode that only allows privkeys to be imported for those two networks. This change allows custom networks to import private keys, as well. See https://github.com/bitpay/bitcore/issues/1265 for discussion.

Issue #1265, opened 2015-06-08
Bitcoin-Specific Hardcode in Privkey.js


brandonrobertz:
I'm writing a namecoin module for bitcore, and mostly everything is working well. Networks are modular and it's easy to add new ones. I've added a `namecoin` network, which basically allows bitcore to generate namecoin private keys and addresses out-of-the-box.

The problem comes when I generate a private key using something like vanitygen and then import it with bitcore.PrivateKey. I'm doing this:

    var bitcore = require('./'); // I'm in bitcore root
    // I've added namecoin to networks.js as default
    var nmc = bitcore.Networks.get('namecoin');

    var wif = '74pxNKNpByQ2kMow4d9kF6Z77BYeKztQNLq3dSyU4ES1K5KLNiz';
    var privKey = bitcore.PrivateKey.fromWIF( wif);

I get a thrown error of `Invalid Network` which comes from the following code in `privatekey.js` from the `PrivateKey._transformBuffer` function:

    info.network = Networks.get(buf[0], 'privatekey');
    if (buf[0] === Networks.livenet.privatekey) {
      info.network = Networks.livenet;
    } else if (buf[0] === Networks.testnet.privatekey) {
      info.network = Networks.testnet;
    } else {
      throw new Error('Invalid network');
    }

Is there a reason why we're hardcoding `Networks.testnet` and `Networks.livenet` for private key import? When I patch that line with another else if, looking for `Networks.namecoin` everything appears to work, but is there some gotcha I'm not aware of?

maraoz:
You're right. That could be done in a more generic way by iterating through all the available networks and checking if `buf[0]` is equal to a network's `privatekey` value.

brandonrobertz:
I can do a PR for this, we will need to expose `Networks.networks` or a new `Networks.all()` method to iterate through the networks. Is there any interest?

brandonrobertz:
I created a PR for this issue with tests. https://github.com/bitpay/bitcore/pull/1267

Any comments?

brandonrobertz:
New PR is here: https://github.com/bitpay/bitcore/pull/1268

braydonf:
Yep, we've previously updated other areas of the code to use any network defined in networks, instead of `livenet` and `testnet` specifically.

braydonf:
Included in https://github.com/bitpay/bitcore/commit/8e13b68405b591f90597fec34b9f4c26594247b4

PR #1268, opened 2015-06-09
modularize network version check issues/1265 (2nd)

brandonrobertz:
This is a fix for the Bitcoin (mainnet & testnet) specific hardcode that only allows privkeys to be imported for those two networks. This change allows custom networks to import private keys, as well. See #1265 for discussion.

This is mostly the same as the old, closed, PR (https://github.com/bitpay/bitcore/pull/1267) but with slightly different test structure.

### DIFF
--- a/lib/privatekey.js
+++ b/lib/privatekey.js
@@ -158,16 +158,9 @@ PrivateKey._transformBuffer = function(buf, network) {
 
   info.network = Networks.get(buf[0], 'privatekey');
 
-  var allNetworks = Networks.all();
-  var matches = _.filter( allNetworks, function( network) {
-    return buf[0] === network.privatekey;
-  });
-
-  if (matches.length !== 1) {
+  if (!info.network) {
     throw new Error('Invalid network');
   }
-
-  info.network = matches[0];
 
   if (network && info.network !== Networks.get(network)) {
     throw new TypeError('Private key network mismatch');
--- a/lib/privatekey.js
+++ b/lib/privatekey.js
@@ -157,13 +157,17 @@ PrivateKey._transformBuffer = function(buf, network) {
   }
 
   info.network = Networks.get(buf[0], 'privatekey');
-  if (buf[0] === Networks.livenet.privatekey) {
-    info.network = Networks.livenet;
-  } else if (buf[0] === Networks.testnet.privatekey) {
-    info.network = Networks.testnet;
-  } else {
+
+  var allNetworks = Networks.all();
+  var matches = _.filter( allNetworks, function( network) {
+    return buf[0] === network.privatekey;
+  });
+
+  if (matches.length !== 1) {
     throw new Error('Invalid network');
   }
+
+  info.network = matches[0];
 
   if (network && info.network !== Networks.get(network)) {
     throw new TypeError('Private key network mismatch');
--- a/test/privatekey.js
+++ b/test/privatekey.js
@@ -44,6 +44,35 @@ describe('PrivateKey', function() {
     should.exist(a.bn);
   });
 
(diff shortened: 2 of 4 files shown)


lib/privatekey.js
```diff
@@ -157,14 +157,18 @@ PrivateKey._transformBuffer = function(buf, network) {
   }
 
   info.network = Networks.get(buf[0], 'privatekey');
```
fanatid: This does not make a sense, `info.network` redefined later. Probably someone wanted make fast check based on `Networks.get`, but it wasn't done in old code.

brandonrobertz:
@fanatid The code already did the redundant define when I touched it, so I left it. Removing the line doesn't impact the tests.

lib/privatekey.js
brandonrobertz: @fanatid The code already did the redundant define when I touched it, so I left it. Removing the line doesn't impact the tests, so it should be able to be removed safely.

matiu:
LGTM, it would be nice to test:
https://github.com/bitpay/bitcore/pull/1268/files#diff-18bd3d7f5d6d6ffb54b82f6f432f140aR166
ie: if network is not recognised.

brandonrobertz:
I added a test for explicitly passing in a bad network. That will test the filter length being equal to zero and throwing error. This is in addition to an existing test in `test.privatekey.js` that checks a bad WIF where no existing network matches.

matiu:
thanks. 

ACK.

braydonf:
Using `.all()` may not be necessary since .`get(buf[0], 'privatekey')` will use all networks already.

lib/privatekey.js
braydonf: Yep, line 160 to 164 wasn't necessary from before since `info.network = Networks.get(buf[0], 'privatekey');` is all that is needed.

brandonrobertz:
@braydonf You're correct. I removed unnecessary code. Tests still passing. Good catch.

braydonf:
Is there a need for `.all()` in networks still?

brandonrobertz:
I cleaned out the tests and the functions I'd previously added. I went through my bitcore-namecoin tests to make sure that `Networks.get` is all we need here. All my tests are passing, so I believe we've got all bases covered here.

test/privatekey.js
```diff
@@ -114,6 +140,22 @@ describe('PrivateKey', function() {
...
+        alias: 'invalidnet',
+        pubkeyhash: 0xFF,
+        privatekey: 0xFF,
+        scripthash: 0xFF,
+        xpubkey: 0xFFFFFFFF,
+        xprivkey: 0xFFFFFFFF,
+        networkMagic: 0xFFFFFFFF
+      };
+      expect(function(){
+        var a = new PrivateKey( wifNamecoin, invalidNet );
```
braydonf: `var a = new PrivateKey(wifNamecoin, 'testnet');` will also work here, since invalidNet isn't added to the networks.

test/privatekey.js
```diff
@@ -44,6 +45,31 @@ describe('PrivateKey', function() {
...
+      networkMagic: 0xf9beb4fe,
+      port: 20001,
+      dnsSeeds: [
+        'localhost',
+        'mynet.localhost'
+      ]
+    };
+    Networks.add(nmc);
+    var nmcNet = Networks.get('namecoin');
+    var a = new PrivateKey( wifNamecoin, nmcNet);
```
braydonf: nit: spacing: `var a = new PrivateKey(wifNamecoin, nmcNet);`

braydonf:
Landed in 0.12.10 Thanks, and good catch.

PR #1271, opened 2015-06-11
Fix/fee estimation

isocolsky:
This PR does two things:
- Fix fee calculation used to determine if a change address is needed (when estimating tx size)
- Allow configuration of fee per KB for each transaction.


matiu:
LGTM. having the possibility to set fee per KB will be useful in many use cases.

pnagurny:
ACK

PR #1272, opened 2015-06-12
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "bitcore",
   "main": "./bitcore.min.js",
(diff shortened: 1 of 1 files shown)


maraoz:
Agreed but needs update to bitcore-build

braydonf:
Yep this needs to be updated it bitcore-build, spawned an issue there: https://github.com/bitpay/bitcore-build/issues/28

PR #1276, opened 2015-06-24
get-signature-count needed to measure txs with many inputs

greggzigler:


maraoz:
LGTM

braydonf:
LGTM.

We can add `getP2SHSignatureOperationsCount()` in another PR to handle, the other argument type: https://github.com/bitcoin/bitcoin/blob/master/src/script/script.cpp#L190

PR #1279, opened 2015-06-26
Improved performance of block parsing

braydonf:
- Increases performance of block parsing by ~9x.
- Removes expensive precondition from `readReverse`
- Optimizes `readUInt64LEBN`
- Adds benchmarks for both of the above methods
- Adds additional tests around `readUInt64LEBN`

### DIFF
--- a/lib/encoding/bufferreader.js
+++ b/lib/encoding/bufferreader.js
@@ -91,8 +91,16 @@ BufferReader.prototype.readUInt64BEBN = function() {
 };
 
 BufferReader.prototype.readUInt64LEBN = function() {
-  var data = Array.prototype.slice.call(this.buf, this.pos, this.pos + 8);
-  var bn = new BN(data, 10, 'le');
+  var second = this.buf.readUInt32LE(this.pos);
+  var first = this.buf.readUInt32LE(this.pos + 4);
+  var combined = (first * 0x100000000) + second;
+  var bn;
+  if (combined <= 0x1fffffffffffff) {
+    bn = new BN(combined);
+  } else {
+    var data = Array.prototype.slice.call(this.buf, this.pos, this.pos + 8);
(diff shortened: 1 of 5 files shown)


pnagurny:
LGTM

lib/encoding/bufferreader.js
```diff
@@ -91,11 +91,16 @@ BufferReader.prototype.readUInt64BEBN = function() {
...
 
 BufferReader.prototype.readUInt64LEBN = function() {
-  var buf = this.buf.slice(this.pos, this.pos + 8);
-  var reversebuf = BufferReader({
-    buf: buf
-  }).readReverse();
-  var bn = BN.fromBuffer(reversebuf);
+  var second = this.buf.readUInt32LE(this.pos);
+  var first = this.buf.readUInt32LE(this.pos + 4);
+  var combined = (first * 0x100000000) + second;
```
maraoz: Can't this lose precision? See:
```
> 0x100000000 * 0x10000000 
1152921504606847000
> 0x100000000 * 0x10000000 + 1
1152921504606847000
```
Both answers give the same value. Does this mean that if `second` is `0` or `1` and `first` is `0x10000000`, this would return the same value?

braydonf: It can lose precision with numbers equal or above `2 ^ 53`.

```js
> 0x1fffffffffffff === Math.pow(2, 53) - 1
// true
```
```js
> 1152921504606847000 < 0x1fffffffffffff
// false
```

braydonf:
Rebased on master and added comment to code with info about number/array BN instantiation.

lib/encoding/bufferreader.js
maraoz: Ahhh, gotcha. I didn't understand the `< 0x1fffffffffffff` part. LGTM

PR #1281, opened 2015-06-29
WIP: Added RawBlock for faster block parsing use cases.

braydonf:


braydonf:
Error in benchmark lead to misrepresented timings, little performance gain from including all parsing in one constructor, bufferreader is likely where optimizations can be made instead.

Issue #1266, opened 2015-06-08
SIGHASH_ANYONECANPAY maybe should be separated from sigtype


snakecharmer1024:
Transaction.prototype.applySignature will break if the sigtype does not include the correct anyonecanpay flag. This is probably the correct behavior.

Signature.prototype.hasDefinedHashtype will break if the signature's hashtype has the anyonecanpay flag set. I think this is a bug. 

I propose the following change:

Signature.prototype.hasDefinedHashtype = function() {
  var hashtype = this.nhashtype & 0xf;
//  if (this.nhashtype < Signature.SIGHASH_ALL || this.nhashtype > Signature.SIGHASH_SINGLE) {
  if (hashtype < Signature.SIGHASH_ALL || hashtype > Signature.SIGHASH_SINGLE) {
    return false;
  }
  return true;
};

This issue came up in working thru payment channels

braydonf:
This should be revolved with #1284 from @mruddy and landed in v0.12.11

PR #1284, opened 2015-06-30
fix signature hasDefinedHashtype

mruddy:
lib/crypto/signature.hasDefinedHashtype was not very closely matching Bitcoin Core's IsDefinedHashtypeSignature: https://github.com/bitcoin/bitcoin/blob/bc60b2b4b401f0adff5b8b9678903ff8feb5867b/src/script/interpreter.cpp#L180-189

This PR narrows that gap.

The purpose of the isNumber check is to protect against instances that simply have not had an nhashtype set (yet).

The previous version of this method was not considering Signature.SIGHASH_ANYONECANPAY at all. That was the obvious thing to fix.

### DIFF
--- a/lib/crypto/signature.js
+++ b/lib/crypto/signature.js
@@ -4,6 +4,7 @@ var BN = require('./bn');
 var _ = require('lodash');
 var $ = require('../util/preconditions');
 var BufferUtil = require('../util/buffer');
+var JSUtil = require('../util/js');
 
 var Signature = function Signature(r, s) {
   if (!(this instanceof Signature)) {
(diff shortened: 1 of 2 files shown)


braydonf:
LGTM

mruddy:
I just updated to also check with JSUtil.isNaturalNumber.

The reason is that I added another test case, nhashtype being set to the floating point 1.1, and that was passing. I figured it'd be better to have that case fail/return false because that might help library clients fail faster and narrow down a bug to an incorrectly set nhashtype.

I also saw that #1266 mentions the main hasDefinedHashtype bug fixed by this PR.

PR #1287, opened 2015-07-02
Add CLTV (BIP65) support

eordano:
Mostly to be compatible with https://github.com/bitcoin/bitcoin/pull/6124


lib/script/interpreter.js
```diff
@@ -184,6 +184,9 @@ Interpreter.false = new Buffer([]);
 
 Interpreter.MAX_SCRIPT_ELEMENT_SIZE = 520;
 
+Interpreter.LOCKTIME_THRESHOLD = 500000000;
+Interpreter.LOCKTIME_THRESHOLD_BN = new BN(500000000);
```
maraoz: I would do `new BN(Interpreter.LOCKTIME_THRESHOLD)`

lib/script/interpreter.js
```diff
@@ -311,6 +317,52 @@ Interpreter.prototype.evaluate = function() {
...
 };
 
+/**
+ * Checks a locktime parameter with the transaction's locktime
+ *
+ * @param {BN} nLockTime the locktime read from the script
+ * @return {boolean} true if the transaction's locktime is less than or equal to
+ *                   the transaction's locktime
+ */
+Interpreter.prototype.checkLockTime = function(nLockTime) {
```
braydonf: We should likely reference `TransactionSignatureChecker::CheckLockTime` from Bitcoin Core in the comment here, and further information (such as comments?) can be found there.

braydonf:
I think it would be good to add unit tests around `Interpreter.prototype.checkLockTime`

eordano:
All right! Updated comments + test coverage.

lib/script/interpreter.js
eordano: Updated (although for some reason this diff is not shown as outdated)

braydonf:
LGTM

braydonf:
The data for tx_invalid.json differs from the data in https://github.com/bitcoin/bitcoin/pull/6124/files Shouldn't these be identical? e.g. We should be able to copy tx_invalid.json directly.

eordano:
Updated to match. Thanks for reviewing! I overlooked those tests by mistake

braydonf:
Great! 

May be a good idea to submit to Bitcoin Core the test data for the last test "Failure due to empty stack before CHECKLOCKTIMEVERIFY in redeemScript".

PR #1294, opened 2015-07-06
Objects passed as arguments are unmodified.

braydonf:
Ran into an issue where arguments passing into a Transaction constructor were mysteriously being modified in the transaction `toObject/fromObject` round trip tests, this will fix the issue so that arguments are not modified.


pnagurny:
LGTM

PR #1296, opened 2015-07-08
Optimizations for `script.toAddress()`

braydonf:
Several optimizations for `script.toAddress()` for around a 4x speed improvement.

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -7,7 +7,6 @@ var Base58Check = require('./encoding/base58check');
 var Networks = require('./networks');
 var Hash = require('./crypto/hash');
 var JSUtil = require('./util/js');
-var Script = require('./script');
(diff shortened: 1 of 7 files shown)


lib/address.js
```diff
@@ -505,3 +504,5 @@ Address.prototype.inspect = function() {
 };
 
 module.exports = Address;
+
+var Script = require('./script');
```
maraoz: why this change?

maraoz:
LGTM

lib/address.js
braydonf: Resolves the circular dependency between `address.js` and `script.js` without needing to have several `address.js` require statements in functions in `script.js` that slowed each function down each time it was invoked. Using a late definition here is faster.

pnagurny:
LGTM

PR #1298, opened 2015-07-13
Speed up isPublicKeyOut

braydonf:


maraoz:
LGTM, what's the performance increase?

braydonf:
For the block in the benchmarks it's about 120 times faster, since public key outputs are infrequent.

maraoz:
nice :shipit:

eordano:
LGTM

PR #1303, opened 2015-07-30
Remove versionGuard error if versions are same.

olalonde:
I'm not sure why there is this version guard in the first place. It's extremely annoying and non idiomatic. What's the big deal if there is more than one bitcore in memory?

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,7 @@ var bitcore = module.exports;
 // module information
 bitcore.version = 'v' + require('./package.json').version;
 bitcore.versionGuard = function(version) {
-  if (version !== undefined) {
+  if (version !== undefined && version !== bitcore.version) {
     var message = 'More than one instance of bitcore found with versions: ' + bitcore.version +
       ' and ' + version + '. Please make sure to require bitcore and check that submodules do' +
       ' not also include their own bitcore dependency.';


braydonf:
There is still an issue with running multiple instances of Bitcore with the same version. `instanceof` handling will fail, and the networks object is a singleton and thus adding and getting networks will behave unexpectedly.

If Bitcore plugins are used, including bitcore in the main `package.json` dependencies will elevate issues as would be expected when using `peerDependencies` and npm@3.

olalonde:
Ok guess we can close this though I'm not really sure that's how `peerDependencies` work. The most annoying thing is that it break `npm link`.

kaibakker:
I got the same issue and I am not that experienced with npm.
I added 
```
"peerDependencies": {
    "bitcore": "^0.12.15"
  },
```
but when running ```npm update``` it gives ```The package bitcore does not satisfy its siblings' peerDependencies requirements!```. ```npm outdated``` says 0.12.15 is the only version of bitcore I use.

A little more explanation would be great for those who are less experienced with npm.

braydonf:
Please see: https://github.com/bitpay/bitcore/issues/1328 there are more details there. It's likely that bitcore needs to be required at the top-level.

PR #1305, opened 2015-08-06
Memoize transaction.hash

olalonde:
What this basically does is avoid computing the hash if we already did so previously. It just memorizes the last computed hash so doesn't leak memory.


braydonf:
Looks like there is an issue here with Node.js v0.10 with the tests.

How is computing the hash again leaking memory? And wouldn't storing the buffer increase the memory footprint for each transaction?

There may be performance benefits to something like this, especially if hash is called multiple times, however `toBuffer()` may actually be the bottleneck and not the hashing. Though benchmarks would help prove such. Any idea of possible performance improvements?

PR #1310, opened 2015-08-11
Also check for a publickeyout in toAddress

### LINKED ISSUE #1306: toAddress() publickey out addresses
So that adding this additional case isn't needed:
```js
 var address = script.toAddress(this.db.network);
 if (!address && script.isPublicKeyOut()) {
    var pubkey = script.chunks[0].buf;
    address = Address.fromPublicKey(new PublicKey(pubkey), this.db.network);
 }
```

And can be:

```js
 var address = script.toAddress(this.db.network);
```

braydonf:
Fixes #1306

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -221,9 +221,10 @@ Address._transformBuffer = function(buffer, network, type) {
  */
 Address._transformPublicKey = function(pubkey) {
   var info = {};
-  if (!(pubkey instanceof PublicKey)) {
-    throw new TypeError('Address must be an instance of PublicKey.');
-  }
+  $.checkArgument(
+    pubkey instanceof PublicKey,
+    'Address argument is expected to be an instance of PublicKey.'
+  );
   info.hashBuffer = Hash.sha256ripemd160(pubkey.toBuffer());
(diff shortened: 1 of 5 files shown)


maraoz:
~NACK. There's no actual address in a pay to public key output. It could be useful to have this tough... @jgarzik thoughts?

braydonf:
Currently insight.bitpay.com displays this output with an address: https://insight.bitpay.com/block/00000000839a8e6886ab5951d76f411475428afc90947ee320161bbf18eb6048

braydonf:
Just took a look at other block explorers, and the early publickey-outs are being indexed with an address (even though there technically isn't one because it's just the public key): 

https://blockchain.info/block/00000000839a8e6886ab5951d76f411475428afc90947ee320161bbf18eb6048

https://www.blocktrail.com/BTC/block/00000000839a8e6886ab5951d76f411475428afc90947ee320161bbf18eb6048

https://btc.blockr.io/block/info/1

braydonf:
We can handle this with the block handler, closing until there is a better solution.

PR #1311, opened 2015-08-11
Include hash in toObject/toJSON results.

### LINKED ISSUE #1308: `toObject()` and `toJSON()` missing hashes
While the hash of a transaction or block are not necessary for serialization, it may make sense to include the hash as a part of the `toJSON()` and `toObject()` results for convienience, and so that additional computation isn't needed.

braydonf:
Fixes #1308 and https://github.com/bitpay/bitcore-node/issues/74

### DIFF
--- a/lib/block/blockheader.js
+++ b/lib/block/blockheader.js
@@ -163,6 +163,7 @@ BlockHeader.fromBufferReader = function fromBufferReader(br) {
  */
 BlockHeader.prototype.toObject = function toObject() {
   return {
+    hash: this.hash,
     version: this.version,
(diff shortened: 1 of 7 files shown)


maraoz:
I would add a check in `fromObject` to see if the calculated hash equals the received hash

braydonf:
Currently the hash property in `fromObject` is ignored, and the hash would then always be the computed value...

braydonf:
Updated! The argument hash properties are checked for consistency with the calculated hash for both Transaction and BlockHeader.

maraoz:
Nice. LGTM

Issue #1307, opened 2015-08-07
`JSON.stringify()` standardize `toJSON()` and `toObject()`


braydonf:
Calling `JSON.stringify()` will call the `toJSON()` method of an object, which currently returns an already stringified string in Bitcore, however the standard (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON%28%29_behavior) is for it to return an object that can be stringified. Where Bitcore objects have a `toObject()` method that returns this as a result, it's non-standard and doesn't work with `JSON.stringify()`, and an additional wrapper is needed to call `toObject()` instead. The expected behavior, with Buffer as an example:
```js
> var a = new Buffer('abcdef', 'hex');
> a.toJSON()
{ type: 'Buffer', data: [ 171, 205, 239 ] }
> JSON.stringify(a);
'{"type":"Buffer","data":[171,205,239]}'
```

PR #1312, opened 2015-08-12
Changes behavior of `toJSON` to work as expected with `JSON.stringify`

braydonf:
- see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON%28%29_behavior
- Updates CONTRIBUTING with changes to Stardard Methods
- Aliases toJSON for toObject
- Removes all `fromJSON` methods, and many cases replaces with `fromObject`
- Constructors expect an object parsed via `JSON.parse` for JSON input
- Closes https://github.com/bitpay/bitcore/issues/1307

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,15 +85,27 @@ var bufferUtil = require('./util/buffer');
 #### G7 - Standard Methods
 
 When possible, bitcore objects should have standard methods on an instance prototype:
-* `toObject` - A plain JavaScript object that can be JSON stringified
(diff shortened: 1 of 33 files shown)


pnagurny:
ACK

braydonf:
Rebased on master

dskloet:
What's the reason for removing fromJSON as well? The Mozilla page doesn't mention it so I see no reason not to leave it for convenience.

PR #1313, opened 2015-08-13
WIP Improve fee estimation

### LINKED ISSUE #1309: Improve fee estimation in transactions
Fees should be estimated as a function of the projected size of the transaction without rounding up to the nearest kB.
See https://github.com/bitpay/bitcore/blob/master/lib/transaction/transaction.js#L865

braydonf:
- Calculates fees based on bytes instead of steps of kilobytes
- Will only estimate fee to include the size of the change output if the change is greater than dust amount thresholds
- Will not estimate fee below the minimum threshold
- Fixes  #1309


lib/util/js.js
```diff
@@ -80,5 +80,16 @@ module.exports = {
...
+  /**
+   * Checks that a value is an integer.
+   *
+   * @param {*} value
+   * @return {Boolean}
+   */
+  isInteger: function isInteger(value) {
+    return typeof value === 'number' &&
+      isFinite(value) &&
+      Math.floor(value) === value;
```
maraoz: Number.isInteger(value)?

braydonf: Nice! Switched to use that instead.

braydonf: Looks like the tests failed with Node.js v0.10, as `isInteger` is missing: https://travis-ci.org/bitpay/bitcore/jobs/75524680#L3598

Switched it back to use the util

PR #1316, opened 2015-08-14
Sequence number needs to be a nonmax value for locktime to be effective.

### LINKED ISSUE #1315: Correct Sequence Number for lockTime transactions
The default sequence number of all inputs is U_INT_MAX.

If a lock time is set on a transactions will all sequence numbers are U_INT_MAX, the transactions is immediately valid for a block.

Setting a locktime using the API does not change the sequence number of the inputs, this makes the locktime set essentially meaningless.

Proposal: Setting the locktime on a transaction should set the sequence number on all inputs to a none MAX value.

zmanian:
This sets all input sequence numbers to a non max value on setting a locktime

Proposal to fix #1315

### DIFF
--- a/lib/transaction/transaction.js
+++ b/lib/transaction/transaction.js
@@ -414,6 +414,11 @@ Transaction.prototype.lockUntilDate = function(time) {
   if (_.isDate(time)) {
     time = time.getTime() / 1000;
   }
+
+    for (i = 0; i < this.inputs.length; i++) {
+      this.inputs[i].sequence = Input.DEFAULT_LOCKTIME_SEQNUMBER;
+    }
+
   this.nLockTime = time;
   return this;
 };
@@ -433,6 +438,12 @@ Transaction.prototype.lockUntilBlockHeight = function(height) {
   if (height < 0) {
     throw new errors.Transaction.NLockTimeOutOfRange();
   }
+
+    for (i = 0; i < this.inputs.length; i++) {
+      this.inputs[i].sequence = Input.DEFAULT_LOCKTIME_SEQNUMBER;
+    }
+
+
   this.nLockTime = height;
   return this;
 };
--- a/lib/transaction/transaction.js
+++ b/lib/transaction/transaction.js
@@ -415,7 +415,7 @@ Transaction.prototype.lockUntilDate = function(time) {
     time = time.getTime() / 1000;
   }
 
-    for (i = 0; i < this.inputs.length; i++) {
+    for (var i = 0; i < this.inputs.length; i++) {
       this.inputs[i].sequence = Input.DEFAULT_LOCKTIME_SEQNUMBER;
     }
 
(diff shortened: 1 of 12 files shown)


maraoz:
why not set them to 0 instead of 0xFFFFFFF0?

zmanian:
Updated...

lib/transaction/transaction.js
```diff
@@ -414,6 +414,11 @@ Transaction.prototype.lockUntilDate = function(time) {
   if (_.isDate(time)) {
     time = time.getTime() / 1000;
   }
+
+    for (var i = 0; i < this.inputs.length; i++) {
```
braydonf: A few things:
- This may not serialize because I think the property for the sequence number on an input is  `.sequenceNumber`. 
- There should be tests to cover these statements
- Indentation is incorrect

zmanian:
Yep I'll write up the test cases and update. Thanks!

zmanian:
Just FYI,

Transaction 224660f2196fd22becb54f7c69f7a0ca07526e456276ef9f000c31a783275ab8 on testnet has 0 sequence number and was created with a locktime. Appears to be a correct tx.

Working on test cases for the pull request

zmanian:
updates with tests

braydonf:
Also please make sure to rebase on the current master.

braydonf:
Please make sure to rebase on master, this should clean up the commits and the diff.

test/transaction/transaction.js
```diff
@@ -748,6 +748,36 @@ describe('Transaction', function() {
         return new Transaction().lockUntilBlockHeight(-1);
       }).to.throw(errors.Transaction.NLockTimeOutOfRange);
     });
+    it('has a non-max sequenceNumber for effective date locktime tx', function() {
+      var transaction = new Transaction()
+        .from(simpleUtxoWith1BTC)
+        .lockUntilDate(date);
+      expect(transaction.inputs[0].sequenceNumber
```
braydonf: Using only `should` I think will work here (and the others), e.g.:
```js
transaction.inputs[0].sequenceNumber.should.not.equal(Transaction.Input.DEFAULT_SEQNUMBER);
```

test/transaction/transaction.js
```diff
@@ -748,6 +748,36 @@ describe('Transaction', function() {
...
+        .lockUntilBlockHeight(blockHeight);
+      expect(transaction.inputs[0].sequenceNumber
+        .should.not.equal(Transaction.Input.DEFAULT_SEQNUMBER));
+    });
+    it('should serialize correctly for date locktime ', function() {
+      var transaction= new Transaction()
+        .from(simpleUtxoWith1BTC)
+        .lockUntilDate(date);
+      var serialized_tx = transaction.uncheckedSerialize();
+      var copy = new Transaction(serialized_tx);
```
braydonf: I think it would also be good to check that the serialized version includes the correct sequenceNumber.

zmanian:
I'm just going to open a cleaner commit on a new pull request.

From correctness for the current Bitcoin protocol point of view, any non max sequence number would be correct. 

When/if CHECKSEQUENCEVERIFY gets implemented, any non-max field should be accepted in the field.

zmanian:
#1324 is a cleaner commit

PR #1317, opened 2015-08-14
Restore error when trying to derive a hardened path with HDPublicKey.

braydonf:
- Restores InvalidIndexCantDeriveHardened error in spec
- Passes hardened argument in HDPublicKey (in case the feature is attempted to be used similarly to HDPrivateKey)
- Fixes undefined error with InvalidLength
- Fixes tests to check for the error type
- Various formatting cleanup improvements


pnagurny:
LGTM

PR #1314, opened 2015-08-14
crypto-browserify library has ripemd160 support

braydonf:
See: https://github.com/crypto-browserify/crypto-browserify/issues/135

PR #1322, opened 2015-08-21
fix web worker issues

ryanxcharles:
bitcore should be able to run inside a web worker. There were two minor issues
preventing bitcore from running inside a web worker. The first was that lodash
was outdated, and that version of lodash had a problem with web workers. The
second was that the wrong version of ripemd160 was being called, because
global.window does not exist inside a web worker (global.self does instead). A
better way to check if you are in a browser is with process.browser.

This may not be a comprehensive list of issues, they're they are the ones I've noticed.


braydonf:
LTGM

We may also not need to check `process.browser` with: https://github.com/bitpay/bitcore/pull/1314

braydonf:
Landed in 0.13.1

PR #1331, opened 2015-09-04
add getDifficulty method to BlockHeader

pnagurny:


lib/block/blockheader.js
```diff
@@ -202,15 +204,24 @@ BlockHeader.prototype.toBufferWriter = function toBufferWriter(bw) {
  * @link https://en.bitcoin.it/wiki/Difficulty
```
braydonf: Docs need to be updated with the new first argument.

lib/block/blockheader.js
```diff
@@ -202,15 +204,24 @@ BlockHeader.prototype.toBufferWriter = function toBufferWriter(bw) {
...
+
+  var target = new BN(bits & 0xffffff);
+  var mov = 8 * ((bits >>> 24) - 3);
   while (mov-- > 0) {
     target = target.mul(new BN(2));
   }
   return target;
 };
 
+BlockHeader.prototype.getDifficulty = function getDifficulty(genesisBits) {
```
braydonf: Also needs JSdocs

lib/block/blockheader.js
```diff
@@ -202,15 +204,24 @@ BlockHeader.prototype.toBufferWriter = function toBufferWriter(bw) {
...
+  var difficulty1Target = this.getTargetDifficulty(genesisBits || GENESIS_BITS);
+  var currentTarget = this.getTargetDifficulty();
+
+  return parseFloat(difficulty1Target.div(currentTarget).toString(10));
```
braydonf: BN will never return a float.

lib/block/blockheader.js
```diff
@@ -212,6 +217,22 @@ BlockHeader.prototype.getTargetDifficulty = function getTargetDifficulty(info) {
 };
 
 /**
+ * @link https://en.bitcoin.it/wiki/Difficulty
+ * @param {Number} genesisBits
```
braydonf: It should noted that this is optional, and will default to the bits of livenet and testnet genesis bits.

braydonf: Also, the test case for passing in the bits isn't covered. I don't currently see a need for the argument yet, so another option would be to remove the argument, and we can add if needed later.

test/block/blockheader.js
```diff
@@ -248,6 +248,44 @@ describe('BlockHeader', function() {
...
+      });
+      x.getDifficulty().should.equal(43971662056.08958);
+    });
+
+    it('should throw an error if difficulty is higher than maximum javascript number', function() {
+      var x = new BlockHeader({
+        bits: 0x0900c2a8
+      });
+      (function() {
+        x.getDifficulty().should.throw('Difficulty exceeds maximum Javascript number');
```
braydonf: This test is passing, however `getDifficulty()` isn't throwing in this case.

braydonf:
Travis CI seems to not be running the tests, however I've run all the tests in Node.js v0.12 and v0.10 and iojs.

braydonf:
ACK

PR #1335, opened 2015-09-17
fix issue with validation within Script.isMultisigIn()

adjoinant:
I don't think signatures necessarily have a length of 0x47, so the logic in Script.isMultisigIn() is overly restrictive and returns false for valid txs e.g. tx 79209005e90fc446cb62636d230d67e6951ac6afd5fab2b224585a5192d50713


braydonf:
Great catch, thanks. This has been fixed in other methods as well.

test/script/script.js
```diff
@@ -347,7 +347,7 @@ describe('Script', function() {
       Script('OP_0 0x47 0x3044022002a27769ee33db258bdf7a3792e7da4143ec4001b551f73e6a190b8d1bde449d02206742c56ccd94a7a2e16ca52fc1ae4a0aa122b0014a867a80de104f9cb18e472c01').isMultisigIn().should.equal(true);
     });
     it('should identify multisig in 2', function() {
-      Script('OP_0 0x47 0x3044022002a27769ee33db258bdf7a3792e7da4143ec4001b551f73e6a190b8d1bde449d02206742c56ccd94a7a2e16ca52fc1ae4a0aa122b0014a867a80de104f9cb18e472c01 0x47 30450220357011fd3b3ad2b8f2f2d01e05dc6108b51d2a245b4ef40c112d6004596f0475022100a8208c93a39e0c366b983f9a80bfaf89237fcd64ca543568badd2d18ee2e1d7501').isMultisigIn().should.equal(true);
+      Script('OP_0 0x47 0x3044022002a27769ee33db258bdf7a3792e7da4143ec4001b551f73e6a190b8d1bde449d02206742c56ccd94a7a2e16ca52fc1ae4a0aa122b0014a867a80de104f9cb18e472c01 0x49 0x30450220357011fd3b3ad2b8f2f2d01e05dc6108b51d2a245b4ef40c112d6004596f0475022100a8208c93a39e0c366b983f9a80bfaf89237fcd64ca543568badd2d18ee2e1d7501').isMultisigIn().should.equal(true);
```
maraoz: could you add another test instead of modifying an existing one please?

maraoz:
Great find.

test/script/script.js
adjoinant: I can happily add another test if you'd like me to. However, I modified this test as it had incorrect data in it and stopped passing when I fixed the validation check.

maraoz: ACK

braydonf: Should the datapush opcode be 0x48 and not 0x49?

```
> var a = new Buffer('30450220357011fd3b3ad2b8f2f2d01e05dc6108b51d2a245b4ef40c112d6004596f0475022100a8208c93a39e0c366b983f9a80bfaf89237fcd64ca543568badd2d18ee2e1d7501', 'hex');
> a.length.toString(16)
'48'
```

adjoinant: Yes, that is correct - thanks. 
Changed and pushed commit

braydonf:
LGTM

PR #1337, opened 2015-09-23
Return zero for coinbase transaction fee.

braydonf:
See: https://github.com/bitpay/bitcore-node/blob/master/lib/services/address/history.js#L211

```js
var fees = !transaction.isCoinbase() ? transaction.getFee() : null
```

Can simply be:
```js
var fees = transaction.getFee();
```


pnagurny:
LGTM

PR #1340, opened 2015-09-30
Privacy improvements by sorting inputs and outputs

eordano:
See BIP69 for more details:
https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki

### DIFF
--- a/lib/util/buffer.js
+++ b/lib/util/buffer.js
@@ -170,6 +170,49 @@ module.exports = {
   hexToBuffer: function hexToBuffer(string) {
     assert(js.isHexa(string));
     return new buffer.Buffer(string, 'hex');
+  },
+
+  /**
+   * Browser shim for buffer compare
+   *
+   * MIT. Copyright by Feross Aboukhadijeh, and other contributors.
+   * Originally forked from an MIT-licensed module by Romain Beauxis.
+   * @see {https://github.com/feross/buffer/blob/bc33ea4618e6846148e35aee78dfc28d999bdd1c/index.js#L264}
+   */
+  compare: function(a, b) {
+    if (Buffer.compare) {
+      return Buffer.compare(a, b);
+    } else {
+      if (a === b) {
+        return 0;
+      }
+
+      var x = a.length;
+      var y = b.length;
+
+      var i = 0;
+      var len = Math.min(x, y);
+      while (i < len) {
+        if (a[i] !== b[i]) {
+          break;
+        }
+
+        ++i;
+      }
+
+      if (i !== len) {
+        x = a[i];
+        y = b[i];
+      }
+
+      if (x < y) {
+        return -1;
+      }
+      if (y < x) {
+        return 1;
+      }
+      return 0;
+    }
   }
 };
 
--- a/lib/util/buffer.js
+++ b/lib/util/buffer.js
@@ -170,49 +170,6 @@ module.exports = {
   hexToBuffer: function hexToBuffer(string) {
     assert(js.isHexa(string));
     return new buffer.Buffer(string, 'hex');
-  },
-
-  /**
-   * Browser shim for buffer compare
-   *
-   * MIT. Copyright by Feross Aboukhadijeh, and other contributors.
-   * Originally forked from an MIT-licensed module by Romain Beauxis.
-   * @see {https://github.com/feross/buffer/blob/bc33ea4618e6846148e35aee78dfc28d999bdd1c/index.js#L264}
-   */
-  compare: function(a, b) {
-    if (Buffer.compare) {
-      return Buffer.compare(a, b);
-    } else {
-      if (a === b) {
-        return 0;
-      }
-
-      var x = a.length;
-      var y = b.length;
-
-      var i = 0;
-      var len = Math.min(x, y);
-      while (i < len) {
-        if (a[i] !== b[i]) {
-          break;
-        }
-
-        ++i;
-      }
-
-      if (i !== len) {
-        x = a[i];
-        y = b[i];
-      }
-
-      if (x < y) {
-        return -1;
-      }
-      if (y < x) {
-        return 1;
-      }
-      return 0;
-    }
   }
 };
 
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "bip21",
     "bip32",
     "bip37",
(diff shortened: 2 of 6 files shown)


maraoz:
nice! ping @kristovatlas

dcousens:
You could also use https://github.com/bitcoinjs/bip69 to save yourself writing more code and tests,  its written generically.

dcousens:
It'd just mean mapping `.satoshis` to `.value` and `.outputIndex` to `.vout`.
You could do that inline or if its really terrible,  could make alternate bindings in `bip69`.
Happy to add maintainers if necessary,  juts thought it might be useful.

kristovatlas:
Nice :)

eordano:
Thanks @dcousens! So, I see two issues with using directly bip69:
- Mapping would mean adding new properties, and that makes node jit classes less efficient
- We use buffers for storing the script, not strings, so the `localeCompare` function won't work

If it's OK with you, I added a comment linking back to bip69 and copied over the fixtures. I feel weird about adding mappings for bitcore to bip69 because of item number two above... what do you think?

package.json
```diff
@@ -68,6 +68,7 @@
     "bip21",
     "bip32",
     "bip37",
+    "bip39",
```
braydonf: Did you mean to add bip69 here?

braydonf:
Great! I agree, where we can combine our efforts the better. I do think having a `sort()` (or similar) method on Transaction makes sense.

package.json
eordano: Fixed!

dcousens:
> We use buffers for storing the script, not strings, so the localeCompare function won't work

The `localeCompare` is used for the transaction hashs, but certainly that could be changed to make use of `compare`,  its just the comparison would have to be done on the reversed values for it to be correct.

In fact,  your implementation,  if `Buffer`s: `BufferUtil.compare(first.prevTxId, second.prevTxId)`,  is wrong if its sorting by ascending.

> If it's OK with you, I added a comment linking back to bip69 and copied over the fixtures. 
Awesome,  maybe I should expose them for you so you can version it with a `devDependency`? `bip69/fixtures`?

lib/util/buffer.js
```diff
@@ -170,6 +170,49 @@ module.exports = {
...
     return new buffer.Buffer(string, 'hex');
+  },
+
+  /**
+   * Browser shim for buffer compare
+   *
+   * MIT. Copyright by Feross Aboukhadijeh, and other contributors.
+   * Originally forked from an MIT-licensed module by Romain Beauxis.
+   * @see {https://github.com/feross/buffer/blob/bc33ea4618e6846148e35aee78dfc28d999bdd1c/index.js#L264}
+   */
```
dcousens: Perhaps you could use `buffer-compare` for this?
Seems odd to constantly bring in code from other libraries without appropriate tests.

eordano: Agree. @braydonf would you be OK with a new dependency?

We used to have `buffer-utils` as a dep back in a previous version. I think it was removed for a reason I don't remember.

braydonf:
> However, if prevTxIds are already in the reversed state (which they may be, I'm not familiar with bitcore internals) then it should be fine.

The `prevTxIds` are in [BE state](https://github.com/bitpay/bitcore/blob/master/lib/transaction/input/input.js#L90), the only buffers that I'm aware of that are LE still are the block header hash buffers, and the bitcore-p2p inventory buffers.

lib/util/buffer.js
braydonf: Considering that it's mostly to support Node.js v0.10, it may be okay to have this inlined. However I do agree, we should have tests that cover it.

eordano:
@braydonf I added the `buffer-compare` dep which is very very small and already tested. PTAL! https://github.com/soldair/node-buffer-compare

I don't know how confortable you feel with the `var compare = Buffer.compare || require('buffer-compare');` line...

I kinda changed my mind because it doesn't add much weight to the library, it's already tested, and it's faster if we don't add more tests to bitcore.

dcousens:
> The prevTxIds are in BE state, the only buffers that I'm aware of that are LE still are the block header hash buffers, and the bitcore-p2p inventory buffers.

By BE state, do you mean regular byte order [as they come from the hash function]?
Its [unfortunately quite an ambiguous statement](https://github.com/kristovatlas/bips/pull/1#issuecomment-135306343).
If they are straight from the hash function,  then the comparison should be inverted (though I'd be surprised if the tests didn't catch that?)

braydonf:
@eordano:

With only using:
`var compare = Buffer.compare` 

instead of:
`var compare = Buffer.compare || require('buffer-compare');`

all of the tests pass in Node.js v0.12.7 and v4.1.2 as well as Firefox 41.0 and Chromium 45, however it won't pass in Node.js v0.10.

The compare function is already included as a part of the of browserify build for Buffer from https://github.com/feross/buffer/blob/master/index.js#L264 (that was included in an earlier commit). So we may not need to double up there.

eordano:
@braydonf not sure if I wouldn't drop node 0.10 support, would you?

braydonf:
Yeah, we should keep support for 0.10 currently.

eordano:
Let me know if you won't give green light to this because of the extra dep. I think it's better than inlining it.

Privacy is important in bitcoin, and this feature is important for bitcore, as the many wallets that use could follow the same strategy very easily (which is the spirit of the BIP).

braydonf:
LGTM

I went through and modified the test vectors locally to make sure that sorting wasn't passing erroneously, and all is passing.

braydonf:
Checked the browser builds, and `Buffer.compare` from https://github.com/feross/buffer is used in those cases.

fanatid:
@braydonf can you please tell why you should keep support 0.10? bitpay uses 0.10 in some projects that use bitcore?

braydonf:
Yep, there are bitcore dependencies, sub-modules, running tests against 0.10 currently, and I think it's still fairly common to do so still.

dcousens:
@braydonf is that a NACK on https://github.com/bitcoinjs/bip69/issues/1?

braydonf:
@dcousens Since the sort methods are fairly small, I think it makes sense to implement within Transaction.

PR #1345, opened 2015-10-12
allow Transaction.to take array (#1334)

### LINKED ISSUE #1334: Transaction.to like Transaction.from
`Transaction.from` may accept `array`, but `Transaction.to` accept only `(address, amount)`.. I think will be helpful if `Transaction.to` could accept form like this: `[{address: string, amount: number}, ...]`, thoughts?

fanatid:
Resolve #1334, but need discuss.

### DIFF
--- a/lib/transaction/transaction.js
+++ b/lib/transaction/transaction.js
@@ -481,6 +481,14 @@ Transaction.prototype._newTransaction = function() {
 /* Transaction creation interface */
 
 /**
+ * @typedef {Object} Transaction~fromObject
+ * @property {string} prevTxId
+ * @property {number} outputIndex
+ * @property {(Buffer|string|Script)} script
+ * @property {number} satoshis
+ */
+
+/**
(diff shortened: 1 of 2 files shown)


braydonf:
I think it makes sense to add this, good idea. I think it would be good to add new tests, or leave some of the tests with the chained `to` methods, as both should work.

fanatid:
I changed some tests, now both form work at the one test for example.

braydonf:
Also need to update the jsdocs at: https://github.com/bitpay/bitcore/pull/1345/files#diff-3ff19b4e31a68503f7ddd89615e606d7R700 to include that an array with objects with address and satoshis, can also be sent.

fanatid:
Not sure that my jsdoc is right.

braydonf:
Interesting, looking at the jsdocs reference, it appears correct. We haven't used typedef yet.

maraoz:
LGTM!

PR #1346, opened 2015-10-15
fix Script.set

fanatid:
For example problem with `ad0a426817666f56d5b4fba97177f455bbfba3b80faca32a58ae9f947896ec68` from testnet:
```js
var bitcore = require('bitcore')
var tx = bitcore.Transaction('0100000001cbf75e37f8f57f338dd435d564c7302a02d9c6d03e26c8f0f24fa46052c0c0d1010000008b48304502210095cf913f96c52242b3a88dc59a9aed2a568dc80446541348c89c5430727d548402206adbe00b00db7d71458b990ab69090a038244a6b94faca98652656fd0b9be4e4014104dac5c4da1fd483b8655ab6a5d8e5dd2226ef20d1023d7ff38df29bfa8fcfdf73fb99a07efa0c66112c1fa438a8ed8571364bc04f21eaadf54c4918d43aba8ed3ffffffff02804f1200000000001976a914d40f073684cfac11be0366dc6a56a8d3cc87ece488ac1027000000000000044e6f6e6500000000')
console.log(tx.outputs[1]) // <Output (10000 sats) 4e6f6e65>
var script = bitcore.Script(tx.outputs[1].script) // TypeError: Cannot read property 'chunks' of null
```

### DIFF
--- a/lib/script/script.js
+++ b/lib/script/script.js
@@ -36,15 +36,17 @@ var Script = function Script(from) {
     return Script.fromAddress(from);
   } else if (from instanceof Script) {
     return Script.fromBuffer(from.toBuffer());
-  } else if (typeof from === 'string') {
+  } else if (_.isString(from)) {
     return Script.fromString(from);
-  } else if (typeof from !== 'undefined') {
+  } else if (_.isObject(from) && _.isArray(from.chunks)) {
     this.set(from);
   }
 };
 
 Script.prototype.set = function(obj) {
-  this.chunks = obj.chunks || this.chunks;
+  $.checkArgument(_.isObject(obj));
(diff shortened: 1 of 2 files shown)


braydonf:
Looks like this is another case where the script is invalid and the instance of Script can't be initialized, another here with a test: https://github.com/bitpay/bitcore-lib/blob/461852e10412959396c633b8a769b45b2b3ea5c8/test/transaction/output.js#L135

`.script` in this case returns null, `._scriptBuffer` will have the actual data. What's the use case for a script with null?

fanatid:
moved to bitcore-lib repository

PR #1355, opened 2015-10-20
Fix double bitcore-lib loading issue

braydonf:
- Pass along services location and additional services for the default.
- Reference specific commit of bitcore-build and remove mode hints.

### DIFF
--- a/bin/bitcore
+++ b/bin/bitcore
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
-// vi: ft=javascript -*- mode: JavaScript; -*-
 
-var node = require('bitcore-node');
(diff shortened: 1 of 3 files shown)


bin/bitcore
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
-// vi: ft=javascript -*- mode: JavaScript; -*-
 
-var node = require('bitcore-node');
-node.cli.bitcore();
+var bitcore = require('bitcore-node/lib/cli/bitcore');
```
maraoz: :see_no_evil:

gabegattis:
we tested it and it worked

PR #1359, opened 2015-10-21
improved upstart script

bitpayalan:
* Sends proper signal when shutting down
* Better start event so that it starts at proper time
* Additional comments for clarity


braydonf:
LGTM

PR #1427, opened 2016-10-23
update license

greenbigfrog:
Is it "lawyer" correct/confirm like this?


greenbigfrog:
Sorry, github automatically makes me compare to this repo... (I think I've learned it enough now)

(BTW, the license info goes only till 2015, and we are already 10 months into 2016)

PR #1451, opened 2017-05-18
Documentation upgrades

arshbot:
Added prerequisite section which bitcore-node has but isn't easily visible
Added '$' at the beginning of every line of bash lines


DanielRHayes:
@nitsujlangston I think we can close this

PR #1455, opened 2017-06-07
Add info about cURL requirement.

anduck:
Add info about cURL requirement.


DanielRHayes:
Won't include for now.  We are going to be making lots of changes to docs for v8

PR #1512, opened 2018-04-09
Broacast method for BTCStateProvider

micahriggan:
Broadcasting a transaction via posting to 
```
/{chain}/{network}/tx/send
```
 should call the sendrawtransaction RPC method

Builds on https://github.com/bitpay/bitcore/pull/1518


nitsujlangston:
Need to have rpc pull from config

PR #1517, opened 2018-04-16
WIP Adding wallet utils and authentication

nitsujlangston:
# Goals:

- Provide client library and CLI utils for wallet operations (able to handle millions of imported keys in wallet client storage)
- Stop using mongo id as wallet id
- Authenticate wallet requests using payload signatures, signed by request auth key (derived from xPubKey on set path)


packages/bitcore-node/lib/routes/wallet.js
```diff
@@ -2,27 +2,62 @@ const util = require('util');
...
 const _ = require('underscore');
+const secp256k1 = require('secp256k1');
 
 const logger = require('../logger');
 const ChainStateProvider = require('../providers/chain-state');
+const bitcoreLib = require('bitcore-lib');
+
+const verifyRequestSignature = (params) => {
+  const { message, pubKey, signature} = params;
+  const pub = (new bitcoreLib.HDPublicKey(pubKey)).deriveChild('m/2').publicKey.toString('hex');
```
nitsujlangston: Need to remove these toStrings - they were from debugging

micahriggan:
Initial glance over seems good. I like the authenticate middleware. 

Is the idea that the new bitcore-client thing will replace bitcore-wallet? 

As for the converting from (arg, arg, arg) to (params), I think that's cool, but in the future we're going to have to read the code to see what's being pulled off of params.

Perhaps we should write the new stuff in TypeScript. I think it'd be beneficial.

packages/bitcore-node/lib/routes/wallet.js
nitsujlangston: fixed

packages/bitcore-node/server.js
```diff
@@ -25,7 +25,7 @@ async.series([
     await Promise.all(p2pServices.map(p2pService => p2pService.start()));
   }
 ], function () {
-  if (cluster.isWorker) {
+  if (!cluster.isWorker) {
```
micahriggan: Debug line

packages/bitcore-node/lib/routes/wallet.js
```diff
@@ -2,27 +2,62 @@ const util = require('util');
...
   }
 });
 
-router.get('/:walletId', async function(req, res) {
+router.get('/:walletId', authenticate, async function(req, res) {
   try {
-    let {chain, network, walletId} = req.params;
-    let wallet = await ChainStateProvider.getWallet(chain, network, walletId);
+    let {chain, network, pubKey} = req.params;
+    let wallet = await ChainStateProvider.getWallet({ chain, network, pubKey });
```
micahriggan: Wouldn't you have req.wallet here?

packages/bitcore-client/bin/wallet-import
```diff
@@ -0,0 +1,26 @@
...
+  .option('--path <path>', 'REQUIRED - Where wallet is stored')
+  .parse(process.argv);
+
+const main = async () => {
+  const {path} = program;
+  let wallet;
+  try {
+    wallet = await Wallet.loadWallet({ path });
+    const keyInput = [
+      { address: 'mhse1Lz4otzATQWXGo7bGp2apbQGubvnjE', privKey: 'cNZPB2AgNpNjcLQQ9zjuCGgPBGP2ZNyNav6JLVHPUvHBByKruHT9' },
```
micahriggan: Debug lines

packages/bitcore-node/lib/routes/wallet.js
```diff
@@ -34,33 +69,23 @@ router.get('/:walletId', async function(req, res) {
 
 router.get('/:walletId/addresses', async function(req, res) {
```
micahriggan: Needs to have pubKey here

packages/bitcore-node/lib/routes/wallet.js
```diff
@@ -150,10 +175,10 @@ ListTransactionsStream.prototype._transform = function(transaction, enc, done) {
   done();
 };
 
-router.get('/:walletId/transactions', async (req, res) => {
-  let {walletId, chain, network} = req.params; 
+router.get('/:pubKey/transactions', async (req, res) => {
```
micahriggan: Why doesn't this have authenticate, but the utxos does?

nitsujlangston: it should, all wallet methods should. Create wallet is the only one that has diminished value, but even that could if the incoming pubkey is used for verification

PR #1518, opened 2018-04-16
v8 Better config resolution

micahriggan:
# Goal
Load a config file based off of different paths

## Sources
- ENV, for Docker
- --config argument
- Search path (home,  bitcore, and bitcore-node)

# Breaking Changes
After this is merged you'll need to rename config.json to bitcore.config

You'll also need to put the entire config object in 
```
{
  bitcoreNode: oldConfig
}
```


micahriggan:
for https://github.com/bitpay/bitcore/projects/1#card-8880904

PR #1534, opened 2018-05-18
Separate P2P from BTC/BCH to allow more chains.

illegalprime:
This introduces a new P2P interface that can be implemented by different chains.
What's new:

 - One can implement this new interface to be able to monitor arbitrary chains that are close in structure to bitcoin.
 - Old code is updated to ES6 (w/ async, await) and with this some order of execution issues are fixed.
 - Unit tests for the P2P Abstraction.
 - End to end (p2p node to mongo) tests!
 - Fixes to docker compose end-to-end test bed.
 - More efficient cache for syncing and watching blocks.

Crude benchmarks show syncing is now **~11% faster!**


packages/bitcore-node/src/services/p2p/index.ts
```diff
@@ -0,0 +1,268 @@
...
+        syncer.add(pair.block.hash);
+      }
+      else {
+        logger.info(`Added block ${pair.block.hash}`, {
+          chain: this.chain,
+          network: this.network,
+        });
+      }
+      return pair;
+    })).pipe(share());
```
micahriggan: not sure if I get need for share() here. Is this to make the Observable fire off without a subscriber?

micahriggan: I've used publish and connect before, is this related? 
 https://www.learnrxjs.io/operators/multicasting/publish.html

packages/bitcore-node/src/services/p2p/index.ts
```diff
@@ -0,0 +1,268 @@
...
+  // disconnects from peer and stops all pending tasks,
+  // afterwards `start()` can be called again.
+  stop(): Promise<void>;
+
+  // when `true` only emit blocks that result from the syncing process
+  syncing: boolean;
+}
+
+
+export type StandardP2p = P2pService<Bitcoin.Block, Bitcoin.Transaction>;
```
micahriggan: You can call this BitcoinP2p

packages/bitcore-node/src/services/p2p/index.ts
```diff
@@ -0,0 +1,256 @@
...
+  } = {
+    BCH: () => new BtcP2pService(config),
+    BTC: () => new BtcP2pService(config),
+  };
+  logger.debug(`Building p2p service for ${chain}.`)
+  return namesToChains[chain]();
+}
+
+
+export async function start() {
```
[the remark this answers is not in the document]
illegalprime: I only moved it here because of: https://github.com/bitpay/bitcore/blob/v8.0.0/packages/bitcore-node/src/server.ts#L16

packages/bitcore-node/src/server.ts
```diff
@@ -1,34 +1,17 @@
-import { P2pService } from './services/p2p';
+import * as P2P from './services/p2p';
```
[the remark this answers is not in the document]
illegalprime: I don't understand when a p2p service will get configured, is it already coming out pre-configured from P2pProvider?

PR #1545, opened 2018-05-24
Feature/encrypting keys

micahriggan:
# Purpose
To add encryption for the following:
* The masterKey, or the privatekey that other keys can be derived from
* The encryption key
* The privateKeys stored in levelDB

## Encryption Algorithms
We use aes-256-cbc for encryption of privatekeys and the encryptionKey

### Encryption Key
- Each wallet has an encryption key of 32 random bytes. 
- This encryption key is stored in the wallet encrypted
- We encrypt the encryption key with a passphrase.
  - We take the SHA512 of the passphrase, and use the first 32 bytes as the key, and the remaining 16 bytes as the IV.

### PrivateKeys
- We use the decrypted encryption key to encrypt the private keys stored in the wallet file and in levelDB
  - We use the encryptionKey as the key
  - We use the SHA256(SHA256(pubKey))[0:16] as the IV 

# Side effects / Future improvements
Currently interacting with wallet endpoints required the wallet signing key, which is only available when the wallet has been unlocked. If this is not desired, we could expose this signing key unencrypted.

# Not in scope
signing of transactions. 
This feature hasn't been tested with this MR


packages/bitcore-client/bin/wallet-import
```diff
@@ -7,6 +7,7 @@ const Wallet = require('../lib/wallet');
 program
   .version(require('../package.json').version)
   .option('--path <path>', 'REQUIRED - Where wallet is stored')
+  .option('--password <password>', 'REQUIRED - password to unlock the wallet so the private keys can be encrypted')
```
nitsujlangston: we need to prompt for the password, don't want that in your bash history ;)

micahriggan: Makes sense! Other than that looks good? I have another coming soon with the level dB storage and retrieval working

PR #1549, opened 2018-05-30
Create storage worker on the same node as P2P workers.

illegalprime:
This only reverts bitcore-node/src/server.ts of commit 37931b0df1129bd776db84a2fd5f3cb2261642c9, to fix p2p syncing. Without this p2p syncing is stuck at the after it gets headers (but works in debug mode, where this is not a problem).


DanielRHayes:
works for me now on regtest on my mac.  I'll check on docker-compose in a bit

PR #1554, opened 2018-06-01
P2P uses adapters to remove BTC specifics from bitcore.

illegalprime:
Use adapters to further isolate chain specifics from p2p and the rest of `bitcore-node`.


illegalprime:
@HarshitJoshi I made changed to your `block` unit tests, they might be wrong or test the wrong thing, I'm not sure. Take a look and lmk:
https://github.com/bitpay/bitcore/pull/1554/files#diff-9795175bc5d4257929f89ed920bcefae

micahriggan:
We ended up going in a different direction for p2p with the preference towards each chain implementing its own bitcorelib, closing for now

PR #1556, opened 2018-06-01
Feature/api as a service

DanielRHayes:
Moved API from `server.ts` to it's own service.  More organized IMO.


micahriggan:
Let's get API starting storage

PR #1576, opened 2018-06-06
[WIP] Verify entire DB against BTC node.

### LINKED ISSUE #1565: E2E Testing  - Validator for database
Validate every record in the database

illegalprime:
Tests to verify our db to check if all our state matches up (useful when checking for regressions).

- [x] `BlockModel`
- [x] `TransactionModel`
- [ ] `CoinModel`

Fixes #1565


micahriggan:
What's the status on this one?

micahriggan:
@illegalprime could you give an example of how to run this? Is there a npm run verify or what?

illegalprime:
@micahriggan run with `npm run test:verify`

packages/bitcore-node/test/util/verify.ts
```diff
@@ -0,0 +1,232 @@
...
+      for (const our of ours) {
+        expect(our.spentHeight, 'tx spent height').to.equal(tx.blockHeight);
+        expect(tx.wallets).to.include.members(Array.from(our.wallets));
+      }
+    }
+  }
+}
+
+
+if (require.main === module) (async () => {
```
micahriggan: This looks weird

illegalprime: nah it's good

illegalprime: https://nodejs.org/docs/latest/api/all.html#modules_accessing_the_main_module

DanielRHayes: So it's something that is exported (so it can be required), but if it is called directly it runs as a script

DanielRHayes: To accomplish the same thing you could move the logic to another file that exports and have the scripts just require and run the file

PR #1579, opened 2018-06-07
Simplify wallet transaction export

nitsujlangston:
* Stops using $lookup, since there were issues with passing chain and network to the coin query
* Removes irrelevant warning about missing change addresses


micahriggan:
closed #1579

PR #1581, opened 2018-06-07
Formatting and cleaning up wallet-derive script

micahriggan:


packages/bitcore-client/bin/multi-sig/wallet-derive-address
```diff
@@ -64,23 +64,39 @@ function getAllAddresses(publicKeyRing, numberOfAddresses, isChange, m, network)
   return addresses;
 }
 
+async function readFromPrompt() {
+  let counter = 1;
+  let xpubkey = await promptly.prompt(`Enter xpubkey ${counter} `);
+  let promptResults = [];
+  while (xpubkey != '') {
```
DanielRHayes: `xpubkey.trim() !== ''` so we can bail if they just put in spaces?

packages/bitcore-client/bin/multi-sig/wallet-derive-address
```diff
@@ -64,23 +64,39 @@ function getAllAddresses(publicKeyRing, numberOfAddresses, isChange, m, network)
   return addresses;
 }
 
+async function readFromPrompt() {
```
DanielRHayes: I can't comment on the correct lines.  Can you remove the `console.log` on line 62.  And the comments on line 45 to 50.  This afternoon a way to configure this to use configurable derivation strategies.

packages/bitcore-client/bin/multi-sig/wallet-derive-address
```diff
@@ -64,23 +64,39 @@ function getAllAddresses(publicKeyRing, numberOfAddresses, isChange, m, network)
...
-      let xpubkey = await promptly.prompt(`Enter xpubkey ${counter} `);
-      while (xpubkey != '') {
-        keyInput.push(xpubkey);
-        counter++;
-        xpubkey = await promptly.prompt(`Enter xpubkey ${counter} `, { default: '' });
-      }
+      keyInput = await readFromPrompt();
     }
     const mOfN = Number(m);
     let results = getAllAddresses(keyInput, depth, false, mOfN, network);
```
DanielRHayes: on line 105 you should do `keyInput.length < mOfN' which will be a number instead of a string

packages/bitcore-client/bin/multi-sig/wallet-derive-address
```diff
@@ -64,23 +64,39 @@ function getAllAddresses(publicKeyRing, numberOfAddresses, isChange, m, network)
...
-      let counter = 1;
-      let xpubkey = await promptly.prompt(`Enter xpubkey ${counter} `);
-      while (xpubkey != '') {
-        keyInput.push(xpubkey);
-        counter++;
-        xpubkey = await promptly.prompt(`Enter xpubkey ${counter} `, { default: '' });
-      }
+      keyInput = await readFromPrompt();
     }
     const mOfN = Number(m);
```
DanielRHayes: if m is not a number then throw

PR #1589, opened 2018-06-11
Simplify Wallet Create

DanielRHayes:
Simplified the code in wallet create a bit to make it more readable.  @micahriggan I haven't tested by running the utils with this, can we do that when we are doing the other wallet utils run throughs?


micahriggan:
Closing after incorporating this feedback into a recent MR where authKey was added to the unencrypted wallet

Issue #1586, opened 2018-06-08
v8 Wallet Client Config


micahriggan:
We should have a config file that stores the path to wallets, and stores the api url to hit

HarshitJoshi:
#1593

PR #1593, opened 2018-06-13
Feature/wallet config

HarshitJoshi:
adds a config and listWallets scripts that lists out all the wallets in a given path


HarshitJoshi:
Fixes #1586

PR #1599, opened 2018-06-19
No Mongoose

### LINKED ISSUE #1587: Better Typescript Mongoose Models
Currently the typescript models are hacky interfaces being applied to a mongoose schema. 
We should be able to do something like the following to remove all the interfaces 

```
class BlockModel extends Schema {


}
```

micahriggan:
Replacing mongoose and using mongo's typings, which are pretty good.

Closes #1587

### DIFF
--- a/packages/bitcore-node/src/models/block.ts
+++ b/packages/bitcore-node/src/models/block.ts
@@ -27,10 +27,13 @@ export type IBlock = {
 }
 
 @LoggifyClass
-class Block extends BaseModel<IBlock> {
+export class Block extends BaseModel<IBlock> {
 
   constructor() {
     super('blocks');
+  }
+
+  async onConnect() {
     this.collection.createIndex({ hash: 1 });
     this.collection.createIndex({ chain: 1, network: 1, processed: 1, height: -1 });
     this.collection.createIndex({ chain: 1, network: 1, timeNormalized: 1 });
@@ -207,8 +210,11 @@ class Block extends BaseModel<IBlock> {
 
 
 
-  _apiTransform( block: IBlock, options: TransformOptions) {
-    let transform = {
+  _apiTransform( block: IBlock, options: TransformOptions): IBlock | string {
+    let transform: IBlock = {
+      chain: block.chain,
+      network: block.network,
+      processed: block.processed,
       hash: block.hash,
       height: block.height,
       version: block.version,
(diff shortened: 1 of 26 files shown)


packages/bitcore-node/src/models/block.ts
```diff
@@ -27,251 +26,226 @@ export interface IBlock {
...
+    chain: string;
+    network: string;
+  }) {
+    const { block, chain, network, parentChain, forkHeight } = params;
+    const header = block.header.toObject();
+    const blockTime = header.time * 1000;
+
+    await this.handleReorg({ header, chain, network });
+
+    const previousBlock = (await this.collection.findOneAndUpdate({
```
micahriggan: This was a logic change, does this make sense to do?

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -26,126 +24,94 @@ export interface ITransaction {
...
+
+    let mintWallets = await CoinModel.collection
+      .aggregate([
+        {
+          $match: { mintTxid: { $in: txids }, chain, network }
+        },
+        { $unwind: "$wallets" },
+        { $group: { _id: "$mintTxid", wallets: { $addToSet: "$wallets" } } }
+      ])
+      .toArray() as Array<ICoin & {_id: string}>;
```
micahriggan: might be able to pass type into aggregate

packages/bitcore-node/src/providers/chain-state/internal/transforms.js
```diff
@@ -1,9 +1,10 @@
+const { CoinModel } = require('../../../models/coin');
 const {Transform} = require('stream');
 const util = require('util');
 const _ = require('underscore');
-const logger = require('../../../logger');
-const mongoose = require('mongoose');
-const Coin = mongoose.model('Coin');
+/*
+ *const logger = require('../../../logger');
```
micahriggan: can remove

packages/bitcore-node/src/services/storage.ts
```diff
@@ -1,20 +1,26 @@
...
 import logger from '../logger';
 import config from '../config';
 import { LoggifyClass } from "../decorators/Loggify";
+import { MongoClient, Db } from "mongodb";
 import "../models"
 
 @LoggifyClass
 export class StorageService {
-  start(args: any): Promise<any> {
+  client: MongoClient | undefined;
```
micahriggan: Switch to optional

PR #1618, opened 2018-06-29
fix for commander not handling missing required flag

HarshitJoshi:
this fixes the missing flag not throwing an error by commander while running wallet utils


packages/bitcore-client/lib/program.js
```diff
@@ -0,0 +1,20 @@
...
+  _parse(args);
+
+  const requiredOptions = program.options.filter(opt => opt.required && opt.required !== 0);
+
+  const options = requiredOptions.map(option => {
+    return option.long.replace('--', '');
+  });
+
+  if (!!options.some(element => !Object.getOwnPropertyNames(program).includes(element))) {
+    throw new Error(`Missing a required flag`);
```
micahriggan: Go ahead and put the option in this message so we know which flag was missing.

packages/bitcore-client/lib/program.js
```diff
@@ -0,0 +1,24 @@
...
+program.parse = (args) => {
+  _parse(args);
+
+  const requiredOptions = program.options.filter(opt => opt.required && opt.required !== 0);
+
+  const options = requiredOptions.map(option => {
+    return option.long.replace('--', '');
+  });
+
+  const match = !options.some(element => !Object.getOwnPropertyNames(program).includes(element));
```
micahriggan: This some logic looks duplicated below

PR #1629, opened 2018-07-05
Feature/estimate smart fee

HarshitJoshi:
add estimateSmartFee RPC call to calculate total transaction fee required for the transaction in wallet-tx util


HarshitJoshi:
Fixes #1630

packages/bitcore-client/bin/wallet-tx
```diff
@@ -8,23 +8,39 @@ program
...
+  .option('--target [target]', 'optional - custom target block for confirmation')
   .option('--path [path]', 'optional - Custom wallet storage path')
   .parse(process.argv);
 
 const main = async () => {
   const { name, path } = program;
   let wallet;
   try {
-    const { recipients, utxos, change, fee } = program;
+    const { recipients, utxos, change, fee, target=2 } = program;
```
micahriggan: I don't think you need to default here since you've got some defaults in the method.

packages/bitcore-client/bin/wallet-tx
```diff
@@ -8,23 +8,39 @@ program
...
+
+    const parsedUtxos = utxos.includes('[') ? JSON.parse(utxos) : [];
+    const parsedRecipients = recipients.includes('[') ? JSON.parse(recipients) : [];
+
+    if (!parsedRecipients.length || !parsedUtxos.length) {
+      throw new Error('invalid arguments');
+    }
+
+    const calculatedNetworkFeeKb = await wallet.getNetworkFee({ target });
+    const netWorkFeeSat = calculatedNetworkFeeKb.feerate * 1E8/1000;
```
micahriggan: Seems weird to have 1E8/1000

This is to convert the returned data from BTC to sats?

packages/bitcore-client/bin/wallet-tx
```diff
@@ -8,23 +8,39 @@ program
...
+
+    const transactionSize = (parsedUtxos.length * 148 + parsedRecipients.length * 34 + 10 + parsedUtxos.length);
```
micahriggan: let's break some of these values into constants above so we can see what these numbers mean

const utxoBytes = 148;
const outputBytes = 34;

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -224,6 +224,19 @@ export class InternalStateProvider implements CSP.IChainStateService {
     Storage.apiStreamingFind(CoinModel, query, stream);
   }
 
+  async getFee(params: CSP.GetEstimateSmartFeeParams) {
```
micahriggan: Let's go ahead and use the asyncCall method :)

packages/bitcore-node/src/rpc.ts
```diff
@@ -120,4 +120,9 @@ export class RPC {
   async blockAsync(hash: string): Promise<any> {
     return await this.asyncCall('getblock', [hash]);
   }
+
+  getEstimateSmartFee(target: number, callback: CallbackType) {
+    this.callMethod('estimatesmartfee', [target], callback);
```
micahriggan: lets switch this part to asyncCall, so we don't have to wrap it in a promise later

packages/bitcore-node/src/types/namespaces/ChainStateProvider.ts
```diff
@@ -27,6 +27,9 @@ export declare namespace CSP {
   export type GetBlockParams = ChainNetwork & {
     blockId: string;
   };
+  export type GetEstimateSmartFeeParams = ChainNetwork & {
```
micahriggan: Lets move this down below getBlocks, I think the typings may be declared in order

packages/bitcore-client/bin/wallet-tx
HarshitJoshi: yes, that's how it's done over frontend for insight @SonicWizard mentioned this

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
HarshitJoshi: resolved

packages/bitcore-node/src/rpc.ts
HarshitJoshi: resolved

packages/bitcore-client/bin/wallet-tx
HarshitJoshi: resolved

packages/bitcore-node/src/types/namespaces/ChainStateProvider.ts
HarshitJoshi: resolved

packages/bitcore-client/bin/wallet-tx
HarshitJoshi: would also like to get feedback from you and @nitsujlangston about the conversion part

packages/bitcore-client/bin/wallet-tx
HarshitJoshi: resolved

PR #1642, opened 2018-07-14
Feature/stream experiment

micahriggan:
Tried to simplify the logic of the transform, so I could understand it better.

One major change, is creating "sends" by looking at the inputs, instead of the outputs.

@nitsujlangston can you tell me if this change is correct?

Major issues with mongo 3.0.1, driver would randomly kill cursors.

3.1.1 fixes this issue.

### DIFF
--- a/packages/bitcore-node/package.json
+++ b/packages/bitcore-node/package.json
@@ -72,7 +72,7 @@
     "cors": "^2.8.4",
     "express": "^4.16.3",
     "mkdirp": "^0.5.1",
-    "mongodb": "^3.0.10",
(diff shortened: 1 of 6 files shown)


packages/bitcore-node/package.json
```diff
@@ -72,7 +72,7 @@
     "cors": "^2.8.4",
     "express": "^4.16.3",
     "mkdirp": "^0.5.1",
-    "mongodb": "^3.0.10",
+    "mongodb": "^3.1.10",
```
nitsujlangston: 3.1.1

PR #1653, opened 2018-07-26
Adding fork copy logic

micahriggan:


packages/bitcore-node/src/services/p2p.ts
```diff
@@ -279,10 +285,64 @@ export class P2pService {
...
+    StateModel.collection.findOneAndUpdate(
+      {},
+      { $addToSet: { initialSyncComplete: `${chain}:${network}` } },
+      { upsert: true }
+    );
     return true;
   }
 
+  async createFork(newChain: string, parentChain: string, forkHeight: number) {
+    const blockCursor = BlockModel.collection.find({ parentChain, height: { $lt: forkHeight } }, { batchSize: 100 });
```
nitsujlangston: will want to get localtip for the child chain add add as part of the parentChain block Cursor Query - that way it can be restarted in middle of process

packages/bitcore-node/src/services/p2p.ts
```diff
@@ -279,10 +285,64 @@ export class P2pService {
...
+    while (blockCursor.hasNext()) {
+      const block = await blockCursor.next();
+      const newBlock = Object.assign({}, block, { chain: newChain });
+      const mongoOperations = new Array();
+      mongoOperations.push(BlockModel.collection.insert(newBlock));
```
nitsujlangston: can do inserts on all these as long as block processed doesnt get written until the 3 steps are done for each block

packages/bitcore-node/src/services/p2p.ts
nitsujlangston: sort by height ascending

packages/bitcore-node/src/services/p2p.ts
```diff
@@ -279,10 +285,64 @@ export class P2pService {
...
+
+      const blockTransactions = await TransactionModel.collection
+        .find({
+          chain: parentChain,
+          blockHeight: { $lt: forkHeight }
```
nitsujlangston: since now in the scope of a block, can just use the blockHeight from the current block

Issue #1658, opened 2018-08-03
V8 DeprecationWarning UrlParser


osagga:
When running a clean install of bitcore v8, the log shows the following warning
```
(node:23459) DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect.
```

I'm using the required node version
```
$ node -v
v8.9.4
```

micahriggan:
This has to do with the mongo version change, should be an easy fix

osagga:
So I tried adding `useNewUrlParser` to the connect call as follows:
```
return MongoClient.connect(
          connectUrl,
          {
            keepAlive: 1,
            poolSize: config.maxPoolSize,
            useNewUrlParser: true
            /*
             *nativeParser: true
             */
          }
        );
```

But it throws this runtime error
```
src/services/storage.ts(36,13): error TS2345: Argument of type '{ keepAlive: number; poolSize: number; useNewUrlParser: boolean; }' is not assignable to parameter of type 'MongoClientOptions | undefined'.
  Object literal may only specify known properties, and 'useNewUrlParser' does not exist in type 'MongoClientOptions | undefined'.
```

I'm not really sure if there's a problem with the mongodb package or something else?

micahriggan:
That's a typings error, try updating package.json to have @types/mongodb: 3.1.3, 

May fix it, I'm checking now

micahriggan:
That MR should fix at least the mongo error. Will test the generate blocks issue with this await change and see if it affects it.

osagga:
I'll test it out and keep you updated, thanks!

PR #1659, opened 2018-08-03
MongoError + Awaiting peerBlock

micahriggan:
We were having a urlParser error, and adding blocks on peer emit wasn't awaiting.


micahriggan:
https://github.com/bitpay/bitcore/issues/1658

PR #1661, opened 2018-08-07
Feature/suggest address tx fee wallet export

micahriggan:
Add an endpoint  to detect missing addresses from a wallet
Add fee calculation on sync


packages/bitcore-node/src/models/transaction.ts
```diff
@@ -125,11 +125,43 @@ export class Transaction extends BaseModel<ITransaction> {
...
+      megaOr = megaOr.concat(spentInputQueries);
+    }
+    //console.log('Mega OR Len', megaOr.length);
+    const coinInputs = await CoinModel.collection
+      .aggregate<{ _id: string; total: number }>([
+        { $match: { $or: megaOr } },
+        { $group: { _id: '$spentTxid', total: { $sum: '$value' } } }
+      ])
+      .toArray();
+    //console.log('Aggregate finished');
```
micahriggan: remove

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -125,11 +125,43 @@ export class Transaction extends BaseModel<ITransaction> {
...
+          return {
+            mintTxid: txInput.prevTxId,
+            mintIndex: txInput.outputIndex
+          };
+        }
+        return;
+      });
+      megaOr = megaOr.concat(spentInputQueries);
+    }
+    //console.log('Mega OR Len', megaOr.length);
```
micahriggan: remove

PR #1663, opened 2018-08-07
Suggest Missing Wallet Addresses

micahriggan:
Adding endpoint to stream back addresses that a wallet has spent from, but aren't tracked by bitcore.


packages/bitcore-node/src/providers/chain-state.ts
```diff
@@ -204,23 +250,36 @@ export class InternalStateProvider implements CSP.IChainStateService {
     };
     if (args) {
       if (args.startBlock) {
-        query.blockHeight = { $gte: Number(args.startBlock) };
+        query.mintHeight = { $gte: Number(args.startBlock) };
```
micahriggan: Revert this function

packages/bitcore-node/src/providers/chain-state.ts
```diff
@@ -190,6 +193,49 @@ export class InternalStateProvider implements CSP.IChainStateService {
...
+    const wallet = await WalletModel.collection.findOne({ pubKey });
+    const walletId = wallet!._id;
+    const query = { chain, network, wallets: walletId, spentHeight: { $gte: 0 } };
+    const cursor = CoinModel.collection.find(query);
+    const seen = {};
+    const stringifyWallets = (wallets: Array<ObjectId>) => wallets.map(w => w.toHexString());
+    const allMissingAddresses = new Array<string>();
+    let totalMissingValue = 0;
+    const missingStream = cursor.pipe(
+      through2(
```
micahriggan: @jguthrie7 this is what I was talking about yesterday, new Transform() can do this?

PR #1666, opened 2018-08-08
V8 with raw tx

osagga:
This follows from bitpay/bitcore-node#574 This will add the API endpoint `tx/:txid/raw` that will return the hex string of the transaction (using the RPC command `getrawtransaction`). 

In the previous versions of the node, There used to be a separate endpoint for rawTx. The raw transaction can used as a common way to initialize a Transaction object in various bitcoin libraries, that's why it's really helpful.

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/index.ts
+++ b/packages/bitcore-node/src/providers/chain-state/index.ts
@@ -52,6 +52,10 @@ class ChainStateProxy implements CSP.ChainStateProvider {
     return this.get(params).streamTransaction(params);
   }
 
+  streamTransactionRaw(params: CSP.StreamTransactionParams) {
+    return this.get(params).streamTransactionRaw(params);
+  }
+
   async createWallet(params: CSP.CreateWalletParams) {
     return this.get(params).createWallet(params);
   }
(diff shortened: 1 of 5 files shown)


micahriggan:
We're creating transactions in bitcore-client without using any raw TX endpoints
https://github.com/bitpay/bitcore/blob/v8.0.0/packages/bitcore-client/lib/wallet.js#L155

I would prefer keeping the API separate from RPC as much as possible. If there's something out transaction model doesn't have on it, I'd rather add that than expose the RPC endpoint

osagga:
The main idea I'm trying to state is that it's better to just provide the raw hex in case some application is using that as a universal way to represnet and parse transactions, to avoid different transaction representations across different APIs. 

Another way we can achieve this (without exposing the RPC) is to have the transaction module save the raw hex string of the transaction. 
I currently don't yet understand how the transactions are being communicated from the bitcoin-client to the bitcore-node (through p2p), I assume it has to do with `pool.on('peertx')` and `pool.on('peerblock')` here 
https://github.com/bitpay/bitcore/blob/459c40dcc474bdc5fac7ceb2ca6a91af4b90af4d/packages/bitcore-node/src/services/p2p.ts#L70-L95 

But I think that the bitcoin-client may be prodiving the raw hex string, it's just not being used by the bitcore-node yet. 
Do you mind providing some pointers to where the parsing of the bitcoin-client messages is being done by the bitcore-node? Is it in `bitcore-lib` or `bitcore-p2p`?

PR #1669, opened 2018-08-09
Add network prefix to BCH addresses

### LINKED ISSUE #1668: V8 doesn't show BCH network prefix in addresses
It seems that after this commit a9622e2c786063bb72ffcf353dedbb4ae83e9773, the BCH addresses wouldn't show the network in the address (the bool `stripCash` is set to `true`), for example:
it should show `"bchreg:xxxxxx"`, but now it shows `"xxxxx"`. 

This behavior is inconstant with the address format BCH node (`bitcoin-abc`) returns when asking for a new address. Why?

osagga:
+ This reverts the commit a9622e2c786063bb72ffcf353dedbb4ae83e9773
+ Fixes #1668


osagga:
Not needed anymore according to #1668

PR #1672, opened 2018-08-14
Batch blocks

micahriggan:
This branch breaks processing a block into getting the mongo data back, and then running the operation in a separate function.

This allows use to send a list of mongo operations into the process function, that multiplies the "mintMap" functionality across multiple blocks, expanding the usefulness and reducing spentOps.

Also has the added benefit of reducing updates for setting nextBlockHash on previous blocks.


micahriggan:
Currently the mintOps length in p2p is set to 25k before a save occurs. I think we could have it set to 200 blocks or 200k mint ops, whichever occurs first, and then we'd get better speeds later in the chain.

micahriggan:
A small modification to this could increase the speed even more.

By saving the block hashes that are queued up in the promises, you could skip the last await at the end of the while loop, and send those hashes into getHeaders. This would allow you to continually get the blocks, while only waiting for the previous batch to be finished when the next batch is ready to process.

Issue #1657, opened 2018-08-03
V8 can't handle 'generate' RPC command when running in 'regtest'


osagga:
After a clean install of `bitcore v8`, when running in `regtest`, sending a generate command
```
bitcoin-cli -datadir=full-nodes/bitcoin/regtest/ generate 101
```

causes the bitcore-node to behave in an undefined state. The log of the node is as follows:

```
info: Added block 1a4213ca8450217c3270925a9dc47650451d8c8ffe28ed5b3112fec2ce9843ab chain=BTC, network=regtest
info: Added block 43eab1f45bf71508c9dae0fde4e55b77ac00a81c2ed4bfdbd7c037267c3517ef chain=BTC, network=regtest
info: Added block 07886645493302d718cf3dd1e53e79c7e395cf83bb201aa28b5d011e73390af4 chain=BTC, network=regtest
info: Added block 179f07b3fb8a1a14e9496447be070d592a0ecd7d7b80cfd9a8ccbb6f7401d6d9 chain=BTC, network=regtest
info: Added block 464685059f702da9f3cc8f2756ebcba65216f0bab39abd2628fadb157a94476c chain=BTC, network=regtest
info: Added block 60e06e2535fc0b6ddac994e25f5d6dd075a53e945bc3c5b3432679e0815e0a24 chain=BTC, network=regtest
info: Added block 5d42b08a884d9d70847ac34e93e5f2285416992d28430caa7100d82d3e785f6e chain=BTC, network=regtest
(node:23459) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): reorg
(node:23459) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 5): reorg
(node:23459) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 7): reorg
info: Added block 44cb18980f357758d9e1b74fad8579749758bdf67ee4012302d0d1e2f10ce373 chain=BTC, network=regtest
(node:23459) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 9): reorg
info: Added block 3cec06c47ce039dafb89ae3c1c1563cfd06f2119fd318dd34cc0a710e2ef5310 chain=BTC, network=regtest
info: Added block 5ca802a6f4194397ce2cd4ac582caa5814d2d1d5cc8ee30a3a7f0b637b8fe6f0 chain=BTC, network=regtest
[...]

PR #1673, opened 2018-08-14
Fixing generate blocks on regtest

micahriggan:
This basically fixes a scenario where multiple blocks come in at once, and start sync multiple times. What should happen is while synching, you can’t start sync again (it just returns false)

This way the first block to come out of order will trigger a reorg, which will restart sync. Future blocks that come in under this condition can attempt to start a sync, but won't have any effect since sync is already started.


micahriggan:
closes https://github.com/bitpay/bitcore/issues/1657

PR #1681, opened 2018-08-21
Fix the verify.ts file

osagga:
Given the recent state of this branch, v8.0.0 doesn't start and throw some errors regarding `verify.ts`. This commit should fix that.


micahriggan:
@osagga thanks!

PR #1683, opened 2018-08-24
AuthKey is no longer derived from Wallet PrivKey

micahriggan:
* Changing the derived key to be it's own private key
* Switching all the api routes to use the pubkey generated from the new auth key
* Fixing an issue with register that caused the "chain/network" piece to not be included until after the first run

### DIFF
--- a/packages/bitcore-client/lib/client.js
+++ b/packages/bitcore-client/lib/client.js
@@ -33,8 +33,8 @@ Client.prototype.register = async function (params) {
 };
 
 Client.prototype.getBalance = async function (params) {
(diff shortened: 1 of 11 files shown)


packages/bitcore-client/lib/client.js
```diff
@@ -80,7 +82,7 @@ Client.prototype.getFee = async function (params) {
 };
 
 Client.prototype.importAddresses = async function(params) {
-  const { payload, pubKey } = params;
```
micahriggan: This wasn't supposed to be removed

PR #1686, opened 2018-08-27
bitcore-lib-cash cleanup

micahriggan:
Cleaning some variables out that were useful when enableRegtest was a thing


matiu:
test are failing, not sure why.

Can you please also send the changes to https://github.com/bitpay/bitcore-lib-cash ?

micahriggan:
some npm module is failing to install, let me see if I can get that sorted

PR #1697, opened 2018-09-10
Fix confirmations value computation for Transactions

### LINKED ISSUE #1667: V8 `/tx/` endpoints don't provide the number of confirmations
This follows from bitpay/bitcore-node#575, my argument is as follows:

It is true that the user can calculate the confirmations locally (using currentHeight and txHeight), but I think it would be much faster and more convenient if the `\tx\` API endpoints return the confirmations value.

If the user needs to derive the confirmations, the user will need to make two API calls per transaction (one to get the current chain tip, and another to get the tx.height) to calculate the confirmations, instead of only one if the API endpoint that provides it.

All the node has to do is to locally compute the confirmations when a transaction is requested (or multiple transactions) through the API. 

Having the API providing the confirmations will not increase the storage, it would only slightly impact the response time (not by much compared to the user making two API calls to compute the same value).

osagga:
+ Fixes #1667

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/internal/internal.ts
+++ b/packages/bitcore-node/src/providers/chain-state/internal/internal.ts
@@ -155,9 +155,12 @@ export class InternalStateProvider implements CSP.IChainStateService {
     const tip = await this.getLocalTip(params);
     const tipHeight = tip ? tip.height : 0;
     return Storage.apiStreamingFind(TransactionModel, query, { limit }, stream, t => {
-      const txHeight = t.blockHeight || 0;
+      let confirmations = 0;
+      if (t.blockHeight && t.blockHeight >= 0){
+        confirmations = tipHeight - t.blockHeight + 1;
+      }
       const convertedTx = TransactionModel._apiTransform(t, { object: true }) as Partial<ITransaction>;
-      return JSON.stringify({ ...convertedTx, confirmations: tipHeight - txHeight });
+      return JSON.stringify({ ...convertedTx, confirmations: confirmations });
     });
   }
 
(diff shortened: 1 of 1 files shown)


skddc:
It would seem that this is missing a test?

(I'm not familiar with this repo's policy on merging PRs yet, but according to https://github.com/bitpay/bitcore/blob/master/CONTRIBUTING.md#d2---tests one should write tests for all of their code, and personally I think that's a very valuable rule.)

osagga:
@skddc This PR is not introducing any new code, it's simply fixing an existing code, so I don't think the test rule applies here. The tests should've been added when add the "transaction confirmations" feature. 

Also keep in mind that PRs on this repo go through a CI tests when opened, so this PR at least passed those ones.

skddc:
You just explained why the test would have been even more important to add. If something can break, but the test suite is still green, then it means a test is *missing* for the existing feature. This means two things:

1. As the existing test suite isn't able to catch the bug, the existing tests also do not prove that the code changes in the PR actually fix the bug as intended.
2 As the feature is still not tested sufficiently, it can still break again anytime, without anyone noticing. Also, new PRs that touch related code can break it without the tests telling anyone.

skddc:
No, my entire point is that when proposing a bugfix, then there should usually be a test that both proves it fixes the issue, as well as prevents it breaking in the future. That's just normal procedure, when the policy/goal is 100% test coverage. Tests being green when there's a bug is a normal thing, and it's also normal to fix tests for a broken feature when writing a bugfix.

osagga:
@skddc yeah that's a good point, I'll keep that in mind for future PRs! Thanks!

PR #1713, opened 2018-09-20
Fix env var name for DB port config

galfert:
The wrong variable name prevents to use a different port other than the default.


micahriggan:
Thanks!

galfert:
Not sure why CI failed, but it looks like it also failed for all the other pull requests since yesterday.

micahriggan:
Bitcoin core docker image removed, fixed in another mr. no worries

PR #1714, opened 2018-09-20
streamWalletTransactions error

micahriggan:
The query had start and end date even when the date wasn't valid.


packages/bitcore-node/src/models/transaction.ts
```diff
@@ -317,6 +317,7 @@ export class Transaction extends BaseModel<ITransaction> {
     let originalQuery = params.query;
     const { query, options } = Storage.getFindOptions(this, params.options);
     const finalQuery = Object.assign({}, originalQuery, query);
+    console.log('final', finalQuery, options);
```
micahriggan: remove this line

PR #1739, opened 2018-10-02
Block event triggers sync

micahriggan:


osagga:
This theoretically makes sense, but unfortunately this will increase the processing time of new blocks as we would call sync for every new block. 

Also what if we get a new block while a sync call is almost finishing, this new block event will be dropped since sync is locked, and we won't sync this block/s until a future sync, so the explorer will be off by some blocks in some case until a new block event gets triggered when sync is not locked.

I don't think this should be designed this way (at least not for the long term), we should act on each block event, and not drop data if we don't need to. 

Also the underlying problem in the p2p library should be looked into such that the p2p lib would trigger block events in order, or at least patch block events.

micahriggan:
I think If a new block comes in as sync is finishing, then the call to getHeaders() in the while loop would return that block. This could be tested by generating 1000 blocks, and then generating 1 block, and see if all 1001 blocks were synced.

I'll test that now

skddc:
The tests were actually failing after the last commit.

micahriggan:
Saw that, they pass on the v8 branch after this merge

PR #1741, opened 2018-10-03
Reverting raw, too storage intensive for api calls

micahriggan:


micahriggan:
I think that's a good idea. I think we need to work on a design where these features can be toggled on and off.

I'm going to shelf this for the moment, until we get a good setup for that.

PR #1742, opened 2018-10-16
Streaming find sort was being ignored

micahriggan:
This is causing blocks to not be sorted correctly.


micahriggan:
@bitjson this should fix that issue where the blocks weren't being sorted

PR #1757, opened 2018-11-06
Using random bytes for block hashes

micahriggan:


packages/bitcore-node/test/benchmark/benchmark.ts
```diff
@@ -7,7 +7,11 @@ import { Storage } from '../../src/services/storage';
 import { BlockModel } from '../../src/models/block';
 import { BitcoinBlockType } from '../../src/types/namespaces/Bitcoin/Block';
 import { resetDatabase } from "../helpers/index.js";
+import * as crypto from 'crypto';
 
+function randomBytes() {
```
skddc: I think `randomString()` might be more expressive here. Btw, any reason for the indentation missing?

micahriggan: Thanks, I'll change that.

I may not have auto-formatted, I'll see.

PR #1759, opened 2018-11-08
Script can be null

micahriggan:


osagga:
@micahriggan Why could the Script be null? just curious

PR #1762, opened 2018-11-12
bitcore-node: Socket-IO 

micahriggan:
Adding socket events for 

new blocks
mempool transactions
mempool transactions to specific addresses


matiu:
LGTM. Ill test it with some clients to see if there is some missing info.

PR #1763, opened 2018-11-13
feat(api): cache fee estimates that hit rpc

nitsujlangston:


packages/bitcore-node/src/routes/api/fee.ts
```diff
@@ -1,17 +1,23 @@
 import { Request, Response } from 'express';
 import { ChainStateProvider } from '../../providers/chain-state';
 const router = require('express').Router({ mergeParams: true });
+const feeCache = {};
 
 router.get('/:target', async (req: Request, res: Response) => {
   let { target, chain, network } = req.params;
-  if (network === 'regtest') {
```
nitsujlangston: i removed this because it's only reasonable for BTC and BCH, and it assumes the client wants to default it when the actual response should be the error, is better to pass along as long as this is just a passthrough

PR #1779, opened 2018-11-27
feat(sync): prune mempool transactions and coins

nitsujlangston:


nitsujlangston:
@micahriggan updated with implicit promise

PR #1785, opened 2018-12-07
[Insight-prev] FIX: fixes for search bar. Input control before make requests added

rastajpa:


packages/insight-previous/src/providers/bwc/bwc.ts
```diff
@@ -0,0 +1,46 @@
+import { Injectable } from '@angular/core';
+
+import * as BWC from 'bitcore-wallet-client';
```
micahriggan: Why is BWC needed in insight?

rastajpa: Is because the search bar checks if you entered a valid address before making requests. But, you're right, it's not necessary to use BWC, I have to use bitcore-lib and bitcore-lib-cash instead. Thanks!

PR #1794, opened 2018-12-13
Insight prev

micahriggan:
Merging this in, so we can have work on both going side by side


galfert:
After this has been merged, I'm getting the following warning on `npm install`:

> lerna WARN Package name "@bitpay/insight" used in multiple packages:
          	/[...]/bitcore/packages/insight-previous
	/[...]/bitcore/packages/insight

And then a little later this error:

> lerna ERR! npm run postinstall [@bitpay/insight] Output from stderr:
sh: webdriver-manager: command not found

micahriggan:
Hi! Which folder is that happening in? Root level?

galfert:
Hi. Yes, it's happening when running `npm install` on the bitcore repo's root.

micahriggan:
Thanks for the report! I have pushed a commit which fixes that for me, does that resolve it for you?

galfert:
Yes, thanks. This fixed that error for me as well.

But now I'm getting an error, because it's trying to install node-sass version 4.5.0, which doesn't seem to be compatible with Node 10.

micahriggan:
Okay, I'm on node 8.12 I think. Do you know if there's a version that's compatible?

skddc:
According to https://github.com/sass/node-sass/releases node 10 support was introduced in `4.9.0`. I'd just update to the latest version of it.

micahriggan:
I updated to the latest, switched to node 10, and it appears to handle npm install without errors. 

Let me know if something else comes up.

PR #1799, opened 2018-12-17
[insight-prev] FIX: Corrections to show vin address

rastajpa:


micahriggan:
Were there changes here? Or just cleanup / renaming

rastajpa:
> Were there changes here? Or just cleanup / renaming

On the one hand, change the name for greater clarity and the other hand must match the response of the endpoint. In V8.0.0, the request returns vin.address instead of vin.addr

Before this PR the address inside the square on the image below is not shown
insight

micahriggan:
Perfect! Thanks for explaining :)

PR #1800, opened 2018-12-17
feature(node): improve types, enable new APIs for Insight

bitjson:


packages/bitcore-node/src/routes/api/tx.ts
```diff
@@ -42,16 +42,35 @@ router.get('/:txId',  async (req, res) => {
   try {
     const tx = await ChainStateProvider.getTransaction({ chain, network, txId });
     if (!tx) {
-      return res.status(404).send(`txid ${txId} could not be found`);
+      return res.status(404).send(`The requested txid ${txId} could not be found.`);
     } else {
-      return res.send([tx]);
+      return res.send(tx);
```
bitjson: @matiu – I'm not sure if you're using this endpoint yet, but just wanted to make sure you saw this. @micahriggan and I couldn't see why this would be wrapped in an array, so I unwrapped it. Is there any reason we'd want it to be an array here?

packages/bitcore-node/test/integration/models/block.integration.ts
```diff
@@ -489,7 +498,7 @@ describe('Block Model', function() {
       expect(unspentCoins[0].chain).to.equal('BTC');
       expect(unspentCoins[0].network).to.equal('regtest');
       expect(unspentCoins[0].mintTxid).to.equal('a2262b524615b6d2f409784ceff898fd46bdde6a584269788c41f26ac4b4919g'),
-      expect(unspentCoins[0].mintIndex).to.equal(0);
+        expect(unspentCoins[0].mintIndex).to.equal(0);
```
micahriggan: space

packages/bitcore-node/src/models/coin.ts
```diff
@@ -97,20 +84,80 @@ class Coin extends BaseModel<ICoin> {
...
+        chain: string;
+        network: string;
+        authbase: string;
+        identityOutputs: ICoin[];
+      }>([
+        {
+          $match: {
+            mintTxid: mintTxid.toLowerCase(),
+            mintIndex: 0,
+            ...(typeof chain === 'string' ? { chain: chain.toString() } : {}),
```
micahriggan: typeof check, and then calling toString() seems redundant

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -199,6 +201,27 @@ export class InternalStateProvider implements CSP.IChainStateService {
...
+        CoinModel._apiTransform(output, { object: true })
+      );
+      return {
+        chain: found.chain,
+        network: found.network,
+        authbase: found.authbase,
+        identityOutputs: transformedCoins
+      };
+    } else {
+      return undefined;
```
micahriggan: shouldn't there always be an authHead?

packages/bitcore-node/test/integration/models/block.integration.ts
bitjson: Ah! There's a comma instead of a semicolon on the line above. (I was wondering why Prettier would have added that space 😄)

packages/bitcore-node/src/models/coin.ts
bitjson: Fixing in next commit, thanks 👍

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
bitjson: Yes, but the provided `txId` might not exist.

PR #1811, opened 2018-12-19
Insight-Prev: Fixing some data model changes

micahriggan:


micahriggan:
@rastajpa  Just FYI, we broke some things yesterday, this should fix most of it

PR #1817, opened 2018-12-21
feature(node): adding cache headers

micahriggan:
Adding cache middleware to set headers, defaulting to 1 second for everything, and longer on things
deep in the chain

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/internal/internal.ts
+++ b/packages/bitcore-node/src/providers/chain-state/internal/internal.ts
@@ -1,3 +1,4 @@
+import { TransactionJSON } from "../../../types/Transaction";
 import config from '../../../config';
 import through2 from 'through2';
 
@@ -166,7 +167,6 @@ export class InternalStateProvider implements CSP.IChainStateService {
     if (blockHash !== undefined) {
       query.blockHash = blockHash;
     }
-    // TODO: we should get this from a cache rather than waiting for this database call
     const tip = await this.getLocalTip(params);
     const tipHeight = tip ? tip.height : 0;
     return Storage.apiStreamingFind(TransactionModel, query, args, req, res, t => {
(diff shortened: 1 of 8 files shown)


packages/bitcore-node/src/routes/index.ts
```diff
@@ -50,7 +50,8 @@ function getRouterFromFile(path) {
 }
 
 app.use(cors());
-app.use(LogRequest);
+app.use(LogMiddleware());
+app.use(CacheMiddleware(CacheTimes.Second));
```
micahriggan: Default cache of 1 second for all endpoints.

Endpoints that should be excluded can use the no-cache middleware

CacheMiddleware(CacheTimes.None)

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -194,7 +194,7 @@ export class InternalStateProvider implements CSP.IChainStateService {
       if (found.blockHeight && found.blockHeight >= 0) {
         confirmations = tipHeight - found.blockHeight + 1;
       }
-      const convertedTx = TransactionModel._apiTransform(found, { object: true }) as Partial<ITransaction>;
+      const convertedTx = TransactionModel._apiTransform(found, { object: true }) as TransactionJSON;
```
micahriggan: This is annoying. Need to get conditional types for the return types of _apiTransform

packages/bitcore-node/src/routes/middleware.ts
```diff
@@ -13,41 +13,52 @@ function LogObj(logOut: { [key: string]: string }) {
...
-
-  const LogPhase = (phase: string) => () => {
-    const endTime = new Date();
-    const startTime = req.startTime ? req.startTime : endTime;
-    const totalTime = endTime.getTime() - startTime.getTime();
-    const totalTimeMsg = `${totalTime} ms`.padStart(10, ' ');
-    logOut.phase = phase.padStart(8, ' ');
-    logOut.took = totalTimeMsg.padStart(10, ' ');
-    logOut.status = res.statusCode.toString().padStart(5, ' ');
+export function LogMiddleware() {
```
micahriggan: Middleware function returns the original function now

PR #1820, opened 2018-12-24
fix(node): removing confirmations

micahriggan:
Confirmations should be calculated client side by doing tip.height - item.blockHeight

Confirmations defeat caching. We're removing the concept server side


micahriggan:
@matiu @nitsujlangston breaking changes here.

As discussed, confirmations break caching

nitsujlangston:
Looks good. Resolve merge conflicts and should be ready to merge. Cant deploy until bws is ready

PR #1824, opened 2018-12-26
feat(node): service dependency injection

micahriggan:


packages/bitcore-node/src/server.ts
```diff
@@ -1,43 +1,45 @@
...
-  console.error('Unhandled Rejection at:', error.stack || error)
+process.on('unhandledRejection', error => {
+  console.error('Unhandled Rejection at:', error.stack || error);
 });
 
 const startServices = async () => {
-  await Storage.start({});
   await Worker.start();
-  P2pService.startConfiguredChains();
+  await P2P.start();
```
nitsujlangston: can't await this currently, it would block until connected

PR #1825, opened 2018-12-27
feat(api): break balance response into confirmed and unconfirmed 

nitsujlangston:
BREAKING CHANGE: balance response has changed from `{balance: number}` to `{confirmed: number, unconfirmed: number, balance: number}`

@micahriggan take a look at the eth and erc20 providers, they still need some work
@matiu this is a breaking change for the balance endpoints, BWS will need to be ready for this change

### DIFF
--- a/packages/bitcore-node/src/models/coin.ts
+++ b/packages/bitcore-node/src/models/coin.ts
@@ -64,7 +64,7 @@ class Coin extends BaseModel<ICoin> {
     );
   }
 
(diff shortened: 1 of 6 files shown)


nitsujlangston:
@matiu in this case, I could also add balance back to the response object, as a combination of unconfirmed and confirmed, which would make it backward compatible.

matiu:
ACK

PR #1836, opened 2018-12-28
Native cashaddr in BWS

matiu:
use cashAddr internally for BCH wallets
create new endpoints for updated api (createTx / getPendingTxs / getHistory)
update BWC tests


matiu:
ready for review.

PR #1842, opened 2018-12-31
Bug/sweep paper wallet v8

matiu:


packages/bitcore-wallet-service/lib/blockchainexplorers/v8.js
```diff
@@ -186,6 +186,31 @@ V8.prototype.getConnectionInfo = function() {
   return 'V8 (' + this.coin + '/' + this.v8network + ') @ ' + this.host;
 };
 
+V8.prototype._transformUtxos = function(unsent) {
+  let ret = _.map(unspent, function(x) {
```
micahriggan: Why lodash map instead of the array prototype map?

PR #1850, opened 2019-01-02
Fixing undefined config properties and changing to isDisabled

micahriggan:


packages/bitcore-node/src/routes/middleware.ts
```diff
@@ -71,7 +71,8 @@ export function RateLimiter(method: string, perSecond: number, perMinute: number
   return async (req: express.Request, res: express.Response, next: express.NextFunction) => {
     try {
       const identifier = req.header('CF-Connecting-IP') || req.socket.remoteAddress || '';
-      if (Config.for('api').rateLimiter.whitelist.includes(identifier)) {
+      const rateLimiter = Config.for('api').rateLimiter;
```
nitsujlangston:
```
const whitelist = Config.for('api').rateLimiter && Config.for('api').rateLimiter.whitelist
```

PR #1851, opened 2019-01-02
Don't use .confirmation at v8's UTXOs

matiu:
- rm cache on deregistration + rm old code
- Don't use .confirmation at v8's UTXOs


micahriggan:
Looks good to me

PR #1852, opened 2019-01-03
feat(api): Adds check wallet endpoint

nitsujlangston:
Provides last imported address and hash of all wallet addresses

example response:
```
{ lastAddress: 'mxHr228NH6mLVGz5PMXBrqDX8mvVXGdQ7C',
  hash: '5e40213ef562528958cdfda88859b8020ccbcdeb65b327f4d122eeaae606a981' }
```


matiu:
great! we need to update the readme with the new endpoint.

PR #1854, opened 2019-01-03
fix walletCheck

matiu:
as per title.


packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -266,8 +266,10 @@ export class InternalStateProvider implements CSP.IChainStateService {
       const hash = crypto.createHash('sha256');
       let lastAddress;
       addressStream.on('data', (walletAddress: IWalletAddress) => {
-        lastAddress = walletAddress.address;
-        hash.update(walletAddress.address);
+        if (walletAddress.address) {
```
micahriggan: we should probably prevent adding a walletAddress without an address

Issue #1866, opened 2019-01-06
v8: On second+ runs, repeatedly showing syncing same 2000 blocks


jamespharvey20:
After completing a bitcoin core 0.16.3 initial sync, cloning git d5ce6fa, running `npm install` and `npm start`, bitcore synced up to around block 185581 (I didn't record the exact block number here.)  I needed to reboot the machine, so pressed `CTRL+C`, to what was running from `npm start`, hoping that would lead to a clean shutdown.

After rebooting, I ran `npm start` (from within `bitcore.git/packages/bitcore-node`.)  It started off with `height=185581`.  I checked 1.5 hours later, and saw it was at `186826`.  Knowing that it shouldn't be going that slow, I scrolled up and saw it had kept looping through the same blocks.

Note this second start didn't show the error in #1865.

Note the first start seemed to correctly only show an increasing height number.

In the terminal scrollback over the 1.5 hour period, I see 17,000 lines like this:

```
info: Connected to peer 127.0.0.1 chain=BTC, network=mainnet                                                                                                                    
info: This worker is now the syncing node for BTC mainnet                                                                                                                       
info: Syncing 2000 blocks for BTC mainnet
info: Sync  chain=BTC, network=mainnet, height=185581
info: Sync  chain=BTC, network=mainnet, height=185582
info: Sync  chain=BTC, network=mainnet, height=185583
...
info: Sync  chain=BTC, network=mainnet, height=187578
info: Sync  chain=BTC, network=mainnet, height=187579
info: Sync  chain=BTC, network=mainnet, height=187580
info: Syncing 2000 blocks for BTC mainnet
info: Sync  chain=BTC, network=mainnet, height=185581
info: Sync  chain=BTC, network=mainnet, height=185582
info: Sync  chain=BTC, network=mainnet, height=185583
...
(this repeats around 10 times)
...
info: Sync  chain=BTC, network=mainnet, height=186824
info: Sync  chain=BTC, network=mainnet, height=186825
info: Sync  chain=BTC, network=mainnet, height=186826
^C
```

So, I pulled current git (5 new commits than what I started with) to b42c0adf, ran `npm install` (in `bitcore.git`), and ran `npm start` (in `bitcore.git/packages/bitcore-node`) for my third time, hoping it was a fluke:

```
info: Connected to peer 127.0.0.1 chain=BTC, network=mainnet
info: This worker is now the syncing node for BTC mainnet
info: Syncing 2000 blocks for BTC mainnet
[...]

Issue #1865, opened 2019-01-05
v8: Unhandled rejection at: TypeError: Cannot read property 'syncingNode:BTC:mainnet' of null


jamespharvey20:
No idea if this matters, but in the spirit of beta testing, after completing a bitcoin core `0.16.3` initial sync, cloning git d5ce6fa0, running `npm install` and `npm start`, I got the following error.  It's continued on, and up to syncing a height of about 50k so far.

```
...
info: Starting Socket Service
info: Starting P2P Manager
info: Connected to peer 127.0.0.1 chain=BTC, network=mainnet
Unhandled Rejection at: TypeError: Cannot read property 'syncingNode:BTC:mainnet' of null                                                                               
    at StateModel.getSyncingNode (/home/bitcore/bitcore.git/packages/bitcore-node/build/src/models/state.js:20:27)                                                      
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
info: This worker is now the syncing node for BTC mainnet
info: Syncing 2000 blocks for BTC mainnet
info: Sync  chain=BTC, network=mainnet, height=1
info: Sync  chain=BTC, network=mainnet, height=5
...
```

micahriggan:
Thanks for the report!

micahriggan:
@nitsujlangston https://github.com/bitpay/bitcore/blob/ea0bc2cc36a86f1082e59e94ba83f584dde56617/packages/bitcore-node/src/models/state.ts#L23

Shouldn't this have the mongo option 
```
returnNewDocument: true
```

micahriggan:
Seems Node.js native driver uses 
```
23       { returnOriginal: false }
```

PR #1867, opened 2019-01-07
fix(node): config should use merge, findOneAndUpdate should return new

micahriggan:
lodash merge is more in mind with the goal of config's setup. There is an issue with using the state
model's syncing node initially, likely do the original data being returned, which would result in an
error


micahriggan:
Closes #1865

micahriggan:
Closes #1866

Issue #1871, opened 2019-01-08
Missing chain data/APIs for Insight


bitjson:
I'll just update this issue as I come across info we still need:

## Need immediately
- [ ] all unlocking scripts (we're not saving any currently, and we need them all for input views and for re-serializing transactions from the DB)
- [ ] Coinbase information on blocks
  - [ ] the transaction hash of the coinbase transaction
  - [ ] the coinbase unlocking script (which contains signaling information and miner identity claims like which pool mined the block) – this need to be full-text searchable so we can query blocks by signaling information, miner identity claim, etc.
- [ ] serialized, raw block headers - `[...]/block/[hash]/raw`
- [ ] serialized, raw transactions - `[...]/tx/[hash]/raw`

## Future planned features
- [ ] disassembled bitcoin script as strings (for full-text search)

PR #1881, opened 2019-01-10
node: add raw transaction API, raw block header API, and other data for Insight

bitjson:
ISSUES CLOSED: #1871


micahriggan:
some tests will need to be changed for this one

osagga:
@bitjson This shouldn't close #1871 as it doesn't address the missing APIs for raw block/tx as shown in the issue.

bitjson:
Thanks for the reminder @osagga – I'll try to finish those in this PR over the next few days 👍

packages/bitcore-node/src/config.ts
```diff
@@ -57,6 +57,7 @@ const Config = function(): ConfigType {
     dbHost: process.env.DB_HOST || '127.0.0.1',
     dbName: process.env.DB_NAME || 'bitcore',
     dbPort: process.env.DB_PORT || '27017',
+    asmLengthLimit: 1_000_000,
```
bitjson: This can definitely be set much lower without missing the primary use cases. With a lower limit, we'll have to be more thoughtful on the application side to gather info which might otherwise be simple mongo queries, but all OP_RETURN protocols can be fully served with ~80 bytes on BTC, and 223 bytes on BCH (`MAX_OP_RETURN_RELAY`). I'll update this PR to set it to 223 for now.
 
(I'm continuing to sync a node with a 1MB limit right now anyway, so I'll review after it's done.)

bitjson:
@micahriggan this is ready for review 🚀

micahriggan:
@matiu @nitsujlangston  @rastajpa 

BWS and Insight changes with refactor of mintTxid to mintTxId, and spentTxId

Thanks for fixing tests, will need to play with this.

Mind posting the final DB size after your sync is finished?


@nitsujlangston this MR brings up the question of how we do code changes that would require a re-sync, or that add data to already sync'd blocks.

matiu:
ACK with BWS / insight-previous changes.

I strongly suggest that, in the future, changes in variable names, like spendTx*i*d, go in individual PRs and no merged in big changes, like this one.

bitjson:
@micahriggan: sync is still going (I'm syncing all 4 chains at the same time.) I'll post stats as soon as it's done.

@matiu: ya, sorry about that. It's something we wanted to do for consistency, but it requires a full re-sync, so logistically it made sense here. But it's still gross. 😅 

On total sizes: I'm expecting the disassembled scripts to take an additional ~40GB by default for each main chain. Rationale: I think it's better to enable this by default and let users consciously disable it if they need to save space. If you don't have it, and you need it for your use case, you need to resync everything, which might require a week to get back up-and-running. If you do have it on, and just want to save space, you can disable it instantly, and even clean up the disk space with a query that will run in minutes. (See last section.)

The other big change is switching from storing the scripts in MongoDB's BSON "binData" format to storing them in hexadecimal (which is how we store everything else currently). I'm not sure how that will affect the on-disk size yet. At this rate, it looks like each mainnet chain will take a total of around 400 GB on disk.

Other changes affecting storage – we're storing all the block coinbase data now too, which adds a bit. Worth mentioning, we're also storing two pieces that likely don't need to be stored (`coinbaseSequenceNumber` and `coinbaseMintTxId`). Both are required to be certain values by consensus on BTC and BCH currently, and aren't likely to change in a fork. (@micahriggan and I talked about it and decided that we should store them anyways since they're not too large and are part of the P2P serialized data. Principle of least surprise.)

The nice thing about having all these things stored as text is that the [`$regex` operator](https://docs.mongodb.com/manual/reference/operator/query/regex/) works on them, and they can easily be added to a full-text search index. So they're much more useful in a search/analysis context like Insight.

--- 

On that note, if we're interested in drastically reducing database size (in exchange for querying abilities) – we could switch data types currently encoded as text into the MongoDB "binary" format (IDs/hashes, addresses, scripts). We'd give up searching within the contents of those fields, but for something like BWS, we're really just doing lookups, not search/analysis.

So a setting like `binaryEncode`, and a generic `encoding` type parameter on each of the database types. The `JSON` response types could actually be exactly the same, we'd only have to disable any endpoints which rely on search. A higher-level setting like `reduceStorage` which sets both `lockingScriptAsmByteLimit` and `unlockingScriptAsmByteLimit` to `0`, disables storing the extraneous coinbase data, and enables `binaryEncode`, would probably together get the DB well below 200GB per mainnet chain.

Though I'm not sure if this is worth our time/effort yet ^. Storage is cheap, and our value proposition is database flexibility/performance. Use cases which really want to save space can often just use the satoshi implementation as a backend.

bitjson:
Update: I rolled back the `mintTxid` -> `mintTxId` and `spentTxid` -> `spentTxId` change here. It makes more sense to do that in a separate PR, and we can actually run a simple Mongo query to do the rename without a re-sync later.

bitjson:
@micahriggan script disassembly disabled by default, this is ready for another review. 👍

osagga:
@bitjson @micahriggan any update on this? is it on hold until it passes the CI?

micahriggan:
LGTM, I'll run this a bit locally and try to have this merged in the next day or two

monsterbitar:
I'm going to make a note that I too would like to use these features.

matiu:
Closing this for now. 

The increase on DB storage requirements and the changes needed to support this feature seems to be for most users.

PR #1882, opened 2019-01-10
Ref/locker

matiu:
Replace Locker with mongo based locks.
Adds walletCheck as a param for wallet-support.


matiu:
ready for review

PR #1890, opened 2019-01-11
Created logger to replace console.log(err) 

justinkook:
- added winston logger in dependencies, 
- utils/parseArgv file for winston logger
- replaced all console log err with logger error

### DIFF
--- a/packages/insight-previous/package.json
+++ b/packages/insight-previous/package.json
@@ -65,7 +65,7 @@
     "prettier": "^1.14.3",
     "rxjs": "5.5.12",
     "sw-toolbox": "3.6.0",
-    
+    "winston": "^3.1.0",
     "zone.js": "0.8.12"
   },
   "devDependencies": {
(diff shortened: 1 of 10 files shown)


packages/insight-previous/src/logger.ts
```diff
@@ -0,0 +1,16 @@
+const { createLogger, format, transports } = require('winston');
```
cmgustavo: I think that we should avoid `require`

packages/insight-previous/package.json
```diff
@@ -65,7 +65,7 @@
     "prettier": "^1.14.3",
     "rxjs": "5.5.12",
     "sw-toolbox": "3.6.0",
-    
+    "winston": "^3.1.0",
```
cmgustavo: No additional packages is necessary here. Please, check Copay how it handles the console.log. https://github.com/bitpay/copay/blob/master/src/providers/logger/logger.ts

justinkook: Thank you for the reference! Removed winston and matched copay logger 👍

PR #1900, opened 2019-01-15
WIP [insight-previous] FIX: Navigation fixes.

rastajpa:
- Fixes for menu navigation
- Avoid duplicate views when clicking or searching for the same query as the previous one.

### DIFF
--- a/packages/insight-previous/src/app/app.component.ts
+++ b/packages/insight-previous/src/app/app.component.ts
@@ -17,7 +17,8 @@ export class InsightApp {
   private network: string;
 
   public rootPage: any;
-  public pages: Array<{ title: string; component: any; icon: any }>;
+  public activePage: any;
+  public pages: Array<{ title: string; component: any; icon: any; }>;
 
   constructor(
     platform: Platform,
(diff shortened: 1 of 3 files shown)


packages/insight-previous/src/app/app.component.ts
```diff
@@ -58,17 +60,14 @@ export class InsightApp {
...
   public openPage(page: any): void {
-    // close the menu when clicking a link from the menu
     this.menu.close();
-    // navigate to the new page if it is not the current page
-    this.nav.push(page.component, {
-      chain: this.apiProvider.networkSettings.value.selectedNetwork.chain,
-      network: this.apiProvider.networkSettings.value.selectedNetwork.network
-    });
+    if (page !== this.activePage) {
+      this.nav.setRoot(page.component);
```
cmgustavo: You should set `chain` and `network` params in the "setRoot" function.

packages/insight-previous/src/components/head-nav/head-nav.ts
```diff
@@ -65,6 +65,8 @@ export class HeadNavComponent {
     this.q = this.q.replace(/\s/g, '');
     const inputDetails = this.searchProvider.isInputValid(this.q);
 
+    if (this.redirProvider.checkLastQuery(this.q)) { return; }
```
cmgustavo: I'm not sure why do we need this here.

PR #1906, opened 2019-01-17
Bug/move txs

matiu:
- Filter out repeated move transactions (from the same wallet) by checking the address's `isChange` flag.
- fix max lock time
- add doc to bwc credentials.js

### DIFF
--- a/packages/bitcore-wallet-service/config.js
+++ b/packages/bitcore-wallet-service/config.js
@@ -44,8 +44,8 @@ var config = {
         url: 'https://api.bitcore.io',
       },
       testnet: {
-        url: 'http://localhost:3000',
-       //url: 'https://api.bitcore.io',
+       // url: 'http://localhost:3000',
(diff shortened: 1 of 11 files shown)


packages/bitcore-wallet-service/config.js
```diff
@@ -33,33 +33,19 @@ var config = {
...
-        //url: 'https://api.bitcore.io',
-        //addressFormat: 'cashaddr',  // copay, cashaddr, or legacy
-        provider: 'v8',
         url: 'https://api.bitcore.io',
       },
       testnet: {
-        provider: 'v8',
-        //url: 'http://localhost:3000',
-       url: 'https://api.bitcore.io',
+        url: 'http://localhost:3000',
```
micahriggan: I imagine that's not intentional

matiu: thanks, fixed.

PR #1946, opened 2019-01-25
[insight-previous] FIX: remove copay code lines

rastajpa:


packages/insight-previous/src/providers/price/price.ts
```diff
@@ -31,8 +31,9 @@ export class PriceProvider {
           this.currency.loading = false;
         },
         err => {
+          const error = err.status === 404 ? `${err.status} (Not Found)` : err;
```
cmgustavo: We should use a provider to handle this type of errors

rastajpa: Provider added. Thanks!

PR #1977, opened 2019-02-05
[insight-previous] FEAT: fiat rates

rastajpa:
- Rate endpoint for USD added
- Fixes for currency selector
- Prettier and tslint fixes


cmgustavo:
Needs rebase

PR #1987, opened 2019-02-12
Bug/fee estimate rounding

nitsujlangston:
* fix rounding issue on feePerKb
* add feePerByte concept (from bitcore-lib-cash)
* only add change for fee if still economical after increased size due to additional change output


packages/bitcore-lib/test/transaction/transaction.js
```diff
@@ -384,7 +384,70 @@ describe('Transaction', function() {
...
+      });
+      var transaction = new Transaction()
+        .from(inputs)
+        .to(toAddress, 950000)
+        .feePerByte(2)
+        .change(changeAddress)
+        .sign(privateKey);
+      transaction._estimateSize().should.be.within(1000, 1999);
+      transaction.outputs.length.should.equal(2);
+      transaction.outputs[1].satoshis.should.be.within(46002, 48000);
```
matiu: nice test.

matiu:
LGTM.

PR #2016, opened 2019-02-25
[bitcore-node] add a route to check the initial sync state

shalzz:
This PR adds a `sync` route to the `status` route to show the `initialSyncComplete` value of a chain and its network.


packages/bitcore-node/src/routes/status.ts
```diff
@@ -18,6 +19,14 @@ router.get('/performance', function(_, res) {
   res.json(PerformanceTracker);
 });
 
+router.get('/:chain/:network/sync', async function(req, res) {
+  let { chain, network } = req.params;
+  const state = await StateStorage.collection.findOne({});
+  const initialSyncComplete =
+    state && state.initialSyncComplete && state.initialSyncComplete.includes(`${chain}:${network}`);
+    res.json({"initialSyncComplete": initialSyncComplete != null});
```
micahriggan: I think the != null part will return true when the includes returns false.

Seems like something like this would work
```
res.json({initialSyncComplete})
```

shalzz: Hi, yes I took that case into consideration and the current implementation returns false for any non configured chains.
Whereas using `res.json({initialSyncComplete})` returns an empty `{}` for even a configured chain for me.

shalzz: Your concern is valid though and the Typescripts type system handles it. The generated javascript code for this PR is `res.json({ "initialSyncComplete": initialSyncComplete !== null });`

micahriggan:
I tested this branch, and it worked as expected, made that change that I suggested with the res.json, and that appeared to work as well, merging after tests

PR #2031, opened 2019-02-27
Feat/json paypro cli

matiu:
USe Json Payment protocol for bitcore-wallet CLI

Todo:

- [x]  fix payment request signature verification
- [x] add test: "should not broadcast payment if 'verify payment'  failed"


packages/bitcore-wallet-client/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "bitcore-wallet-client",
   "description": "Client for bitcore-wallet-service",
   "author": "BitPay Inc",
-  "version": "8.1.0",
```
micahriggan: this is the correct version, we incremented this week.

packages/bitcore-wallet-client/lib/verifier.js
```diff
@@ -180,11 +181,24 @@ Verifier.checkPaypro = function(txp, payproOpts) {
...
+
+  if (amount != payproOpts.amount)
+    return false;
+
+
+  if (txp.coin == 'btc' && toAddress != payproOpts.toAddress)
+    return false;
+
+  // To circunvent cashaddr/legacy address problems...
+  if (txp.coin == 'bch' && (new BCHAddress(toAddress).toString()) != (new BCHAddress(payproOpts.toAddress).toString()))
```
micahriggan: In the future we may want to extract out these two checks into a multi-currency address validation function
I'm imagining something like this
```
CurrencyAddress.isValid(currency, address)
```

packages/bitcore-wallet-client/package.json
matiu: thanks, will update all that

packages/bitcore-wallet-client/lib/verifier.js
matiu: nice. In this case I need `address1.isEqual(address2)`. Do you think is a good idea to add this to our libraries?

matiu: (the problem here is that string comparision  of `.toString()` does not always work, since older versions of bitcore-lib-cash will render the address add legacy string and newer as cashaddr.

matiu:
Updated and ready for review.

- Added test to new flows.
- Refactor  `request` classes to use the same request libraries in browser and nodejs

PR #2044, opened 2019-03-04
Automatically remove indexes if their name is not in an array of desi…

jguthrie7:
The motivation of this is if you have been running bitcore from the past, there are indexes that have changed.

This will remove indexes that are not explicitly defined in the `desiredIndexes` array.
I'm not sure myself if this is the best way to do this (perhaps we can define names from the list when creating the indexes), but something like this is nice to clean up old indexes

### DIFF
--- a/packages/bitcore-node/src/models/transaction.ts
+++ b/packages/bitcore-node/src/models/transaction.ts
@@ -92,31 +92,47 @@ export class TransactionModel extends BaseModel<ITransaction> {
   ];
 
   onConnect() {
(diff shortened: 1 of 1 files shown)


nitsujlangston:
Since the create indexes are going to happen in the background, the remove would remove the existing index before the replacement is created.

micahriggan:
I don't think this belongs in the app logic, and it prevents people adding their own indexes. I'd rather have a script seperate from the startup logic that does this.

PR #2045, opened 2019-03-04
Replace UglifyJS3 (unmaintained, causing a build error) with Terser

DesWurstes:
UglifyJS3 fails because of a reason I couldn't find:

```
[15:39:49] '<anonymous>' errored after 338 ms
[15:39:49] GulpUglifyError: unable to minify JavaScript
Caused by: SyntaxError: Unexpected token name «i», expected punc «;»
File: /home/usr/Desktop/bitcore/packages/bitcore-lib-cash/bitcore-lib.js
Line: 21724
[15:39:49] 'browser' errored after 2.16 s
```

The line 21724 is the line with `for`:

```
module.exports = function base (ALPHABET) {
  if (ALPHABET.length >= 255) throw new TypeError('Alphabet too long')

  const BASE_MAP = new Uint8Array(256)
  BASE_MAP.fill(255)

  for (let i = 0; i < ALPHABET.length; i++) {
    const x = ALPHABET.charAt(i)
    const xc = x.charCodeAt(0)

    if (BASE_MAP[xc] !== 255) throw new TypeError(x + ' is ambiguous')
    BASE_MAP[xc] = i
  }
```

We can see this as an opportunity to replace this unmaintained package with Terser, which supports ES6 (even if we won't use it.)

I can squash it if you want.


micahriggan:
Merging master, so the ci will run.

PR #2052, opened 2019-03-11
Bug/fee cache

matiu:
Better handle of fee rate cache:

1- decrease duration from 30  to 6 minutes
2- ignore bitcore results if response's nBlock does not match query
3- do not cache results with missing responses.


gabrielbazan7:
ACK

PR #2057, opened 2019-03-13
[insight-previous] FEAT: Migration to Angular 5

rastajpa:
- Migration from Angular 4 to Angular 5
- Use HttpClient instead of Http
- Avoid linking to non-existing pages

_(Don't forget to remove **package-lock** and **node_modules**, then run **npm i** and check this PR)_


Gamboster:
The PR has conflicts. Please rebase it

Gamboster:
Looks pretty good so far. Great work! The httpClient update seems to be quite beneficial.

Just a question with this on tslint.json:
`"no-submodule-imports": false`
What is the advantage to use it? Or why it is necessary?

rastajpa:
> Looks pretty good so far. Great work! The httpClient update seems to be quite beneficial.
> 
> Just a question with this on tslint.json:
> `"no-submodule-imports": false`
> What is the advantage to use it? Or why it is necessary?

`" no-submodule-imports ": false 'this line is to avoid tslint errors when it is necessary to import modules from deeper folders

PR #1945, opened 2019-01-25
Add config option to disable API rate limiter

osagga:
- Fixes #1944

PR #2058, opened 2019-03-14
[bitcore-node] make ratelimiting configurable

shalzz:
Allowing turning off the application level rate limiting when in deployments where there is a load balancer with it's own rate limiting capability


osagga:
@shalzz This change has already been added here: #1945

PR #2060, opened 2019-03-14
Small issue with opts not getting propogate to multisig hash

patwhite:
Fixed a very small issue that caused tx.from(utxo, pks, threshold, opts) to fail for a no-sort multi-sig transaction.


patwhite:
No sure what the failure in the tests is there, seems related to mongo? But, this is a very small change, just propagating opts that weren't propagating before.

micahriggan:
If you pull master into this branch it should fix the test issue. That's related to the mongodb types being updated

patwhite:
I'll just close and reopen, no there are conflicts

PR #2061, opened 2019-03-14
Fixed small issue of opts not propagating for multisig noSorting

patwhite:
Fixed a very small issue with noSorting opts param not propagating to a buildMultisigOut call.


patwhite:
Alright @micahriggan, passing now

micahriggan:
Thanks!

patwhite:
Is there a prerelease version somewhere I can reference to use this now? Having trouble building it locally to use. Thanks man!

micahriggan:
It hasn't been npm published, so you'd have to reference it relatively, or by using the github url in the package.json ( I think )

patwhite:
ya, can't reference this directly from npm since this switched to a mono repo, so you end up having to submodule then reference locally. But, all good, i'll figure it out. Thanks man!

PR #2062, opened 2019-03-14
Adding crypto-wallet-core

micahriggan:
a multi-currency tool for deriving addresses, keys, and creating and signing raw transactions

### DIFF
--- a/packages/crypto-wallet-core/src/transactions/eth/index.ts
+++ b/packages/crypto-wallet-core/src/transactions/eth/index.ts
@@ -1,0 +1,40 @@
+import Web3 from 'web3';
+import EthereumTx from 'ethereumjs-tx';
+import { Key } from "../../derivation";
+
+
+const web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545'));
+export class ETHTxProvider {
+  lib = require('bitcore-lib');
+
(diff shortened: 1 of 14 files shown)


packages/crypto-wallet-core/src/transactions/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  create(params) {
+    return this.get(params).create(params);
+  }
+
+  async sign(params): Promise<any> {
+    return this.get(params).sign(params);
+  }
+
+  getSigningAddresses(params) {
```
micahriggan: I don't think this belongs on the top level thing, since it's really bitcoin specific

packages/crypto-wallet-core/src/transactions/eth/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545'));
+export class ETHTxProvider {
+  lib = require('bitcore-lib');
+
+  async create({ recipients, from, fee = 20000 }) {
+    let txCount = await web3.eth.getTransactionCount(from);
+    const { address, amount } = recipients[0];
+
+    // !Important: Amount needs to be passed into utils as a string
```
micahriggan: remove this line

PR #2064, opened 2019-03-14
Fixed deserialization issue with nosort transaction

patwhite:
When they're deserialized from object, noSort multi sig transactions get validated, and redeem script validation will fail because public keys are in different orders. This change allows the transaction constructor to take opts which will propagate to the fromObject command and the multisig hash builder.


patwhite:
All set, removed both those files, thanks for the feedback!

matiu:
LGTM, good catch.

PR #2065, opened 2019-03-15
xversion placeholder to maintain compatibility with BUcash1.5.1

imaginaryusername:
xversion is a new message type in BUCash 1.5.1, this patch prevents bitcore error when used with it.

### DIFF
--- a/packages/bitcore-p2p-cash/lib/messages/builder.js
+++ b/packages/bitcore-p2p-cash/lib/messages/builder.js
@@ -58,7 +58,8 @@ function builder(options) {
       getblocks: 'GetBlocks',
       getheaders: 'GetHeaders',
       mempool: 'MemPool',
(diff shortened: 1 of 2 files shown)


gandrewstone:
is this happening because bitcore does not tolerate any unknown-type messages, or is there some specific problem with xversion?

imaginaryusername:
It did not seem to tolerate any unknown messages, which is why my "dummy" worked, I think - it's modified from `verack.js`. However I'm unsure how to fix the failed test which seems to be for wallet service.

zander:
According to the p2p spec clients should ignore messages they don't understand.

Maybe that is an alternative route that is easier to make working.

matiu:
thanks for the PR. 

CI is broken since last week, we are trying to fix it. The mono repo migration was a real challenge :)

PR #2068, opened 2019-03-18
bws => ts

micahriggan:
# Common Transformations 

1. Prototype => Class
```
- function Wallet () {}
- Wallet.prototype.method = function(arg) {}
+ class Wallet {
+ method(arg) {}
```

2. Require() => Import
```
- const x = require('x');
+ import * as x from 'x'
```


matiu:
d088b10bca6fd7f7b525a08ecf96e39ccf65c0a1 ACK

matiu:
great work!

PR #2069, opened 2019-03-18
[insight-previous] FEAT: error-handler provider

rastajpa:


packages/insight-previous/src/providers/price/price.ts
```diff
@@ -39,8 +37,8 @@ export class PriceProvider {
         },
         err => {
           this.currencyProvider.loading = false;
-          this.logger.error(err);
           this.showErrorToast();
+          throw err;
```
cmgustavo: I don't recommend to show an error ("this.showErrorToast") and then `throw err`. Providers shouldn't send message directly to the user. 
Keep only `throw err` and let the component or page to make the message to the user

packages/insight-previous/src/pages/transaction/transaction.ts
```diff
@@ -71,9 +69,9 @@ export class TransactionPage {
         // Be aware that the tx component is loading data into the tx object
       },
       err => {
-        this.logger.error(err.message);
         this.errorMessage = err.message;
         this.loading = false;
+        throw err;
```
cmgustavo: No sense `throw err` in a view.

PR #2077, opened 2019-03-21
Bug/request case

matiu:


monsterbitar:
Should the `const request = require('request');` in `/bitcore-wallet-client/lib/paypro.js` also reference a local request, rather than an npm dependency?

PR #2092, opened 2019-04-01
[insight-previous] FIX: UI fixes for invalid txs

rastajpa:
Insight

### DIFF
--- a/packages/insight-previous/src/pages/transaction/transaction.html
+++ b/packages/insight-previous/src/pages/transaction/transaction.html
@@ -6,9 +6,9 @@
 
       <loader *ngIf="loading" [type]="'title'"></loader>
 
-      <error [message]="errorMessage" *ngIf="!loading && errorMessage "></error>
+      <error [message]="errorMessage" *ngIf="!loading && errorMessage && confirmations !== -3 "></error>
 
-      <div *ngIf="!loading && !errorMessage">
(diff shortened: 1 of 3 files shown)


packages/insight-previous/src/pages/transaction/transaction.html
```diff
@@ -6,9 +6,9 @@
 
       <loader *ngIf="loading" [type]="'title'"></loader>
 
-      <error [message]="errorMessage" *ngIf="!loading && errorMessage "></error>
+      <error [message]="errorMessage" *ngIf="!loading && errorMessage && confirmations !== -3 "></error>
 
-      <div *ngIf="!loading && !errorMessage">
+      <div *ngIf="(!loading && !errorMessage) || confirmations === -3">
```
matiu: I think is blockHeight == -3, not confirmations.

rastajpa: Yes, you're right, but this function gets confirmations depending on `blockheight`:
```public getConfirmations(blockheight: number): Observable<number> {
    return this.blocksProvider.getCurrentHeight().map(data => {
      return blockheight > 0 ? data.height - blockheight + 1 : blockheight;
    });
  }```

Maybe the `confirmation` param has to change its name, what do you think?

matiu: IC, ok, thanks for the explanation

PR #2097, opened 2019-04-02
fix push notification server start

matiu:


matiu:
email service is still broken, we are working on it.

PR #2100, opened 2019-04-02
BWS Fix: run emailService

Gamboster:


matiu:
needs rebase.

Gamboster:
@matiu Rebase done!

packages/bitcore-wallet-service/src/fiatrateservice/fiatrateservice.ts
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
-const config = require('../config');
+
+var config = require('../config');
```
matiu: why no `const`?

packages/bitcore-wallet-service/src/lib/emailservice.ts
```diff
@@ -93,17 +93,17 @@ export class EmailService {
       });
     };
 
-    this.defaultLanguage = opts.emailOpts.defaultLanguage || 'en';
-    this.defaultUnit = opts.emailOpts.defaultUnit || 'btc';
+    this.defaultLanguage = opts.emailOpts && opts.emailOpts.defaultLanguage ? opts.emailOpts.defaultLanguage : 'en';
```
matiu: maybe add:

`opts.emailOpts = opts.emailOpts || {};`

 before all this checks, to simplify all the following lines?

packages/bitcore-wallet-service/src/fiatrateservice/fiatrateservice.ts
Gamboster: For some reason, if I use `const` I get a typescript error .. something like "You can not declare config 2 times", taking into account the config of fiatrateservice and emailservice
I'm not sure why it compares the declaration of 2 different files

packages/bitcore-wallet-service/src/lib/emailservice.ts
Gamboster: Yes. Done!

PR #2105, opened 2019-04-03
[FEAT] v3/wallets including server messages array

gabrielbazan7:


cmgustavo:
Upgrade BWC is needed also.

PR #2114, opened 2019-04-04
add raw tx in getTxHistory w "includeExtendedInfo:true"

matiu:


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -3887,6 +3888,11 @@ export class WalletService {
         output.message = txpOut ? txpOut.message : null;
       });
       tx.customData = proposal.customData;
+
+      tx.createdOn = proposal.createdOn;
+      if (opts.includeExtendedInfo) {
```
Gamboster: Maybe it's necessary to add `opts = opts || {};` at the beginning of the function?

matiu: thanks. updated.

PR #2124, opened 2019-04-11
cwc compiles

micahriggan:
* cwc didn't have the proper compile scripts in the package.json
* added some typings on the eth-tx creator
* package-lock files for insight and bitcore-build


micahriggan:
@matiu ready for review

PR #2139, opened 2019-04-18
Can now start p2p, api, and full services

micahriggan:
```
npm run api
npm run api:cluster
npm run pruning
npm run p2p
```

Should be able to start services with more granularity

Pruning service should exit if --EXIT true is given as a flag.


packages/bitcore-node/src/services/pruning.ts
```diff
@@ -0,0 +1,70 @@
...
+  coinModel: CoinModel;
+  stopping = false;
+
+  constructor({ transactionModel = TransactionStorage, coinModel = CoinStorage } = {}) {
+    this.transactionModel = transactionModel;
+    this.coinModel = coinModel;
+  }
+
+  async start() {
+    this.detectAndClear().then(() => process.emit('SIGINT'));
```
micahriggan: should this service exit after running? Means you can't bundle it with others

PR #2145, opened 2019-04-22
Pulling logging from insight-api

micahriggan:
Logging should look like this for syncing now.
```
info: 2019-04-22 12:19:51.906 EDT | Syncing... | Chain: BTC | Network: mainnet |   70.52 blocks/min | Height:  346152
```


packages/bitcore-node/src/logger.ts
```diff
@@ -11,4 +11,29 @@ const logger = new winston.Logger({
...
+    .getSeconds()
+    .toString()
+    .padStart(2, '0')}.${date
+    .getMilliseconds()
+    .toString()
+    .padEnd(3, '0')} ${timezone}`;
+
+export const timestamp = () => formatTimestamp(new Date());
+
 export default logger;
```
matiu: Can this line be deleted given:
https://github.com/bitpay/bitcore/pull/2145/files#diff-87388ecdc1725fb9c6acc98785273ceeR5

packages/bitcore-node/src/services/p2p.ts
```diff
@@ -325,30 +326,37 @@ export class P2pWorker {
...
+          const now = Date.now();
+          const oneSecond = 1000;
+          if (now - lastLog > oneSecond) {
+            const blocksProcessed = currentHeight - startingHeight;
+            const elapsedMinutes = (now - startingTime) / (60 * oneSecond);
+            logger.info(
+              `${timestamp()} | Syncing... | Chain: ${chain} | Network: ${network} |${(blocksProcessed / elapsedMinutes)
+                .toFixed(2)
+                .padStart(8)} blocks/min | Height: ${currentHeight.toString().padStart(7)}`
+            );
```
matiu: nice!

packages/bitcore-node/src/logger.ts
micahriggan: Yep! Deleted and reverted the import to using the default import

PR #2148, opened 2019-04-22
Mempool pruning worker

micahriggan:
Based off #2139 

The pruning worker should clean up the following scenario

* A transaction exists in the database as a mempool transaction, TX A
* Another transaction is added to the mempool that spends the outputs of TX A, TX B
* Some other transaction is mined that consumes the inputs of TX A
* The prune mempool logic will cleanup TX A, but it will miss TX B

Ideally this logic can be moved into the add transaction logic


micahriggan:
Testing this now

matiu:
Would it be possible to add tests?  In particular, IMO we need tests for for `clearInvalid` / `detectFor` / `scanInvalid`

micahriggan:
100%

micahriggan:
@matiu two integration tests added

Amin2110:
Pay

packages/bitcore-node/test/integration/services/pruning.integration.ts
```diff
@@ -0,0 +1,71 @@
...
+  it('should detect coins that should be invalid but are not', async () => {
+    await insertBadCoins();
+    const shouldBeInvalid = Pruning.detectInvalidCoins('BTC', 'mainnet');
+    let count = 0;
+    for await (const shouldBeInvalidCoins of shouldBeInvalid) {
+      count++;
+      const txids = shouldBeInvalidCoins.map(c => c.mintTxid);
+      expect(txids.length).eq(2);
+      expect(txids[0]).eq('spentInMempool');
+      expect(txids[1]).eq('spentInMempoolAgain');
```
matiu: nice!

PR #2149, opened 2019-04-23
perf($server): Deletes old tx confirmations subscriptions

matiu:


matiu:
this one fails because https://github.com/bitpay/bitcore/pull/2147 needs to be merged.

PR #2171, opened 2019-05-02
Key Class, for BWC

matiu:
Old `Credentials` have xPriv + wallet metadata (like name, m,n account, etc). One credential is 1 wallet.

With this PR:`Key` that has xPriv + mnemonic. A single `Key` lead to 1 or more `Credentials`.
Each `Credentials` have different accounts, coins, etc.

One `Credentials` is still one wallet. To sign a TX, credentials +  the corresponding `Key` is needed.
The encryption feature is on `Key`.


packages/bitcore-wallet-client/lib/api.js
```diff
@@ -293,126 +272,28 @@ API.prototype.validateKeyDerivation = function(opts, cb) {
...
- * Seed from random with mnemonic
- *
- * @param {Object} opts
- * @param {String} opts.coin - default 'btc'
- * @param {String} opts.network - default 'livenet'
- * @param {String} opts.passphrase
- * @param {Number} opts.language - default 'en'
- * @param {Number} opts.account - default 0
- */
-API.prototype.seedFromRandomWithMnemonic = function(opts) {
```
cmgustavo: Is there a reason why not check this: `$.checkState(false, 'not supported');` instead remove the function?
Copay uses `seedFromRandomWithMnemonic`.

matiu: didnt want to remove the code just yet. will remove it soon.

packages/bitcore-wallet-client/lib/api.js
```diff
@@ -2401,6 +2159,50 @@ API.prototype.getWalletIdsFromOldCopay = function(username, password, blob) {
...
+      c[i] = x[i];
+    }
+  });
+  if (c.externalSource) {
+    throw new Error('External Wallets are no longer supported');
+  }
+  c.coin = c.coin || 'btc';
+  c.addressType = c.addressType || Constants.SCRIPT_TYPES.P2SH;
+  c.account = c.account || 0;
+  c.rootPath = c.getRootPath();
```
Gamboster: Here is missing this: `c.keyId = k.id`

Gamboster: Should it also be uploaded to version 2? (`c.version = 2`) ?

packages/bitcore-wallet-client/lib/api.js
matiu: removed.

packages/bitcore-wallet-client/test/key.test.js
```diff
@@ -0,0 +1,565 @@
...
+      var c = Key.fromExtendedPrivateKey('xprv9s21ZrQH143K3zLpjtB4J4yrRfDTEfbrMa9vLZaTAv5BzASwBmA16mdBmZKpMLssw1AzTnm31HAD2pk2bsnZ9dccxaLD48mRdhtw82XoiBi');
+      var path = c.getBaseAddressDerivationPath({
+        account: 2,
+        coin: 'btc',
+        network: 'testnet',
+        n: 1,
+      });
+      path.should.equal("m/44'/1'/2'");
+    });
+    it.skip('should return path for BIP45', function() {
```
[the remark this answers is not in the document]
matiu: yes. removed. It is not longer allowed to create  *new* credentials using BIP45 (only import old ones).

packages/bitcore-wallet-client/lib/key.js
```diff
@@ -0,0 +1,587 @@
...
+    throw 'Invalid argument';
+  }
+
+  var x = new Key();
+  x.xPrivKey = xPriv;
+  x.mnemonic = null;
+  x.mnemonicHasPassphrase = null;
+
+  x.use48forMultisig = !opts.useLegacyPurpose;
+  x.use145forBCH = !opts.useLegacyCoinType;
```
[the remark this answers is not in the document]
matiu: This is for backward compatibility.

We normally do it that way so when you import "old" objects from the storage, (using `.fromObj` or similar) it works.

But key class is a new class, so there are no "old objects". Ill change that.

packages/bitcore-wallet-client/lib/key.js
```diff
@@ -0,0 +1,587 @@
...
+  if (!_.includes(['livenet', 'testnet'], network)) throw new Error('Invalid network');
+};
+
+/*
+ * This is only used on "create"
+ * no need to include/support
+ * BIP45
+ */
+
+Key.prototype.getBaseAddressDerivationPath = function(opts) {
```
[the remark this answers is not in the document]
matiu: thanks, will change that.

packages/bitcore-wallet-service/tsconfig.json
```diff
@@ -29,6 +29,7 @@
     "sourceMap": true
   },
   "include": [
+    "./scripts/**/*.ts",
```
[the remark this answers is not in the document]
matiu: yes. `v8tool.ts` This is an old fix, no related to this PR (this is for BWS). Sorry about that.

packages/bitcore-wallet-client/lib/key.js
```diff
@@ -0,0 +1,587 @@
...
+      opts = opts.concat(testnet);
+   } else {
+      //  leave only BTC, and no testnet
+      opts = opts.filter((x) => {
+        return x[0] == 'btc';
+      });
+   }
+
+    let clients = [];
+    async.each(opts,
```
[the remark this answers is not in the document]
matiu: Agree, but we are not using async/await / promises anywhere on BWC now. I would prefer to keep old async / callbacks, and then refactor all to modern JS / TS all together.  Does it make sense to you?

packages/bitcore-wallet-client/lib/api.js
```diff
@@ -928,72 +793,6 @@ API.prototype.isComplete = function() {
...
-};
-
-
-/**
- * Checks is password is valid
- * Returns null (keys not encrypted), true or false.
- *
- * @param password
- */
-API.prototype.checkPassword = function(password) {
```
Gamboster: We should implement this function in key.js

packages/bitcore-wallet-client/lib/api.js
```diff
@@ -928,72 +793,6 @@ API.prototype.isComplete = function() {
...
-API.prototype.getPrivKeyExternalSourceName = function() {
-  return this.credentials ? this.credentials.getExternalSourceName() : null;
-};
-
-/**
- * Returns unencrypted extended private key and mnemonics
- *
- * @param password
- */
-API.prototype.getKeys = function(password) {
```
Gamboster: We used to use this function. Will there be something similar in key.js?

matiu: .get() ?

packages/bitcore-wallet-client/lib/api.js
matiu: `new Credentials` returns `.version=2`  should it should work.
adding keyId.

matiu: done.

packages/bitcore-wallet-client/lib/key.js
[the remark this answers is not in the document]
matiu: done

packages/bitcore-wallet-client/lib/key.js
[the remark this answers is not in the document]
matiu: Actually, these are not the same:

`legacyRootPath`  is for getting the path of v1.0 credentials. it will never change, and interpret BIP45, for example.

"getRootPath" is the actual schema, and my change in the future.

packages/bitcore-wallet-client/lib/key.js
```diff
@@ -0,0 +1,586 @@
...
+    });
+
+
+    if (copayerIdAlreadyTested[c.copayerId]) {
+      return  icb();
+    } else {
+     copayerIdAlreadyTested[c.copayerId] = true;
+    }
+
+    let client  = clientOpts.clientFactory ?  clientOpts.clientFactory() :  new Client(clientOpts);
```
Gamboster: `Client` Needs to be imported at the beginning of the file

packages/bitcore-wallet-service/package.json
```diff
@@ -49,6 +49,7 @@
     "socket.io": "^2.2.0",
     "socket.io-client": "^2.2.0",
     "sticky-session": "^0.1.0",
+    "tsc": "^1.20150623.0",
```
micahriggan: seems weird

packages/bitcore-wallet-client/lib/key.js
```diff
@@ -0,0 +1,406 @@
...
+
+Key.prototype.derive = function(password, path) {
+  $.checkArgument(path, 'no path at derive()');
+  var xPrivKey = new Bitcore.HDPrivateKey(this.get(password).xPrivKey,NETWORK);
+  var deriveFn = this.compliantDerivation ? _.bind(xPrivKey.deriveChild, xPrivKey) : _.bind(xPrivKey.deriveNonCompliantChild, xPrivKey);
+  return deriveFn(path);
+};
+
+function _checkCoin(coin) {
+  if (!_.includes(['btc', 'bch'], coin)) throw new Error('Invalid coin');
```
justinkook: What do you think about making a Coin array in Common.Constants instead of hard coding here?

matiu: 💯 thanks.

matiu: done.

packages/bitcore-wallet-client/lib/key.js
```diff
@@ -0,0 +1,587 @@
...
+// Returns { err, clients[] }
+//
+// TODO: name
+Key.import = (opts, clientOpts, cb) => {
+  var self = this;
+
+  $.checkArgument(opts.words || opts.xPrivKey, "provide opts.words or opts.xPrivKey");
+
+  let copayerIdAlreadyTested = {};
+  function checkCredentials(key, opts, icb) {
```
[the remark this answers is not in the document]
matiu: https://github.com/bitpay/bitcore/pull/2171#discussion_r289988829

PR #2177, opened 2019-05-07
Adding a route for block at time

micahriggan:
Adding a method to get a block for a specific time at /block/at-time/:time


packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -52,6 +53,34 @@ router.get('/:blockId', async function(req: Request, res: Response) {
   }
 });
 
+router.get('at-time/:time', async function(req: Request, res: Response) {
```
matiu: Would it be possible to find an endpoint name that reflects that the resulting block's time will be  `$lte` time ?

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -52,6 +53,34 @@ router.get('/:blockId', async function(req: Request, res: Response) {
...
+    const [block] = await BlockStorage.collection
+      .find({
+        $query: {
+          chain,
+          network,
+          timeNormalized: { $lte: new Date(time) }
+        }
+      })
+      .limit(1)
+      .sort({ timeNormalized: -1 })
```
matiu: Is timeNormalized an index already?

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -52,6 +53,34 @@ router.get('/:blockId', async function(req: Request, res: Response) {
...
+      .toArray();
+
+    if (!block) {
+      return res.status(404).send('block not found');
+    }
+    const tip = await ChainStateProvider.getLocalTip({ chain, network });
+    if (block && tip.height - (<IBlock>block).height > 100) {
```
matiu: It would be nice that `100` was defined as a constant in `middleware` or something, in case we change it on the future we dont leave inconsistencies around.

micahriggan:
Intended use is to be able to find a block that is closest to a time. John G was asking for it.

I will make some changes as per your feedback

packages/bitcore-node/src/routes/api/block.ts
micahriggan: Yeah, `src/models/block.ts` 

`{chain: 1, network: 1, timeNormalized: 1}`

packages/bitcore-node/src/routes/api/block.ts
micahriggan: Added a Confirmations enum

packages/bitcore-node/src/routes/api/block.ts
micahriggan: Changed it to `before-time` let me know if you have a better name

PR #2184, opened 2019-05-14
memory improvements for pruneMempool

nitsujlangston:


packages/bitcore-node/src/models/transaction.ts
```diff
@@ -503,25 +502,22 @@ export class TransactionModel extends BaseModel<ITransaction> {
...
       );
-      if (coin) {
-        prunedTxs[coin.spentTxid] = true;
+      if (coin && !prunedTxs.has(coin.spentTxid)) {
+        prunedTxs.add(coin.spentTxid);
+        await Promise.all([
+          this.collection.update(
+            { txid: coin.spentTxid },
+            { $set: { blockHeight: SpentHeightIndicators.conflicting } },
+            { w: 0, j: false, multi: true }
```
nitsujlangston: w 0 j 0 may be a bad idea

PR #2188, opened 2019-05-16
Using volatile for sockets and increasing events collection size

micahriggan:


packages/bitcore-node/src/models/transaction.ts
```diff
@@ -156,19 +156,24 @@ export class TransactionModel extends BaseModel<ITransaction> {
...
-          await mintOps
-            .filter(coinOp => coinOp.updateOne.filter.mintTxid === filter.txid)
-            .map(coinOp => {
-              const address = coinOp.updateOne.update.$set.address;
-              const coin = { ...coinOp.updateOne.update.$set, ...coinOp.updateOne.filter };
-              return () => EventStorage.signalAddressCoin({ address, coin }) as any;
-            })
-            .reduce((promises, promise) => promises.then(promise), Promise.resolve());
+          if (shouldFire(tx)) {
+            await EventStorage.signalTx(tx);
```
nitsujlangston: should we await? would be nice to have more async and ephemeral

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -156,19 +156,24 @@ export class TransactionModel extends BaseModel<ITransaction> {
...
+          }
+        }
+        for (const coinOp of mintOps) {
```
nitsujlangston: should we wait until the block has been synced? otherwise asking right after event may not return it

PR #2189, opened 2019-05-17
Fixing CWC Compile Error + Test

micahriggan:
Adding test framework for crypto-wallet-core with single test


matiu:
Code ACK.

Not sure we want to use `jest` instead of `mocha`. Is `jest` used in any other place in bitcore? `mocha` is used on bws/bwc/bitcore-lib-*/bitcore-mnemonic/etc

matiu:
LGTM

PR #2199, opened 2019-05-22
Removing bitcore-wallet-ui

micahriggan:
We don't need this package


matiu:
Could  you please rebase this? Thanks.

PR #2213, opened 2019-05-30
fix(node): Mempool Events Not Firing While Syncing

micahriggan:
* Overwriting batches for mempool transactions to do setOnInsert
* This allows us to handle mempool transactions while syncing
* Added tests for pruneMempool and events while syncing


packages/bitcore-node/src/decorators/Loggify.ts
```diff
@@ -10,28 +10,26 @@ export function SavePerformance(logPrefix, startTime, endTime) {
...
-      avg: totalTime
+      avg: totalTime,
+      max: totalTime
     };
   } else {
     PerformanceTracker[logPrefix].time += totalTime;
     PerformanceTracker[logPrefix].count++;
     PerformanceTracker[logPrefix].avg = PerformanceTracker[logPrefix].time / PerformanceTracker[logPrefix].count;
+    const currentMax = PerformanceTracker[logPrefix].max;
+    PerformanceTracker[logPrefix].max = totalTime > currentMax ? totalTime : currentMax;
```
micahriggan: Could be Math.max()

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -178,6 +183,27 @@ export class TransactionModel extends BaseModel<ITransaction> {
     }
   }
 
+  toMempoolSafeUpsert(
```
nitsujlangston: the problem with mutating the ops in this separate method is that we now need to maintain both version of the op separately. Having a single op creator that dynamically structures the ops would prevent the two places from ever diverging their construction.

micahriggan: Yeah, not ideal, but this only happens on a single mempool tx, so it should be light.

matiu:
codeACK.

packages/bitcore-node/src/models/transaction.ts
matiu: Given the urgency of this fix, I would suggest we merge it like it is and refactor it later.

matiu:
LGTM. I added an extra test here:
https://github.com/micahriggan/bitcore/pull/20

micahriggan:
Merged the extra test

PR #2218, opened 2019-05-31
Added _estimateSize tests for bitcore-lib fee calculation

justinkook:
- added estimateSize test with 37 inputs and 2 outputs (1 change output)

Max transaction size calculated using:
```
( 37 * 148 ) + ( 2 * 34 + 10 + 37 ) = 5591
```

Min transaction size calculated using:
```
( 37 * 148 ) + ( 2 * 34 + 10 - 37 ) = 5517
```

Results in
```
AssertionError: expected 4053 to be within 5517..5591
```


matiu:
Could you please mv this code to test:
https://github.com/bitpay/bitcore/blob/master/packages/bitcore-wallet-service/src/lib/model/txproposal.ts#L347

?

which is the actual code used for fee estimation on BWS?
thanks.

PR #2219, opened 2019-06-03
Native segwit address sending

andrewtoth:
This PR adds support for sending to native segwit P2WPKH and P2WSH addresses.

### DIFF
--- a/packages/bitcore-lib/lib/address.js
+++ b/packages/bitcore-lib/lib/address.js
@@ -545,6 +545,11 @@ Address.prototype.toObject = Address.prototype.toJSON = function toObject() {
  * @returns {string} Bitcoin address
  */
 Address.prototype.toString = function() {
+  if (this.isPayToWitnessPublicKeyHash() || this.isPayToWitnessScriptHash()) {
+    var prefix = this.network['bech32prefix'];
+    var version = 0; // Only supporting segwit v0 for now
+    return Bech32.encode(prefix, version, this.hashBuffer);
+  }
   return Base58Check.encode(this.toBuffer());
 };
 
--- a/packages/bitcore-lib/lib/address.js
+++ b/packages/bitcore-lib/lib/address.js
@@ -106,7 +106,7 @@ Address.prototype._classifyArguments = function(data, network, type) {
   // transform and validate input data
   if ((data instanceof Buffer || data instanceof Uint8Array) && (data.length === 20 || data.length === 32)) {
     return Address._transformHash(data, network, type);
-  } else if ((data instanceof Buffer || data instanceof Uint8Array) && data.length === 21) {
+  } else if ((data instanceof Buffer || data instanceof Uint8Array) && data.length >= 21) {
     return Address._transformBuffer(data, network, type);
   } else if (data instanceof PublicKey) {
     return Address._transformPublicKey(data);
@@ -173,22 +173,22 @@ Address._transformObject = function(data) {
  * Internal function to discover the network and type based on the first data byte
  *
  * @param {Buffer} buffer - An instance of a hex encoded address Buffer
- * @param {String} prefix - The bech32 human readable prefix
  * @returns {Object} An object with keys: network and type
  * @private
  */
-Address._classifyFromVersion = function(buffer, prefix) {
+Address._classifyFromVersion = function(buffer) {
   var version = {};
 
-  if (prefix) {
-    if (buffer.length === 20) {
+  if (buffer.length > 21) {
+    var info = Bech32.decode(buffer.toString('utf8'));
+    if (info.data.length === 20) {
       version.type = Address.PayToWitnessPublicKeyHash;
-    } else if (buffer.length === 32) {
+    } else if (info.data.length === 32) {
       version.type = Address.PayToWitnessScriptHash;
     } else {
-      throw new TypeError('Invalid buffer length for segwit address')
+      throw new TypeError('Witness data must be either 20 or 32 bytes.')
     }
-    version.network = Networks.get(prefix, 'bech32prefix');
+    version.network = Networks.get(info.prefix, 'bech32prefix');
   } else {
 
     var pubkeyhashNetwork = Networks.get(buffer[0], 'pubkeyhash');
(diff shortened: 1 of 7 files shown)


micahriggan:
Woah, haven't tested this yet, but thanks 😃

andrewtoth:
Not sure if I should include `bitcore-lib.js` and `package-lock.json` in this. It makes the diff huge.

micahriggan:
You don't need to include bitcore-lib.js

andrewtoth:
@micahriggan I've tested this using Copay master frontend, but when getting to the final send flow it talks to a webserver that doesn't have that update, so rejects the address as invalid. However, I've tested that using a bech32 address when sending properly parses the address and constructs and signs a transaction that can be broadcast. If you can give me instructions on how to have a local running Copay frontend talk to a `bitcore-wallet-service` also running locally, then I can test the entire end-to-end flow.

I'd very much like to see this merged and then released with a new version of Copay.

micahriggan:
@andrewtoth awesome! 👏 

Between @justinkook and I, I think we can get you setup.

https://github.com/bitpay/bitcore/tree/master/packages/bitcore-wallet-service#enabling-regtest-mode-for-bws-and-copay

You'll need to run bitcore-node and bitcore-wallet-service

To do that you'll need to modify 
* <BITCORE_DIR>/bitcore.config.json
* <BITCORE_DIR>/packages/bitcore-wallet-service/config.js
* <COPAY_DIR>/app-template/index-template.html

After you've got those setup, you can start them with `npm start` in the directories
or from <BITCORE_DIR>
```
npm run node
npm run bws
```

After you have bitcore-node talking to a bitcoin regtest node, and your BWS is pointing at your local bitcore, you can run copay locally.

After you modify the content-security policy in copay, you'll need to run 
`npm run apply:copay`

Now you should be good to do `npm start` in copay.
While creating a new wallet, click advanced and set the bws url to `http://localhost:3232/bws/api`

Let me know if you need help

andrewtoth:
Cool, thanks for you help. I ended up just using testnet, I didn't know you can set the bws url when creating wallet, so that worked great. I sent a testnet transaction from the wallet to both a P2WSH and P2WPKH address, and it broadcasted fine:
`b6cc50ff19e8cfa86d1ac8610230d1999c4d0a45f45ca67130f118c1e415603c`

andrewtoth:
When checking the sent tx details, the bech32 addresses are just listed as `false`. I added the ability to encode a bech32 address from the output script, but it still shows as false. Not sure if it's an issue with bws or bitcore-lib. Also, any general thoughts on the code in this PR?

micahriggan:
You checked it on your local bitcore-node and that shows as false?

https://github.com/bitpay/bitcore/blob/master/packages/bitcore-node/src/models/transaction.ts#L367

This is the area of the code where bitcore processes the address, perhaps that will help in debugging

As far as I'm concerned the code looks a great. A unit test for converting to a bech32 address would be nice 

@matiu any criteria we want to list for this to go in?

micahriggan:
One common gotcha could be that bitcore-node isn't symlinked to bitcore lib. An npm install at the top level bitcore should fix that by running lerna bootstrap

packages/bitcore-lib/lib/address.js
```diff
@@ -488,6 +545,11 @@ Address.prototype.toObject = Address.prototype.toJSON = function toObject() {
  * @returns {string} Bitcoin address
  */
 Address.prototype.toString = function() {
+  if (this.isPayToWitnessPublicKeyHash() || this.isPayToWitnessScriptHash()) {
+    var prefix = this.network['bech32prefix'];
```
micahriggan: any reason for not doing `this.network.bech32prefix`

andrewtoth: Fixed.

andrewtoth:
The issue was I wasn't testing with `bitcore-node` before and just connecting to `bitcoind` directly. Testing with `bitcore-node` fixed the issue, and you can see the addresses in the sent tx details. Please let me know what else you need from me to get this merged.

maxzzze:
@micahriggan any update on this review? Getting this merged would be really helpful.

matiu:
Hi @andrewtoth 

thanks a lot for this PR!  it is very simple to follow and understand.

Would it be possible to add some more testing?

Please see:
https://github.com/bitpay/bitcore/blob/master/packages/bitcore-lib/test/address.js#L15

we are using the actual bitcoind json tests vector for the tests.

Maybe you can add the tests from:
https://github.com/bitcoin/bitcoin/blob/master/src/test/bech32_tests.cpp
?

thanks a lot.

andrewtoth:
@matiu I added tests for witness address types. The test fixtures you linked are for valid and invalid bech32 encoding, not valid and invalid witness v0 address types. Those are already tested in the package being used to encode/decode the addresses at https://github.com/bitcoinjs/bech32/blob/master/test/fixtures.json. Instead, I copied tests at https://github.com/bitpay/bitcore/blob/master/packages/bitcore-lib/test/address.js and duplicated relevant ones for witness addresses.

One note about this PR, `Address.fromBuffer` doesn't work for witness addresses. I'm not sure what the benefit of storing addresses in buffers is though, instead of using strings. What exactly needs to use address to buffer encoding instead of `toString`?

andrewtoth:
The thing about buffers is that for legacy base58 addresses the entire address can be stored in hex, including the prefix byte which contains all the information needed to recreate the address. For bech32 addresses, the version and hashbuffer can be stored as hex, but the human readable prefix that identifies the network cannot. So to have this work with to/from buffers we need to come up with a scheme to encode/decode the human readable prefix in the buffer along with the version and hash. I'm open to ideas on how to do this if it's a deal breaker for getting this PR in. Otherwise everything should work fine if nothing is using buffer encode/decode for sending to addresses.

andrewtoth:
@matiu @micahriggan I've come up with a solution. When converting native segwit addresses to/from buffers, just encode the entire address string in utf8. That way all the information about its type and network can be preserved.

Please let me know what else I can do to get this merged.

matiu:
Concept and code ACK.

I am OK to get this merged. Great job @andrewtoth !

matiu:
Great job, thanks a lot for the contribution.

andrewtoth:
@matiu Thanks! Any idea of when this will be released with Copay?

matiu:
@andrewtoth we will try to ship it soon. In the next few weeks hopefully.

Thanks again for you contribution.

PR #2223, opened 2019-06-04
fix(bwc): use cashAddress to import from PK (sweep paper wallet)

cmgustavo:


packages/bitcore-wallet-client/lib/api.js
```diff
@@ -624,9 +624,14 @@ API.prototype.getBalanceFromPrivateKey = function(privateKey, coin, cb) {
   var B = Bitcore_[coin];
  
   var privateKey = new B.PrivateKey(privateKey);
-  var address = privateKey.publicKey.toAddress();
+
+  var address = privateKey.publicKey.toAddress().toString();
```
matiu: you can use `toString(true)`  that will work for BTC and remove the prefix for BCH.

packages/bitcore-wallet-client/lib/api.js
```diff
@@ -641,13 +646,17 @@ API.prototype.buildTxFromPrivateKey = function(privateKey, destinationAddress, o
   var coin = opts.coin || 'btc';
   var B = Bitcore_[coin];
   var privateKey = B.PrivateKey(privateKey);
-  var address = privateKey.publicKey.toAddress();
+
+  var address = privateKey.publicKey.toAddress().toString();
```
matiu: see prev comment

PR #2225, opened 2019-06-05
fix(node): fixed issue for coins being tagged with many wallets

micahriggan:
If two wallets interacted, their mints will be tagged with each other's wallets, due to pass by
reference of arrays

This MR fixes that mutation by copying the array of wallets, so they aren't mutated on the mintOps.

### DIFF
--- a/packages/bitcore-node/src/models/transaction.ts
+++ b/packages/bitcore-node/src/models/transaction.ts
@@ -254,7 +254,7 @@ export class TransactionModel extends BaseModel<ITransaction> {
         if (!agg[mintTxid]) {
           agg[mintTxid] = {
             total: value,
-            wallets: [...wallets] || []
+            wallets: wallets ? [...wallets] : []
           };
         } else {
           agg[mintTxid].total += value;
@@ -267,7 +267,7 @@ export class TransactionModel extends BaseModel<ITransaction> {
         if (!agg[coin.spentTxid]) {
           agg[coin.spentTxid] = {
             total: coin.value,
-            wallets: [...coin.wallets] || []
+            wallets: coin.wallets ? [...coin.wallets] : []
           };
         } else {
           agg[coin.spentTxid].total += coin.value;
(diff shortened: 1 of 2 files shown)


packages/bitcore-node/test/integration/wallet-benchmark.integration.ts
```diff
@@ -1,21 +1,132 @@
...
+      await checkWalletUtxos(wallet2, address2);
+      const tx = await rpc.call('createrawtransaction', [
+        utxos.map(utxo => ({ txid: utxo.mintTxid, vout: utxo.mintIndex })),
+        { [address1]: 0.1, [address2]: 0.1 }
+      ]);
+      const fundedTx = await rpc.call('fundrawtransaction', [tx]);
+      const signedTx = await rpc.call('signrawtransaction', [fundedTx.hex]);
+      const broadcastedTx = await rpc.call('sendrawtransaction', [signedTx.hex]);
+
+      await wait(10000);
```
micahriggan: This should wait for the websocket event to fire

packages/bitcore-node/test/integration/wallet-benchmark.integration.ts
```diff
@@ -1,21 +1,148 @@
...
+async function checkWalletReceived(wallet: IWallet, txid: string, address: string) {
+  const broadcastedOutput = await CoinStorage.collection.findOne({
+    chain,
+    network,
+    mintTxid: txid,
+    address: address
+  });
+
+  expect(broadcastedOutput!.address).to.eq(address);
+  expect(broadcastedOutput!.wallets.length).to.eq(1);
```
micahriggan: Ensure coin is only tagged with one wallet.

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -254,7 +254,7 @@ export class TransactionModel extends BaseModel<ITransaction> {
         if (!agg[mintTxid]) {
           agg[mintTxid] = {
             total: value,
-            wallets: wallets || []
+            wallets: [...wallets] || []
```
micahriggan: Before this change, this was by reference, so the following pushes would mutate the original wallets array.

Issue #2245, opened 2019-06-19
[node] `/tx` endpoint returns `-1` for the `fee`


osagga:
I'm using the bitcore node as of this commit 0b5b4210168b622edac2bbbfa039bec9e2aa098f (latest as of now) on `regtest`, my setup is as follows:

```
$node -v
v10.16.0

$ ./bitcoind --version
Bitcoin Core Daemon version v0.18.0
```
When I try to query for a transaction, I get `-1` for the `fee` value, this value used to be correct earlier (at least as early as commit 7a4bc3bbb12a08c3297398a5b0ee1e7d563d5a2e)

Here's an example where fee is `-1` using the latest bitcore node
```json
curl localhost:3000/api/BTC/regtest/tx/d7e7fae59c4dcfe7b14a64c756d66ea83fb475fe5de80f955ff6f909297090f8/ | python -m json.tool
{
    "_id": "5d0a986998032f7e182e8186",
    "blockHash": "650c6e8da4a64fb479b9d1cad057ff78e26013672d076cf1f2be0fd959841e0f",
    "blockHeight": 626,
    "blockTime": "2019-06-19T20:18:29.000Z",
    "blockTimeNormalized": "2019-06-19T20:18:29.004Z",
    "chain": "BTC",
    "coinbase": false,
    "confirmations": 7,
    "fee": -1,
    "inputCount": 161,
    "locktime": 625,
    "network": "regtest",
    "outputCount": 2,
    "size": 10380,
    "txid": "d7e7fae59c4dcfe7b14a64c756d66ea83fb475fe5de80f955ff6f909297090f8",
    "value": 300312104750
}
```

Here's an example when the `fee` value used to work, note that I'm using bitcoind version `0.17.1` and node version `8.94` here
```json
curl -q localhost:3000/api/BTC/regtest/tx/5c0d2a8f73c543cdfac8d93f354c81055003339b08cdb4ccba6c9ea3a665e2b7/ | python -m json.tool
{
    "_id": "5d0aa3c780d1f92609e04ca8",
    "blockHash": "3d32810cb4909dcf7f45838fca00f329d94fd58a4d1d9d4ebcddff005d334fdc",
    "blockHeight": 315,
    "blockTime": "2019-06-19T21:06:53.000Z",
    "blockTimeNormalized": "2019-06-19T21:06:53.001Z",
    "chain": "BTC",
    "coinbase": false,
    "confirmations": 6,
    "fee": 4480,
    "inputCount": 1,
[...]

PR #2247, opened 2019-06-20
Fees cannot be calculated, since spends haven't been processed

micahriggan:
#2245 uncovered an error related to the previous splice change, where fees cannot be calculated until after spends have been processed.

For us to calculate fees, transactions must be processed after mints and spends.
Unfortunately we cannot use splice, since it would clear out the array of mints after partition was called.

We could use splice if we changed addTransaction to query for mints instead of referencing params.mintOps, but that would add an additional db query, which I feel would reduce the gains made from using splice in the first place.

We could also clone the mints, but I feel that would be replicating what slice is doing.

### DIFF
--- a/packages/bitcore-node/src/models/transaction.ts
+++ b/packages/bitcore-node/src/models/transaction.ts
@@ -174,28 +174,6 @@ export class TransactionModel extends BaseModel<ITransaction> {
           }
         })
       );
-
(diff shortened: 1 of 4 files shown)


osagga:
@micahriggan Is it possible this gets merged in today?

PR #2252, opened 2019-06-24
Server Maintenance for bws

bsnowden3:
This PR adds better server maintenance to bws, by sending a 503 error code and message with a link to BitPay's status page any time Bitcoin Wallet Service is down for maintenance.

### DIFF
--- a/packages/bitcore-wallet-service/src/config.ts
+++ b/packages/bitcore-wallet-service/src/config.ts
@@ -48,7 +48,6 @@ module.exports = {
         // url: 'http://localhost:3000',
         url: 'https://api.bitcore.io',
       },
-
     },
   },
   pushNotificationsOpts: {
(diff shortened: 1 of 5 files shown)


packages/bitcore-wallet-service/src/config.ts
```diff
@@ -63,6 +62,9 @@ module.exports = {
     defaultProvider: 'BitPay',
     fetchInterval: 60, // in minutes
   },
+  maintenanceOpts: {
+    bwsIsLive: true,
```
matiu: maybe is better to set  'maintenanceMode: true' / false
so if unset (as it will from upgrading in current installations), 
the server will continue to work

bsnowden3: Thanks, I've now made those changes.

bsnowden3: In the latest commit I had to make some important changes, I've added error handling to Bitcore Wallet Client while Bitcore Wallet Service is under maintenance.  

Secondly I moved the root handler for all requests while bws is under maintenance under CORS policy, as the order in which the handlers were called was causing problems sending error responses to bwc.

matiu:
ACK. please sign the commits.

matiu:
merged here:https://github.com/bitpay/bitcore/pull/2260

PR #2253, opened 2019-06-25
rm "generate" rpc call

matiu:
- Removes `generate` RPC call to make tests run with bitcoin-core v0.19 (and still runs with older versions).
  - adds small test doc


matiu:
merged on https://github.com/bitpay/bitcore/pull/2213

PR #2254, opened 2019-06-27
Feature (BWS) - Get Balance using Bitcore /balance

justinkook:
- added _convertBitcoreBalance function to convert bitcore getBalance response to copay compatible.

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -1689,13 +1689,13 @@ export class WalletService {
    * @param {Object} { unconfirmed, confirmed, balance }
    * @returns {Object} balance - Total amount & locked amount.
    */
-  _convertBitcoreBalance(bitcoreBalance) {
+  _convertBitcoreBalance(bitcoreBalance, locked) {
     const { unconfirmed, confirmed, balance } = bitcoreBalance;
     const convertedBalance = {
       totalAmount: balance,
       totalConfirmedAmount: confirmed,
-      lockedAmount: unconfirmed,
-      lockedConfirmedAmount: unconfirmed,
+      lockedAmount: locked,
+      lockedConfirmedAmount: confirmed - locked,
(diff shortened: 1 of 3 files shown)


matiu:
The `_getUtxosForCurrentWallet`
function will mark "locked" UTXOs that are on pending tx proposals.

Pending tx proposals are: not completely signed proposals (that will not be possible on ETH, given there is no multisig) but also 'unbroadcasted' txp, which there will be on ETH.

How are we going to manage that?

justinkook:
I was thinking using _getUtxosForCurrentWallet as a fallback for UTXO based coins (for backwards compatibility). 

https://github.com/bitpay/bitcore/pull/2254/files#diff-4741582acef9ff9bbe9ba1d9e4826fbaR1742

Use getBalance from bitcore for non-utxo based coins or _getUtxosForCurrentWallet for utxo based coins.

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1684,6 +1684,25 @@ export class WalletService {
     return balance;
   }
 
+  /**
+   * Converts Bitcore Balance Response.
+   * @param {Object} { unconfirmed, confirmed, balance }
```
matiu: please describe the `locked` param,

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1720,38 +1739,52 @@ export class WalletService {
...
-              };
+        if (!Constants.UTXO_COINS[wallet.coin.toUpperCase()]) {
+          bc.getBalance(wallet, (err, balance) => {
+            if (err) {
+              return cb(err);
+            }
+            this.getPendingTxs({}, (err, txps) => {
+              if (err) return cb(err);
+              const lockedSum = _.sumBy(txps, 'amount');
+              const convertedBalance = this._convertBitcoreBalance(balance, lockedSum);
```
matiu: looking good!

PR #2255, opened 2019-06-27
Feature - Crypto-Wallet-Core can validate addresses

justinkook:
- Added validation folder back to crypto-wallet-core
- Deriver.pathFor now takes in livenet or mainnet as network and also account
- Added tests for validation.validateAddress
- Added tests types in tsconfig to use @types/mocha and chai

### DIFF
--- a/packages/crypto-wallet-core/src/validation/btc/index.ts
+++ b/packages/crypto-wallet-core/src/validation/btc/index.ts
@@ -13,13 +13,11 @@ const Bitcore = require('bitcore-lib');
         if (URI.isValid(address)) {
           uri = new URI(address);
           uriAddress = uri.address.toString();
-          if (Address.isValid(uriAddress, network)) return true;
+          return Address.isValid(uriAddress, network);
         }
       }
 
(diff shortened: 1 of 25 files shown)


packages/crypto-wallet-core/src/validation/btc/index.ts
```diff
@@ -0,0 +1,25 @@
...
+      const URI = Bitcore.URI;
+      const Address = Bitcore.Address;
+
+       // Bip21 uri
+      let uri, uriAddress;
+      if (/^bitcoin:/.test(address)) {
+        if (URI.isValid(address)) {
+          uri = new URI(address);
+          uriAddress = uri.address.toString();
+          if (Address.isValid(uriAddress, network)) return true;
```
micahriggan: return Address.isValid

justinkook: done 👍

matiu:
Looking good!

Could you please add some tests? thanks.

PR #2299, opened 2019-07-31
perf(node): improve prunemempool speed

micahriggan:
Rather than do thousands of finds, do one find for all possible mempool spends where the minting
txid is the same, but the spenttxid is not the same. Then update them all to conflicting status

# 5 minute test, height = 1307540

## BEFORE - 132 Blocks
```
    "TransactionModel::pruneMempool": {
        "time": 142947,
        "count": 132,
        "avg": 1082.9318181818182,
        "max": 57309
    },
    "BlockModel::addBlock": {
        "time": 235123,
        "count": 132,
        "avg": 1781.2348484848485,
        "max": 94190
    }
```
## AFTER - 142 Blocks
```
    "TransactionModel::pruneMempool": {
        "time": 7883,
        "count": 143,
        "avg": 55.12587412587413,
        "max": 1309
    },

    "BlockModel::addBlock": {
        "time": 221775,
        "count": 142,
        "avg": 1561.7957746478874,
        "max": 38021
    }
```

# Conclusion
Don't do 200k findOnes, also make sure you're testing with initialSyncComplete -_-


micahriggan:
After some debugging, turns out my benchmarks were missing the pruneMempool logic due to initialSyncComplete.

After setting that, a performance gain can be seen.

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -519,35 +519,40 @@ export class TransactionModel extends BaseModel<ITransaction> {
...
+      c =>
+        spendOps.findIndex(
+          s =>
+            s.updateOne.filter.mintTxid === c.mintTxid &&
+            s.updateOne.filter.mintIndex === c.mintIndex &&
+            s.updateOne.update.$set.spentTxid !== c.spentTxid
+        ) > -1
+    );
+
+    const spentTxids = Array.from(new Set(coins.map(c => c.spentTxid)));
```
nitsujlangston: memory

nitsujlangston: a name indicating these will be conflicting would be more clear, spentTxids is very innocuous

matiu:
Nice work!!


Are the changes in package*.json really necessary? If not, can you revert them?

PR #2300, opened 2019-07-31
Add the ability to start a specific chain and network

micahriggan:
`CHAIN=BCH NETWORK=testnet npm run p2p`


matiu:
Code ACK, will do  some manual tests before merging.

PR #2301, opened 2019-07-31
fix wallet key size

matiu:
in bitcore-client.

### DIFF
--- a/packages/bitcore-lib-cash/lib/privatekey.js
+++ b/packages/bitcore-lib-cash/lib/privatekey.js
@@ -336,8 +336,7 @@ PrivateKey.prototype.toBigNumber = function(){
  * @returns {Buffer} A buffer of the private key
  */
 PrivateKey.prototype.toBuffer = function(){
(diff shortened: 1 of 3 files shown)


micahriggan:
shouldn't the private key itself do logic, so we never return a buffer that is too small

matiu:
Not sure.


 It is not clear to me that a priv key of up to 32 bytes need always to
alloc a 32 byte buffer. It seems more an issue of the receiving function.

micahriggan:
Gotcha, so if we did  new PrivateKey(key.toBuffer())

Would that result in a different key?

micahriggan:
Even more interesting is the comment on toBuffer

https://github.com/bitpay/bitcore/blob/68c4ea440089668dfa10a9695f9c971d34d0b011/packages/bitcore-lib/lib/privatekey.js#L339

matiu:
:)

Sounds good. Lets do it that way then.

matiu:
udpated.

PR #2310, opened 2019-08-05
Validating addresses and putting a length limit on wallet name

micahriggan:


packages/bitcore-node/src/routes/api/wallet.ts
```diff
@@ -155,6 +159,11 @@ router.post('/:pubKey', authenticate, async (req: AuthenticatedRequest, res) =>
   let keepAlive;
   try {
     let addresses = addressLines.map(({ address }) => address);
+    for (const address of addresses) {
+      if (!Validation.validateAddress(chain, network, address)) {
```
fanatid: This add same vulnerability which insight-api have... but in this time `body-parser` have limit 100mb instead 100kb by default.

micahriggan: Which vulnerability? This should prevent large addresses from being created. You could still send 100MB of addresses.

fanatid: But this will not prevent making service very slow. Address validation with 100k symbols length takes ~10s and by default it's possible sent up to 100MB, how much such validation takes?

There should be something like:
```js
const isValid = address.length < 1000 && Validation.validateAddress(chain, network, address)
if (!isValid) {
```

But probably better add length validation to `bitcore-lib` itself.

fanatid:
I did not worked with TypeScript, but it's check types only at compilation, right?
https://github.com/micahriggan/bitcore/blob/fd019fcaf53b79f479041fc676fa7dc99a36a5de/packages/bitcore-node/src/routes/api/wallet.ts#L92-L99
`singleAddress` / `path` can be big strings too. `pubKey` will not work, because MongoDB have limit for indexed columns, but it's still should be validated that it's valid pubkey (and better by verifying some data, like name).

While addresses validated now, it's possible send random valid addresses which not associated with wallet. They should contain index for verifying that they was really belongs to pubkey.

micahriggan:
Added some length checks to bitcore-lib and to the api for those fields you mentioned

PR #2365, opened 2019-09-04
[FEAT] RBF - check for unconfirmed inputs

gabrielbazan7:


micahriggan:
I would prefer we store the sequence number on the coin rather than a flag. So instead of isRBF could you add sequenceNumber, and populate that from the inputs as coins are being minted? 

In the API transform you can populate the isRBF flag if you'd like by doing the same comparison logic

packages/bitcore-node/src/models/coin.ts
```diff
@@ -199,7 +200,8 @@ export class CoinModel extends BaseModel<ICoin> {
       address: valueOrDefault(coin.address, ''),
       script: valueOrDefault(coin.script, Buffer.alloc(0)).toString('hex'),
       value: valueOrDefault(coin.value, -1),
-      confirmations: valueOrDefault(coin.confirmations, -1)
+      confirmations: valueOrDefault(coin.confirmations, -1),
+      sequenceNumber: valueOrDefault(coin.sequenceNumber, undefined)
```
matiu: what's the point of this? (default = undefined will only be use if sequenceNumber=0?)

gabrielbazan7:
> I would prefer we store the sequence number on the coin rather than a flag on transactions. So instead of isRBF could you add sequenceNumber, and populate that from the inputs as coins are being minted?
> 
> In the API transform you can populate the isRBF flag if you'd like by doing the same comparison logic.
> 
> This MR is probably the simplest way to accomplish the goal though. Just seems like we should have the sequence number in the db...
> 
> what do you think?

Done. Thanks for the comment

matiu:
need rebase

Issue #2375, opened 2019-09-11
[node] memory problems due to large BCH blocks


osagga:
This problem is different from #1475 since its not caused by any sorts of memory leaks, but rather `BCH` large blocks.

My `bitcore` node has been recently failing to sync `BCH` on `testnet` starting from the block #1326451 (a 32MB full block), the failure is mostly due to the large number of transactions this blocks holds, more specifically its due to the way the `mintOps` are processed per block in the node.

There are actually two problems that I noticed, both happen in the function `getMintOps` 
https://github.com/bitpay/bitcore/blob/b37e0d6f1c9c256e8a7823be5e932136fc2b9085/packages/bitcore-node/src/models/transaction.ts#L352

The first one is that the array of `mintOps`, defined here: 
https://github.com/bitpay/bitcore/blob/b37e0d6f1c9c256e8a7823be5e932136fc2b9085/packages/bitcore-node/src/models/transaction.ts#L363
gets really large when handling a full 32MB block (since we add a `mintOp` per output per transaction in the block), I've seen the number of items reaching to about 550K ops, and that causes the node to run out of heap and crash with the following error (reported in #1475 ):
```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x8f9d10 node::Abort() [node]
 2: 0x8f9d5c  [node]
 3: 0xaffd0e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xafff44 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xef4152  [node]
 6: 0xef4258 v8::internal::Heap::CheckIneffectiveMarkCompact(unsigned long, double) [node]
 7: 0xf00332 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [node]
 8: 0xf00c64 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 9: 0xf038d1 v8::internal::Heap::AllocateRawWithRetryOrFail(int, v8::internal::AllocationSpace, v8::internal::AllocationAlignment) [node]
10: 0xeccd54 v8::internal::Factory::NewFillerObject(int, bool, v8::internal::AllocationSpace) [node]
[...]

PR #2376, opened 2019-09-11
fix(node): prevent ERR_BUFFER_OUT_OF_BOUNDS when handling large BCH blocks

osagga:
This is one of the fixes addressed in #2375. It works by splitting the mongo `find` query based on `mintOpsAddresses` to prevent an `ERR_BUFFER_OUT_OF_BOUNDS` that's caused by a very large list of `mintOpsAddresses`.

Please check #2375 for more details on the problem and feel free to suggest other alternative solutions

### DIFF
--- a/packages/bitcore-node/src/models/transaction.ts
+++ b/packages/bitcore-node/src/models/transaction.ts
@@ -1,5 +1,5 @@
 import { CoinStorage } from './coin';
-import { WalletAddressStorage } from './walletAddress';
+import { WalletAddressStorage, IWalletAddress } from './walletAddress';
 import { partition } from '../utils/partition';
 import { ObjectID } from 'bson';
 import { TransformOptions } from '../types/TransformOptions';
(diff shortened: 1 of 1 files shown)


packages/bitcore-node/src/models/transaction.ts
```diff
@@ -428,14 +428,24 @@ export class TransactionModel extends BaseModel<ITransaction> {
...
-      for (const mintOp of mintOps) {
-        mintOpsAddresses[mintOp.updateOne.update.$set.address] = true;
-      }
-      let wallets = await WalletAddressStorage.collection
-        .find({ address: { $in: Object.keys(mintOpsAddresses) }, chain, network }, { batchSize: 100 })
-        .project({ wallet: 1, address: 1 })
-        .toArray();
+      let mintOpsAddressesSet = {};
+      for (const mintOp of mintOps) { mintOpsAddressesSet[mintOp.updateOne.update.$set.address] = true; }
+      var mintOpsAddresses = Object.keys(mintOpsAddressesSet);
```
micahriggan: This looks fine, can you use prettier to format?

osagga: Sure, should be pretty now :)

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -428,14 +428,25 @@ export class TransactionModel extends BaseModel<ITransaction> {
...
+      let mintOpsAddressesSet = {};
       for (const mintOp of mintOps) {
-        mintOpsAddresses[mintOp.updateOne.update.$set.address] = true;
+        mintOpsAddressesSet[mintOp.updateOne.update.$set.address] = true;
       }
-      let wallets = await WalletAddressStorage.collection
-        .find({ address: { $in: Object.keys(mintOpsAddresses) }, chain, network }, { batchSize: 100 })
-        .project({ wallet: 1, address: 1 })
-        .toArray();
+      var mintOpsAddresses = Object.keys(mintOpsAddressesSet);
```
micahriggan: can you change this to a `let`

osagga: sure, just changed that

PR #2405, opened 2019-09-19
Save locking script as hex-encoded string

bitjson:
Just wanted to recommend this upstream.

Currently, the `script` field is saved to documents using MongoDB's `BinData()` type. This makes the field mostly-useless for querying since BinData is [sorted by length and size before content](https://docs.mongodb.com/manual/reference/bson-type-comparison-order/#bindata). So it's not possible to index the field for use in e.g. identifying transaction types or OP_RETURN outputs (Cash Accounts, SLP, Memo, etc).

I was expecting this change to increase space on disk, but in testing, it actually reduces database size by about 4%? I have no idea why. (Please let me know if you figure it out.) Tested by syncing mainnet BCH. Before: 365GB, After: 347GB.

(On the name `lockingBytecode` – if anything, this field could be clarified as `lock`/`locking`, and `bytecode` is a little more appropriate than `script`, since we're dealing with the assembled bytecode format, rather than the script/disassembled representation. Also convenient to use a new name to ensure dependent code doesn't silently expect the old format.)


micahriggan:
Thanks for the PR :)

For the size, it may have something to do with leading or trailing zeros, not sure.

matiu:
ACK, thanks! 

Could you please rebase it? any change to do a migration script, so there is no need to resync the complete DB again?

bitjson:
Rebased 👍

I'm not sure about a migration script – if anything, it's probably best to skip the network and run the migration directly in the Mongo shell (rather than trying to do some sort of migration from Bitcore). This worked for me, but could use some error handling and will probably still take a long time:

Upgrade:
```js
db.coins.find().forEach(function(doc){ doc.lockingBytecode = doc.script.hex(); delete doc.script; db.coins.save(doc);  })
```

Downgrade:
```js
db.coins.find().forEach(function(doc){ doc.script = HexData(0, doc.lockingBytecode); delete doc.lockingBytecode; db.coins.save(doc);  })
```

(Also keep in mind that downstream clients might need to be updated if they're depending on the `script` field to be a hex-formatted locking script.)

matiu:
Thanks for the upgrade script!

one more thing. in order to be able to deploy this without requiring the clients to upgrade immediately, that you think we keep `.script` as an output, deriving it from `. lockingBytecode` for sometime?

micahriggan:
We can leave `script` on the apiTransform as the buffer version for some time, and  just have it be derived from lockingByteCode

PR #2406, opened 2019-09-20
Fix(CWC) - exists is not a function

justinkook:
- removes `secp256k1` and replaced with bitcoreLib point
- increase test coverage to all functions.
- remove `ethereumjs-tx` has dependencies on `secp256k1`
- remove `ethereumjs-util` has dependencies on `secp256k1`
- remove `elliptic` unused
- remove `bcrypt` unused
- replaces `ethereumjs` functions with `ethers`


packages/crypto-wallet-core/tests/address.ts
```diff
@@ -17,4 +17,25 @@ describe('Address Derivation', () => {
     const expectedAddress = '0x9dbfE221A6EEa27a0e2f52961B339e95426931F9';
     expect(address).to.equal(expectedAddress);
   });
+
+  it('should be able to generate a valid ETH address, privKey, pubKey', () => {
```
matiu: 😍

packages/crypto-wallet-core/tests/transactions.ts
```diff
@@ -20,7 +20,7 @@ describe('Transaction Creation', () => {
...
+      '0xf9014c808504a817c800809437d7b3bbd88efde6a93cf74d2f5b0385d3e3b08a870dd764300b8000b90124b6b4af05000000000000000000000000000000000000000000000000000dd764300b800000000000000000000000000000000000000000000000000000000004a817c8000000000000000000000000000000000000000000000000000000016ada606a26050bb49a5a8228599e0dd48c1368abd36f4f14d2b74a015b2d168dbcab0773ce399393220df874bb22ca961f351e038acd2ba5cc8c764385c9f23707622cc435000000000000000000000000000000000000000000000000000000000000001c7e247d684a635813267b10a63f7f3ba88b28ca2790c909110b28236cf1b9bba03451e83d5834189f28d4c77802fc76b7c760a42bc8bebf8dd15e6ead146805630000000000000000000000000000000000000000000000000000000000000000';
```
matiu: what this is changed? no only  the beginning `0x`, but also the tail bytes? Is there a missing checksum or something?

matiu: How this will affect upstream libraries?

micahriggan: This could be from us not putting chain code: 1 on the transaction

matiu: Would this difference create problems in consuming libraries?

justinkook: `1c8080` was actually `chainId: 28`. If we pass in `chainId: 1` it results in `018080`

PR #2409, opened 2019-09-20
V8.8.0

matiu:


packages/bitcore-node/package.json
```diff
@@ -78,15 +78,15 @@
...
-    "bitcore-lib-cash": "8.7.3",
-    "bitcore-p2p": "8.7.3",
-    "bitcore-p2p-cash": "8.7.3",
-    "bitcore-wallet-client": "8.7.3",
+    "bitcore-client": "^8.8.0",
+    "bitcore-lib": "^8.8.0",
+    "bitcore-lib-cash": "^8.8.0",
+    "bitcore-p2p": "^8.8.0",
+    "bitcore-p2p-cash": "^8.8.0",
+    "bitcore-wallet-client": "^8.8.0",
```
justinkook: bitcore-wallet-client is not being used by bitcore-node

matiu: this PR only update the version, we can remove it on a following one.

PR #2416, opened 2019-09-24
Fix(node) - updateWallet properly imports addresses

justinkook:
### Bug:

Address was not being set:
```
await WalletAddressStorage.collection.updateMany(
        { chain, network, wallet: wallet._id, address: { $in: addressBatch } },
        { $set: { processed: true } }
      );
```

### Solution:

Uses internal method updateWallet:
`await WalletAddressStorage.updateCoins({ wallet, addresses });`

**AddNewAddressesStream:**
https://github.com/bitpay/bitcore/blob/master/packages/bitcore-node/src/models/walletAddress.ts#L85

### DIFF
--- a/packages/bitcore-node/src/modules/ethereum/api/csp.ts
+++ b/packages/bitcore-node/src/modules/ethereum/api/csp.ts
@@ -15,7 +15,6 @@ import { EthListTransactionsStream } from './transform';
 import { ERC20Abi } from '../abi/erc20';
 import { Transaction } from 'web3/eth/types';
 import { EventLog } from 'web3/types';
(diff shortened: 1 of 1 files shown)


justinkook:
> Does this differ from the Internal state provider?
> If not we could remove this from the eth one

Yes, the internal one does not use EthTransactionStorage so it wont update Eth transactions

justinkook:
> Does this differ from the Internal state provider?
> If not we could remove this from the eth one

The only difference is that it uses EthTransactionStorage updateMany vs TransactionStorage updateMany

PR #2421, opened 2019-09-25
Feature/eth uri

justinkook:
- Can validate ethereum Uris

URIs based on EIP - 681 / EIP - 831

### DIFF
--- a/packages/crypto-wallet-core/src/validation/bch/index.ts
+++ b/packages/crypto-wallet-core/src/validation/bch/index.ts
@@ -3,21 +3,15 @@ const BitcoreCash = require('bitcore-lib-cash');
 
 export class BchValidation implements IValidation {
   validateAddress(network: string, address: string): boolean {
-    // Check if the input is a valid uri or address
-    const URICash = BitcoreCash.URI;
     const AddressCash = BitcoreCash.Address;
-
-    // Bip21 uri
-    let uri, uriAddress;
-    if (/^bitcoincash:|^bchtest:i/.test(address)) {
-      if (URICash.isValid(address)) {
-        uri = new URICash(address);
-        uriAddress = uri.address.toString();
-        return AddressCash.isValid(uriAddress, network);
-      }
-    }
-
     // Regular Address: try Bitcoin Cash
     return AddressCash.isValid(address, network);
+  }
+
+  validateUri(addressUri: string): boolean {
+    // Check if the input is a valid uri or address
+    const URICash = BitcoreCash.URI;
+    // Bip21 uri
+    return URICash.isValid(addressUri);
   }
 }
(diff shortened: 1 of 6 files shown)


packages/crypto-wallet-core/src/validation/eth/index.ts
```diff
@@ -3,6 +3,25 @@ const utils = require('web3-utils');
 
 export class EthValidation implements IValidation {
   validateAddress(_network: string, address: string): boolean {
+    address = this.sanitizeEthereumUri(address);
```
micahriggan: I'm somewhat worried about this change. We rely on this method to tell us whether an address is valid, so we can pass it into methods that rely on an address being provided. 

If we do this, we lose that and instead it's more of a "contains" a valid address

packages/crypto-wallet-core/src/validation/bch/index.ts
```diff
@@ -3,21 +3,25 @@ const BitcoreCash = require('bitcore-lib-cash');
...
+
+  validateUri(network: string, address: string): boolean {
     // Check if the input is a valid uri or address
     const URICash = BitcoreCash.URI;
     const AddressCash = BitcoreCash.Address;
 
     // Bip21 uri
     let uri, uriAddress;
-    if (/^bitcoincash:|^bchtest:i/.test(address)) {
+    if (/^bitcoincash:|^bchtest:/i.test(address)) {
```
matiu: Can we use bitcore-lib-cash for this?

Same thing as here:
https://github.com/bitpay/bitcore/pull/2421/files#diff-57413206f34f99cbc83e311f3b62cde2R11

but with bitcore-lib-cash.

packages/crypto-wallet-core/test/validation.ts
```diff
@@ -17,9 +17,10 @@ describe('Address Validation', () => {
 
   // Uri
   const btcUri = 'bitcoin:1NuKwkDtCymgA1FNLUBaUWLD8s4kdKWvgn';
-  const bchUri = 'bitcoincash:qr8uujscckc56ancdkmqnyyl2rx6pnp24gmdfrf8qd';
+  const bchUri = 'bitcoincash:pp8skudq3x5hzw8ew7vzsw8tn4k8wxsqsv0lt0mf3g';
+  const ethUri = 'ethereum:0x37d7B3bBD88EFdE6a93cF74D2F5b0385D3E3B08A?value=123?gasPrice=123?gas=123?gasLimit=123'
```
matiu: should the URI be:

ethereum:0x37d7B3bBD88EFdE6a93cF74D2F5b0385D3E3B08A?value=123&gasPrice=123&gas=123&gasLimit=123

could you also add a naked one, like:
ethereum:0x37d7B3bBD88EFdE6a93cF74D2F5b0385D3E3B08A

justinkook: added

PR #2422, opened 2019-09-25
Fix(CWC) - sign to non async

justinkook:
Sign was returning a promise instead of a signedRawTx causing ERROR: `invalid proposal signature`


matiu:
why does tx signature need to be async?

matiu:
it seems we can get the signature sync doing this:
```
var signingKey = new ethers.utils.SigningKey(privateKey);
 var signDigest = signingKey.signDigest.bind(signingKey);
                var signature = signDigest(ethers.utils.keccak256(unsignedTx));
var signedTx = ethers.utils.serializeTransaction(transaction, signature)
```

see https://github.com/ethers-io/ethers.js/blob/master/tests/test-wallet.js#L169

PR #2426, opened 2019-09-27
Fix CWC.sign removed utils.toHex from txData

justinkook:
nonce, gasLimit, gasPrice, and value was being pass into serializeTransaction as hex string.

Fixes error:
```
ERROR: 500 - "Returned error: Transaction cost exceeds current gas limit. Limit: 9992977, got: 41900420268743501727197304949355530691197. Try decreasing supplied gas."
```


matiu:
Could you please write a test for it, so we prevent from happening again?

PR #2437, opened 2019-10-02
Add mongo url support

jougene:
In case when we working with mongo replica, need possibility to work with mongo url instead of concatenating different parts of environment variables.
ex. `DB_URL: mongodb://test:test@mongodb-0.mongodb:27017,mongodb-1.mongodb:27017,mongodb-2.mongodb:27017/bitcore`


micahriggan:
Thanks!

PR #2443, opened 2019-10-03
Fix(node) - getAccountNonce nonce

justinkook:
fix getAccountNonce query to count transactions with blockHeight not equal to -1


matiu:
nice catch

PR #2453, opened 2019-10-11
throttles new blocks notifications to 5minutes

matiu:
as per title

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/blockchainmonitor.ts
+++ b/packages/bitcore-wallet-service/src/lib/blockchainmonitor.ts
@@ -11,8 +11,18 @@ const $ = require('preconditions').singleton();
 const Common = require('./common');
 const Constants = Common.Constants;
 const Utils = Common.Utils;
+const Defaults = Common.Defaults;
 let log = require('npmlog');
 log.debug = log.verbose;
+
+
+
+type  throttledNewBlocksFnType = ((that: any, coin: any, network: any, hash: any) => void);
+
+var throttledNewBlocks = _.throttle((that, coin, network, hash) => {
+  that._notifyNewBlock(coin, network, hash);
+  that._handleTxConfirmations(coin, network, hash);
+}, Defaults.NEW_BLOCK_THROTTLE_TIME_MIN * 60 * 1000) as throttledNewBlocksFnType;
 
 export class BlockchainMonitor {
   explorers: any;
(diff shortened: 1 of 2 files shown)


todipratik:
Hey @matiu, @justinkook,

I am using bitcore-wallet-service and was confused with this change. Can you please help me understand why this change was made? In my opinion, this would discard many new block notifications which otherwise should have been processed and stored in mongo table.

matiu:
Hi @todipratik, ETH blocks happen too fast for the Wallet app, and trigger to many requests. Refreshing every 5min is ok for your use case

I see your point if you prefer to make this an adjustable parameter I would be happy to review and merge a PR. thanks!

todipratik:
Hey @matiu, So I am using BTC and the notifications for BTC are getting throttled. I will work to make it adjustable.

But I am just wondering how wallet app is working as it might also be facing similar problem with BTC coin?

todipratik:
Any idea how Copay must be working right now since underlying BWS throttles new blocks? So, how does Copay comes to know if a transaction has been confirmed or not?

justinkook:
Unfortunately copay has no idea if the tx has been confirmed until it gets a new block event.

Every 5 minutes automatically through bitcore-node block events.
OR
Every 30 minutes swiping down to refresh wallets found here:

https://github.com/bitpay/bitcore/blob/master/packages/bitcore-wallet-service/src/lib/common/defaults.ts#L173

We are working to refactor `bwsEvent` function in Copay to be more performant so we can remove the throttling.

todipratik:
Thank you @justinkook.
You linked the code to show how copay gets data from BWS after 30 mins refresh.

Can you also link how copay gets updated from bitcore-node block events?

justinkook:
Full Stack

1. BWS - Listen for newBlock and 5 minute throttle

https://github.com/bitpay/bitcore/blob/master/packages/bitcore-wallet-service/src/lib/blockchainmonitor.ts#L380

2. BWS - server.ts `fetchNotifications`

https://github.com/bitpay/bitcore/blob/master/packages/bitcore-wallet-service/src/lib/server.ts#L3318

3. BWS - v1/notifications route

https://github.com/bitpay/bitcore/blob/master/packages/bitcore-wallet-service/src/lib/expressapp.ts#L808

4. BWC - Get Notifications

https://github.com/bitpay/bitcore/blob/master/packages/bitcore-wallet-client/src/lib/api.ts#L1146

5. Copay/Profile.ts/ - Listen for notification

https://github.com/bitpay/copay/blob/master/src/providers/profile/profile.ts#L367

6. Copay/home.ts - Update wallets, tx history/confirmations, balance, etc on bwsEvent.

https://github.com/bitpay/copay/blob/master/src/pages/home/home.ts#L264

PR #2461, opened 2019-10-13
Improve Sync Speed

### LINKED ISSUE #2459: Regression: 80% of sync speed lost from PR 2403
On 324944c (2019-10-02 13:27:14), I was getting around 12-15 blocks/min.

Upgraded to d12f725 (2019-10-12 01:51:38) and started only getting 2-3 blocks/min.

Bisected to find it starts from #2403, specifically 2833a00.  Although the first block happens quickly, it drops to 3-4 blocks/min.  By the time you get to d12f725, it starts and continues at 2-3.

I understand #2403 is to reduce memory usage, and of course anytime that happens, performance can suffer as a trade-off.

If you're synced and always running, 2-3 blocks/min is obviously fine.  But, if you don't have bitcore running all the time (or are doing an initial sync) this is quite a hit.

With this being such a performance hit, maybe an option could be added to allow keeping an entire block's data in memory all at once, even if it's defaulted to off?

Interestingly, the slowing-down-commit's parent afc96cd I was seeing extreme speeds like 40-50 blocks/min.  I'm wondering if there could be an option to allow this too, although I'm not sure what the risk is for not "await[ing] the bulkWrite".  Don't know if it risks data not being available when required, or data loss during crash, or anything like that.  But, I at least am not seeing any bad effects from the short while it ran at this speed during the bisecting.

I also understand there is a large amount of variance in performance.  I'm only paying attention to huge changes like the 80% I'm consistently seeing.

EDIT: I've never synced testnet.  If the big issue is a/several 32MB block(s) on testnet, perhaps each block's size could be examined when syncing it, and if it's small (<2MB) higher memory usage wouldn't be a problem, but if it's large, the lower memory footprint method could be used on it?  To date, the largest live block size appears to be < 1.3MB.  Of course, someday block sizes will likely grow.

micahriggan:
I'll see if we can address some of the speed issues from #2459 

Closes #2459

Currently trying: 
* Using partition and mapping batches to bulkWrite

# This Branch
```
................................................................................
Benchmark for 80 (1 MB) blocks completed after 44.073 s
1.8151702856624237 MB/s
0.5509125 Seconds/Block

.....
Benchmark for 5 (32 MB) blocks completed after 79.835 s
2.0041335253961297 MB/s
15.966999999999999 Seconds/Block

.
Benchmark for 1 (64 MB) blocks completed after 21.152 s
3.02571860816944 MB/s
21.152 Seconds/Block
```

# Master 
```
................................................................................
Benchmark for 80 (1 MB) blocks completed after 95.397 s
0.8386007945742529 MB/s
1.1924625 Seconds/Block

.....
Benchmark for 5 (32 MB) blocks completed after 159.296 s
1.004419445560466 MB/s
31.859199999999998 Seconds/Block
```

# 324944c
```
................................................................................
Benchmark for 80 (1 MB) blocks completed after 42.783 s
1.869901596428488 MB/s
0.5347875 Seconds/Block

.....
Benchmark for 5 (32 MB) blocks completed after 81.814 s
1.9556555112816878 MB/s
16.3628 Seconds/Block
```

### DIFF
--- a/packages/bitcore-node/src/models/transaction.ts
+++ b/packages/bitcore-node/src/models/transaction.ts
@@ -15,7 +15,7 @@ import { EventStorage } from './events';
 import { Libs } from '../providers/libs';
 import { BaseTransaction, ITransaction } from './baseTransaction';
 import { Readable, Transform } from 'stream';
-import { Collection, BulkWriteOpResultObject } from 'mongodb';
+import { Collection } from 'mongodb';
 import { partition } from '../utils/partition';
 
 export { ITransaction };
@@ -166,23 +166,15 @@ export class MempoolTxEventTransform extends Transform {
 }
 
 export class MongoWriteStream extends Transform {
-  promiseChain = (Promise.resolve() as unknown) as Promise<BulkWriteOpResultObject[]>;
   constructor(private collection: Collection) {
     super({ objectMode: true });
   }
 
   async _transform(data: Array<any>, _, done) {
-    this.promiseChain = this.promiseChain.then(() =>
-      Promise.all(
-        partition([...data], data.length / Config.get().maxPoolSize).map(batch => this.collection.bulkWrite(batch))
-      )
+    await Promise.all(
+      partition([...data], data.length / Config.get().maxPoolSize).map(batch => this.collection.bulkWrite(batch))
     );
     done(null, data);
-  }
-
-  async _flush(done) {
-    await this.promiseChain;
-    done();
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/bitcore-node/src/models/transaction.ts
```diff
@@ -170,7 +171,9 @@ export class MongoWriteStream extends Transform {
   }
 
   async _transform(data: Array<any>, _, done) {
-    await this.collection.bulkWrite(data);
+    await Promise.all(
+      partition([...data], data.length / Config.get().maxPoolSize).map(batch => this.collection.bulkWrite(batch))
```
nitsujlangston: i thought we were on slice, why does the data array need to be cloned?

micahriggan: Oh nice, good catch

PR #2464, opened 2019-10-14
Updated README to show working example

### LINKED ISSUE #2336: Keys is not defined
When I'm trying to make things work with bitcore-wallet-client, it crashes on the following line:
// Generates a new extended private key
var ireneKeys = Keys.create();

where should I get these "Keys"?

d-ondrich:
This commit fixes #2336 

The current example in the README does not work if someone copies the example exactly. I've updated the example so that new users can use this example exactly as shown, removing some confusion for new users.


matiu:
thanks!

PR #2465, opened 2019-10-14
Entry point for copay latest version number

bsnowden3:
Entry point to retrieve the latest version number of copay,  from github /releases/latest.


packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -254,6 +255,21 @@ export class ExpressApp {
       };
     }
 
+    router.get('/latest-version', (req, res) => {
+      let server: WalletService;
+      try {
```
micahriggan:
```
try {
  let server = getServer(req, res)
...
  const htmlString = await rp(latestCopayURL);
  const latestCopayVersion = htmlString.match(/Release v\d.\d.\d/);
  res.json({version: latestCopayVersion})
}
```

packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -857,7 +872,7 @@ export class ExpressApp {
         });
       });
     });
-
+    //
```
micahriggan: Not sure this comment line was needed

cmgustavo:
IMO we should integrate it with the GitHub API instead of parse crude HTML : https://developer.github.com/v3/repos/releases/#get-the-latest-release

packages/bitcore-wallet-service/latest
```diff
@@ -0,0 +1 @@
+<html><body>You are being <a href="https://github.com/bitpay/copay/releases/tag/v6.3.6">redirected</a>.</body></html>
```
matiu: why we need this?

matiu:
ACK, left a small comment regarding packages/bitcore-wallet-service/latest, why we need that file?

packages/bitcore-wallet-service/latest
bsnowden3: mistakenly added will clean

matiu:
Could you add cache to this?  I am afraid this will be called to ofter, and could take a while.

Take a look at:

`checkAndUseGlobalCache (key, duration, cb)`

at `getFeeLevels` for example. it is very simple to use. A cache of ~2hrs would be fine. Thanks!

matiu:
Code ACK. Please add tests.

packages/bitcore-wallet-service/test/expressapp.js
```diff
@@ -50,7 +50,7 @@ describe('ExpressApp', function() {
       });
     });
 
-    describe('Routes', function() {
+    describe.only('Routes', function() {
```
nitsujlangston: only

PR #2470, opened 2019-10-15
Adding ripple module to bitcore-node

micahriggan:
We'll need to implement a CSP and websocket events for XRP

# CSP 
```
    getBalanceForAddress(
      params: GetBalanceForAddressParams
    ): Promise<{ confirmed: number; unconfirmed: number; balance: number }>;
    getBlock(params: GetBlockParams): Promise<IBlock>;
    streamBlocks(params: StreamBlocksParams): any;
    getFee(params: GetEstimateSmartFeeParams): any;
    broadcastTransaction(params: BroadcastTransactionParams): Promise<any>;
    createWallet(params: CreateWalletParams): Promise<IWallet>;
    getWallet(params: GetWalletParams): Promise<IWallet | null>;
    updateWallet(params: UpdateWalletParams): Promise<void>;
    getWalletBalance(
      params: GetWalletBalanceParams
    ): Promise<{ confirmed: number; unconfirmed: number; balance: number }>;
    getWalletBalanceAtTime(
      params: GetWalletBalanceAtTimeParams
    ): Promise<{ confirmed: number; unconfirmed: number; balance: number }>;
    streamAddressUtxos(params: StreamAddressUtxosParams): any;
    streamAddressTransactions(params: StreamAddressUtxosParams): any;
    streamTransactions(params: StreamTransactionsParams): any;
    getAuthhead(params: StreamTransactionParams): Promise<AuthheadJSON | undefined>;
    getDailyTransactions(params: { chain: string; network: string }): Promise<DailyTransactionsJSON>;
    getTransaction(params: StreamTransactionParams): Promise<any | undefined>;
    streamWalletAddresses(params: StreamWalletAddressesParams): any;
    walletCheck(params: WalletCheckParams): any;
    streamWalletTransactions(params: StreamWalletTransactionsParams): any;
    streamWalletUtxos(params: StreamWalletUtxosParams): any;
    streamMissingWalletAddresses(params: StreamWalletMissingAddressesParams);
    getCoinsForTx(params: { chain: string; network: string; txid: string }): Promise<CoinListingJSON>;
    getLocalTip(params): Promise<IBlock | null>;
    getLocatorHashes(params): Promise<any>;
    isValid(params: isValidParams): {isValid: boolean, type:string};
```

# Sockets 
```
/XRP/mainnet/inv => "coin"
/XRP/mainnet/inv => "block"
/XRP/mainnet/inv => "tx"
/XRP/mainnet/{address} => "coin"
```
# Config
```
      "XRP": {
        "mainnet": {
          "chainSource": "rpc",
          "provider": {
            "host": "s1.ripple.com",
            "protocol":"wss"
          }
        }
      }

```


packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  async getBlock(params: CSP.GetBlockParams) {
+    const client = await this.getClient(params.network);
+    const ledger = await client.getLedger({ includeTransactions: true, ledgerHash: params.blockId });
+    return ledger as any;
+  }
+
+  async getFee(params: CSP.GetEstimateSmartFeeParams) {
+    const client = await this.getClient(params.network);
+    const fee = client.getFee();
```
gabrielbazan7: await is missing here

packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  async broadcastTransaction(params: CSP.BroadcastTransactionParams) {
+    const client = await this.getClient(params.network);
+    return client.submit(params.rawTx);
+  }
+
+  async getWalletBalance(params: CSP.GetWalletBalanceParams) {
+    const { chain, network } = params;
+    const addresses = await this.getWalletAddresses(params.wallet._id!);
```
matiu: Do XRP wallets have more that one address?

packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -0,0 +1,140 @@
...
+    return tx;
+  }
+
+  async streamWalletTransactions(params: CSP.StreamWalletTransactionsParams) {
+    const client = await this.getClient(params.network);
+    const addresses = await this.getWalletAddresses(params.wallet._id!);
+    const readable = new Readable({ objectMode: true });
+    const promises = new Array<Promise<FormattedTransactionType[]>>();
+    for (const walletAddress of addresses) {
+      promises.push(client.getTransactions(walletAddress.address));
```
matiu: Could txs be pushed out of order on the stream?

packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+    const allTxs = await Promise.all(promises);
+    for (let txs of allTxs) {
+      this.streamTxs(txs, readable);
```
matiu: maybe we need to sort them here?

packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -0,0 +1,140 @@
...
+      limit: Number(params.args.limit) || 100,
+      binary: false
+    });
+    const readable = new Readable({ objectMode: true });
+    this.streamTxs(txs, readable);
+    readable.push(null);
+    Storage.stream(readable, params.req, params.res);
+  }
+
+  async streamTransactions(params: CSP.StreamTransactionsParams) {
```
matiu: This is for getting the TX from a certain block?

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -585,9 +585,15 @@ export class InternalStateProvider implements CSP.IChainStateService {
   }
 
   private extractAddress(address: string): string {
-    const extractedAddress = address
-      .replace(/^(bitcoincash:|bchtest:|bitcoin:)/i, '')
-      .replace(/\?.*/, '');
+    const extractedAddress = address.replace(/^(bitcoincash:|bchtest:|bitcoin:)/i, '').replace(/\?.*/, '');
```
matiu: can we just do `address.replace(^[a-z]*:/i,'')` [...]   ?

micahriggan: We could, I think this was just a whitespace change

packages/bitcore-node/src/modules/ripple/api/csp.ts
micahriggan: It could, for instance for accounting we may want to pull the transaction history of multiple addresses. Bitcore-node treats each wallet as a bucket of addresses, even if it only has one address

packages/bitcore-node/src/modules/ripple/api/csp.ts
micahriggan: Yeah, this should give you the transactions back for a given ledger, this method is likely not it's final implementation

packages/bitcore-node/src/types/namespaces/ChainStateProvider.ts
```diff
@@ -151,10 +152,10 @@ export declare namespace CSP {
     streamWalletTransactions(params: StreamWalletTransactionsParams): any;
     streamWalletUtxos(params: StreamWalletUtxosParams): any;
     streamMissingWalletAddresses(params: StreamWalletMissingAddressesParams);
-    getCoinsForTx(params: { chain: string; network: string; txid: string }): Promise<CoinListingJSON>;
-    getLocalTip(params): Promise<IBlock | null>;
+    getCoinsForTx(params: GetCoinsForTxParams): Promise<CoinListingJSON>;
+    getLocalTip(params): Promise<any | null>;
```
micahriggan: Let's keep this as IBlock, just need to transform the ledger into an IBlock

packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -0,0 +1,197 @@
...
+      const connUrl = portString ? `${protocol}://${host}:${portString}` : `${protocol}://${host}`;
+      RippleStateProvider.clients[network] = new RippleAPI({ server: connUrl });
+      await RippleStateProvider.clients[network].connect();
+    }
+    return RippleStateProvider.clients[network];
+  }
+
+  async getBalanceForAddress(params: CSP.GetBalanceForAddressParams) {
+    const client = await this.getClient(params.network);
+    const info = await client.getAccountInfo(params.address);
```
jackbonaguro: @micahriggan RippleAPI will throw an error if you give it a valid address but the account has not been created (by sending it the reserve amount of XRP)

micahriggan: Yeah, I've seen that one, I think we may use that to alert to the user that they'll need to deposit 21 or whatever.

Otherwise this should probably return 0.

jackbonaguro: I would think that error should be client-side, based on a balance of 0 (and the node should return 0). Otherwise you'd have to handle this special error in insight, copay, etc.

PR #2473, opened 2019-10-16
Wallet import should tag old transactions

micahriggan:


packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -370,7 +370,12 @@ export class ETHStateProvider extends InternalStateProvider implements CSP.IChai
   }
 
   async getAccountNonce(network: string, address: string) {
-    return EthTransactionStorage.collection.countDocuments({ chain: 'ETH', network, from: address, blockHeight: { $ne: -1 } });
+    return EthTransactionStorage.collection.countDocuments({
+      chain: 'ETH',
+      network,
+      from: address,
+      blockHeight: { $ne: -1 }
```
nitsujlangston: `$ne: -1` seems odd here

justinkook: This is to filter out mempool transactions that have not been mined yet

PR #2477, opened 2019-10-17
New endpoint - /block/:blockHeight/coins

ralexei:
-Added a new endpoint to bitcore-node: /block/:blockHeight/coins, which returns all the coins from a blockHeight

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/index.ts
+++ b/packages/bitcore-node/src/providers/chain-state/index.ts
@@ -11,8 +11,8 @@ class ChainStateProxy implements CSP.ChainStateProvider {
     return services[chain];
   }
 
-  streamUnspentsByBlock(params: CSP.GetUnspentsFromBlockParams) {
-    return this.get(params).streamUnspentsByBlock(params);
+  streamCoinsByBlock(params: CSP.StreamCoinsByBlockParams) {
+    return this.get(params).streamCoinsByBlock(params);
   }
 
   streamAddressUtxos(params: CSP.StreamAddressUtxosParams) {
--- a/packages/bitcore-node/src/providers/chain-state/index.ts
+++ b/packages/bitcore-node/src/providers/chain-state/index.ts
@@ -11,6 +11,10 @@ class ChainStateProxy implements CSP.ChainStateProvider {
     return services[chain];
   }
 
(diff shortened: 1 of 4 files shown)


packages/bitcore-node/src/providers/chain-state/index.ts
```diff
@@ -11,6 +11,10 @@ class ChainStateProxy implements CSP.ChainStateProvider {
     return services[chain];
   }
 
+  streamUnspentsByBlock(params: CSP.GetUnspentsFromBlockParams) {
```
micahriggan: Thanks for this! :)

I'd prefer `streamCoinsByBlock` or `streamCoinsForBlock`, since this implies to me the coins won't be spent

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -565,6 +565,24 @@ export class InternalStateProvider implements CSP.IChainStateService {
...
+    const { limit, since, paging } = args;
+
+
+    if (typeof params.blockHeight !== 'number' || !params.chain || !params.network)
+      throw 'Missing required param';
+
+    let query = {
+      chain: params.chain,
+      network: params.network,
+      mintHeight: params.blockHeight
```
micahriggan: It would be cool if we had a `includeSpent` arg, then you could do something like 
```
if(includeSpent) {
  query.$or = [{mintHeight: params.bockHeight}, {spentHeight: params.blockHeight}]
} else {
  query.mintHeight = params.blockHeight
}
```

packages/bitcore-node/src/providers/chain-state/index.ts
ralexei: Done.
Please check :)

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
ralexei: > It would be cool if we had a `includeSpent` arg, then you could do something with this effect
> 
> ```
> if(includeSpent) {
>   query.$or = [{mintHeight: params.bockHeight}, {spentHeight: params.blockHeight}]
> } else {
>   query.mintHeight = params.blockHeight
> }
> ```

Done.
Please check :)

FXManiacs:
@micahriggan who should resolve the conflict?

PR #2478, opened 2019-10-17
Static Reference to Mongo Connection

micahriggan:
Share the mongo connection with a static reference rather than one connection per model

# Master
```
{
    "current" : 200,
    "available" : 1438,
    "totalCreated" : 200,
    "active" : 1
}
```

# This Branch
```
{
    "current" : 184,
    "available" : 20,
    "totalCreated" : 184,
    "active" : 1
}
```


micahriggan:
After checking, doesn't seem like this solves any issue.
I logged the number of connections established and they are the same on master and this branch.

PR #2490, opened 2019-10-25
support eth testnet

matiu:


micahriggan:
Is this appending :t at the end of addresses before they go into storage?

Would this cause an issue if we had a coin with multiple testnets? 
I was thinking having unique on {address, network} would allow for the same address on multiple networks.

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -170,6 +170,21 @@ export class EthChain implements IChain {
   setInputs() {}
 
   isUTXOCoin() { return false; }
+  isSingleAddress() { return true; }
+
+  addressFromStorageTransform(network, address): void {
+    if (network != 'livenet') {
+      const x =  address.address.indexOf(':t');
```
micahriggan: indexOf(":" + network)

PR #2491, opened 2019-10-25
[REF]: callback replacement with promises in server.ts

rastajpa:


packages/bitcore-wallet-service/test/integration/server.js
```diff
@@ -1463,14 +1460,14 @@ describe('Wallet service', function() {
...
-          server.getWallet({}, (err, w) => {
+          server.getWallet({}).then(w => {
 
             var old = server.getWallet;
             server.getWallet = sinon.stub();
 
             // return main address index to 0;
             w.addressManager.receiveAddressIndex = 0;
-            server.getWallet.callsArgWith(1, null, w);
+            server.getWallet.returns(new Promise((resolve) => { resolve(w) }));
```
micahriggan: you can use yields instead of new promise

micahriggan:
Merge conflicts

packages/bitcore-wallet-service/test/integration/server.js
rastajpa: I think yield is for functions with callbacks... So I replaced the new promise with `resolves`. Let me know if I am wrong

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1464,20 +1473,21 @@ export class WalletService {
...
    * @returns {truthy} The result of the verification.
    */
-  verifyMessageSignature(opts, cb) {
-    if (!checkRequired(opts, ['message', 'signature'], cb)) return;
+  verifyMessageSignature(opts) {
+    return new Promise((resolve) => {
 
-    this.getWallet({}, (err, wallet) => {
-      if (err) return cb(err);
+      if (!checkRequired(opts, ['message', 'signature'])) return;
```
Gamboster: Should this `return` be empty?

packages/bitcore-wallet-service/test/integration/server.js
```diff
@@ -697,8 +697,7 @@ describe('Wallet service', function() {
           helpers.getAuthServer(copayerId, function(server) {
             server.getStatus({
               includeExtendedInfo: true
-            }, function(err, status) {
-              should.not.exist(err);
```
Gamboster: Should be tested a `catch(err => should.not.exist(err))`

PR #2497, opened 2019-10-30
Daily Transactions Range

micahriggan:
Daily transactions should accept a start and end range, and default to the past 30 days
Should not default to aggregating the whole transactions collection.

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/internal/internal.ts
+++ b/packages/bitcore-node/src/providers/chain-state/internal/internal.ts
@@ -474,8 +474,12 @@ export class InternalStateProvider implements CSP.IChainStateService {
     oneMonth.setDate(todayTruncatedUTC.getDate() - 30);
     oneMonth = formatDate(oneMonth);
 
-    const start = startDate ? new Date(startDate) : oneMonth;
-    const end = endDate ? formatDate(new Date(endDate)) : todayTruncatedUTC;
+    const isValidDate = (d: string) => {
+      return new Date(d).toString() !== 'Invalid Date';
+    };
+    const start = startDate && isValidDate(startDate) ? new Date(startDate) : oneMonth;
+    const end = endDate && isValidDate(endDate) ? formatDate(new Date(endDate)) : todayTruncatedUTC;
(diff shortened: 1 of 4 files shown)


packages/bitcore-node/src/routes/api/stats.ts
```diff
@@ -10,7 +10,12 @@ router.get('/', async function(_: Request, res: Response) {
...
   try {
-    let dailyTxs = await ChainStateProvider.getDailyTransactions({ chain, network });
+    let dailyTxs = await ChainStateProvider.getDailyTransactions({
+      chain,
+      network,
+      startDate: req.query.startDate,
+      endDate: req.query.endDate
+    });
     if (!dailyTxs) {
       return res.send(500);
```
matiu: I dont think 500 is correct here.

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -466,9 +466,16 @@ export class InternalStateProvider implements CSP.IChainStateService {
...
-    const todayTruncatedUTC = new Date(new Date().toISOString().split('T')[0]);
+  async getDailyTransactions(params: CSP.DailyTransactionsParams) {
+    const { chain, network, startDate, endDate } = params;
+    const formatDate = (d: Date) => new Date(d.toISOString().split('T')[0]);
+    const todayTruncatedUTC = formatDate(new Date());
+    let oneMonth = new Date(todayTruncatedUTC);
+    oneMonth.setDate(todayTruncatedUTC.getDate() - 30);
+    oneMonth = formatDate(oneMonth);
+
+    const start = startDate ? new Date(startDate) : oneMonth;
```
matiu: is request's `startDate` / `endDate` are invalid, "new Date(startDate)" will return "invalid Date"
that should be handled.

packages/bitcore-node/src/routes/api/stats.ts
micahriggan: This line should never happen anyway, I'm removing it.

PR #2500, opened 2019-10-31
Base URL was being mutated when calling the constructor with baseUrl

micahriggan:


packages/bitcore-client/src/wallet.ts
```diff
@@ -45,17 +45,19 @@ export class Wallet {
...
-    if (this.baseUrl) {
-      this.baseUrl = `${this.baseUrl}/${this.chain}/${this.network}`;
-    } else {
-      this.baseUrl = `https://api.bitcore.io/api/${this.chain}/${this.network}`;
+    if (!this.baseUrl) {
+      this.baseUrl = `https://api.bitcore.io/api`;
     }
     this.client = new Client({
-      baseUrl: this.baseUrl,
+      baseUrl: this.getApiUrl(),
```
matiu: maybe renaming Client's `baseUrl` to `apiUrl` will make the code easier to read

PR #2504, opened 2019-11-01
Reskin insight

rachitajoshi:


micahriggan:
This one looks good to me, after this goes in we'll make the block details rows clickable.

PR #2520, opened 2019-11-07
Refactor rpc so all methods async, sendRawTransaction handles arrays

micahriggan:


packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -425,15 +425,13 @@ export class InternalStateProvider implements CSP.IChainStateService {
...
-      this.getRPC(chain, network).sendTransaction(rawTx, (err: any, result: any) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve(result);
-        }
-      });
-    });
+    const txids = new Array<string>();
+    const txs = typeof rawTx === 'string' ? [rawTx] : rawTx;
```
micahriggan: rawTxs

justinkook:
Screen Shot 2019-11-07 at 2 35 56 PM

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -425,15 +425,13 @@ export class InternalStateProvider implements CSP.IChainStateService {
...
-          reject(err);
-        } else {
-          resolve(result);
-        }
-      });
-    });
+    const txids = new Array<string>();
+    const rawTxs = typeof rawTx === 'string' ? [rawTx] : rawTx;
+    for (const tx of rawTxs) {
+      const txid = await this.getRPC(chain, network).sendTransaction(tx);
```
justinkook: needs error handling `try catch` or `.catch`

justinkook:
This MR fixes:
```
[warn] ERROR: 500 - "Returned error: Invalid params: invalid type: sequence, expected a 0x-prefixed, hex-encoded vector of bytes." 
```

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
micahriggan: This is for bitcoin, so I actually want it to throw if it fails

PR #2534, opened 2019-11-09
Adding caching to stats endpoint

micahriggan:


packages/bitcore-node/src/routes/api/stats.ts
```diff
@@ -7,16 +7,35 @@ router.get('/', async function(_: Request, res: Response) {
   return res.send(404);
 });
 
+let cache = {};
+let cacheExpiry = Date.now() + CacheTimes.Day;
```
nitsujlangston: cache expiration would need to be on `cache[`${chain}:${network}`]`

```
cache[`${chain}:${network}`] = { dailyTxs, expiration}
```

packages/bitcore-node/src/routes/api/stats.ts
```diff
@@ -1,22 +1,43 @@
...
+          startDate: '',
+          endDate: ''
+        });
+        cache[cacheKey] = { dailyTxs, expiry: Date.now() + CacheTimes.Day };
+      }
+    } catch (e) {
+      logger.error(e);
+    } finally {
+      updating = false;
+      return cache[cacheKey].dailyTxs;
```
nitsujlangston: if there is an error, there wont be a `cache[cacheKey]` to have a dailyTxs

packages/bitcore-node/src/routes/api/stats.ts
```diff
@@ -1,22 +1,42 @@
...
-      network,
-      startDate: req.query.startDate,
-      endDate: req.query.endDate
-    });
-    SetCache(res, CacheTimes.Day);
+    const { dailyTxs } = await updateCache();
+    SetCache(res, CacheTimes.Day, CacheTimes.Hour);
     return res.json(dailyTxs);
   } catch (err) {
     return res.status(500).send(err);
```
matiu: why we are sending the err back to the client?
should we just print it to the log and return "Internal server error" message or something?

It seems to me we are disclosing to much information on a production server.

PR #2537, opened 2019-11-11
Fix(Node) - ERC20 txs Array.sort

justinkook:
~~- add .sort({ blockTimeNormalized: 1 })~~ => Sort by `blockHeight` using Array.sort Javascript.

Change `erc20.forEach` to `for (const tx of sortedTxs) {}`

### DIFF
--- a/packages/bitcore-node/src/modules/ethereum/api/csp.ts
+++ b/packages/bitcore-node/src/modules/ethereum/api/csp.ts
@@ -320,7 +320,10 @@ export class ETHStateProvider extends InternalStateProvider implements CSP.IChai
         ]
       };
       console.log(JSON.stringify(query));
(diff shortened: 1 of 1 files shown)


justinkook:
@nitsujlangston 
Would it be more efficient to Array.sort javascript method instead of mongo sort? 

There is no `blockTimeNormalized` index with the ERC20 query unless we add `blockTimeNormalized` to the index:

```
this.collection.createIndex(
      { chain: 1, network: 1, 'abiType.params.0.value': 1, blockTimeNormalized: 1 },
      {
        background: true,
        partialFilterExpression: { chain: 'ETH', 'abiType.type': 'ERC20', 'abiType.name': 'transfer' }
      }
    );
```

PR #2547, opened 2019-11-13
fix(bwc): create address at first time if coin is any token

cmgustavo:


packages/bitcore-wallet-client/src/lib/common/utils.ts
```diff
@@ -162,7 +162,7 @@ export class Utils {
   static deriveAddress(scriptType, publicKeyRing, path, m, network, coin) {
     $.checkArgument(_.includes(_.values(Constants.SCRIPT_TYPES), scriptType));
 
-    coin = coin || 'btc';
+    coin = Constants.ERC20.includes(coin) ? 'eth' : (coin || 'btc');
```
justinkook: I think it would be better to use `this.getChain(coin).toLowerCase()`

justinkook: So every ERC20 token will be `eth` and ETH will also be `eth`

cmgustavo: When coin is `usdc` (it happens when you import or create a new token if you have ever opened the receive view of ETH wallet). Then bitcore[usdc] doesn't exist, so we get an error. 

I think that every ERC20 token is based on ETH (that's why it requires to select your ETH wallet before create the token), but I'm not sure. Please correct me if I'm wrong.

justinkook: Correct thats why I think we should use `chain`

cmgustavo: +1 Thanks!!

justinkook:
Tested on Copay using BWC branch. Address derivation works and no longer get error.

👍

PR #2551, opened 2019-11-13
Fix(BWS) - Duplicate Txps

justinkook:
If token wallet:
- do not assign token wallet.pendingTxs causing duplicate txps

Problem:
ETH and token wallets share the same `this.walletId`

So `this.getPendingTxs` is being assigned twice both to ETH and Token wallet.

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -770,7 +770,7 @@ export class WalletService {
           });
         },
         (next) => {
-          this.getPendingTxs({}, (err, pendingTxps) => {
+          this.getPendingTxs(opts, (err, pendingTxps) => {
             if (err) return next(err);
             status.pendingTxps = pendingTxps;
             next();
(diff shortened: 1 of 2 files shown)


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -3227,6 +3227,9 @@ export class WalletService {
    * @returns {TxProposal[]} Transaction proposal.
    */
   getPendingTxs(opts, cb) {
+    if (opts.tokenAddress) {
```
matiu: why tokenAddress can't have pending tx?

justinkook: Because token wallets are duplicate ETH wallets. So anytime we make a `txp`, there will be two duplicate `txps`. `walletEth.pendingTxs`

justinkook: token wallets do not get stored on the BWS/Node wallets collection so txps cannot get shared between devices for token wallets

PR #2560, opened 2019-11-15
Bug/fromobj

matiu:
- bump to 8.11
- upgrades `new Buffer` in mnemonics
- fixes (toObj) in BWS (not used at the moment)


packages/bitcore-wallet-client/src/lib/credentials.ts
```diff
@@ -132,6 +134,21 @@ export class Credentials {
...
   }
+
+  /* 
+   * creates an ERC20 wallet from a ETH wallet
+   */
+  getTokenCredentials(token: { name: string, symbol: string, address:string }) {
+    const ret = _.cloneDeep(this);
+    ret.walletId = `${ret.walletId}-${token.address}`;
+    ret.coin = token.symbol.toLowerCase();
+    ret.walletNane = token.name;
```
justinkook: walletName

packages/bitcore-wallet-client/src/lib/credentials.ts
```diff
@@ -132,6 +134,21 @@ export class Credentials {
 
     return x;
   }
+
+  /* 
+   * creates an ERC20 wallet from a ETH wallet
+   */
+  getTokenCredentials(token: { name: string, symbol: string, address:string }) {
```
justinkook: `address:string` needs space between

packages/bitcore-wallet-client/src/lib/credentials.ts
```diff
@@ -63,14 +64,15 @@ export class Credentials {
   use145forBCH: any;
   addressType: string;
   keyId: string;
-
+  token: string;
```
justinkook: `token?: string` will not always be available for every wallet.

matiu: https://github.com/bitpay/bitcore/pull/2562

packages/bitcore-wallet-client/src/lib/credentials.ts
matiu: https://github.com/bitpay/bitcore/pull/2562

packages/bitcore-wallet-client/src/lib/credentials.ts
matiu: we are using 
```
> tslint --fix -c tslint.json 'src/**/*.ts'
```
and it does not force that :/

PR #2569, opened 2019-11-18
Feature/ethereum verification

micahriggan:
# Goal
Create a way to multi-plex the verification script without maintaining a verify script per currency.

Then use this to verify the ETH chain.

# Solution
Create a [Verification](packages/bitcore-node/src/services/verification.ts) service, same as [P2P service](https://github.com/bitpay/bitcore/blob/master/packages/bitcore-node/src/services/p2p.ts), and register VerificationWorkers per chain.

The [verification peer](packages/bitcore-node/src/modules/ethereum/p2p/EthVerificationPeer.ts) should function as a P2P worker, only it doesn't process mempool transactions, or start syncing.

# Additional 
* [Add parity to docker-compose](https://github.com/bitpay/bitcore/pull/2569/files#diff-0d4080c652c491057176df93ca5768d1R42)
* [Tests that use parity node](packages/bitcore-node/test/integration/ethereum/p2p.spec.ts)
* [register verification worker for each chain](https://github.com/bitpay/bitcore/pull/2569/files#diff-a41b9270e433d486b48c7de1c40fee23R12)
* [Move internal txs logic into convertBlock](https://github.com/bitpay/bitcore/pull/2569/files#diff-d5a256b8711df11e8125dd8d33ea1170R275)
* [Adding tip parameter to verify function](https://github.com/bitpay/bitcore/pull/2569/files#diff-93e5d9954cc14dd759f4d480c52effb3R47)


packages/bitcore-node/src/modules/ethereum/p2p/p2p.ts
```diff
@@ -157,49 +186,17 @@ export class EthP2pWorker extends BaseP2PWorker<IEthBlock> {
...
-                convertedTxs[errorIndex].error = tx.error;
-              }
-            }
-          }
-        }
-
+        const block = await this.getBlock(currentHeight);
+        const { convertedBlock, convertedTxs } = await this.convertBlock(block);
         await this.processBlock(convertedBlock, convertedTxs);
-        currentHeight++;
```
micahriggan: This caused an issue

packages/bitcore-node/src/modules/ethereum/p2p/p2p.ts
```diff
@@ -157,49 +186,17 @@ export class EthP2pWorker extends BaseP2PWorker<IEthBlock> {
...
         if (currentHeight === bestBlock) {
           bestBlock = await this.web3!.eth.getBlockNumber();
         }
+        tip = await ChainStateProvider.getLocalTip({ chain, network });
```
micahriggan: this should prevent progressing passed non-processed blocks.

packages/bitcore-node/src/modules/bitcoin/VerificationPeer.ts
```diff
@@ -71,10 +72,22 @@ export class VerificationPeer extends BitcoinP2PWorker {
...
 
-  async resync(from: number, to: number) {
+  async getBlockForNumber(currentHeight: number) {
     const { chain, network } = this;
-    let currentHeight = Math.max(1, from);
-    while (currentHeight < to) {
+    const locatorHashes = await ChainStateProvider.getLocatorHashes({
+      chain,
+      network,
+      startHeight: Math.max(1, currentHeight - 30),
```
matiu: where this 30 is coming from? Why we should return a hash that is not the provided Height?

micahriggan: This one is really tricky actually, it comes from https://github.com/bitpay/bitcore/blob/22d121cf9dbe180fc7e4599acf35bd90f9164994/packages/bitcore-node/src/providers/chain-state/internal/internal.ts#L553

Basically getLocatorHashes returns 30 hashes, and the most recent one is first in the list. 

So to use that to get the current hash, you'd need to start from 30 blocks in the past.

matiu: tricky, indeed :)

packages/bitcore-node/src/modules/bitcoin/VerificationPeer.ts
```diff
@@ -71,10 +72,19 @@ export class VerificationPeer extends BitcoinP2PWorker {
...
+      .limit(1)
+      .toArray();
+    return this.getBlock(hash);
+  }
+
+  async resync(start: number, end: number) {
+    const { chain, network } = this;
+    let currentHeight = Math.max(1, start);
+    while (currentHeight < end) {
       const locatorHashes = await ChainStateProvider.getLocatorHashes({
```
matiu: use `.fineOne` also here?

matiu: Also `.findOne` here?

PR #2575, opened 2019-11-19
Feature(CWC) - XRP create to from async to sync

justinkook:
Added additional functions

- getSignatureObject()
- getSignature()
- getHash()
- applySignature()

Fixed `validateUri` for ETH and XRP 
- add URI validation tests for XRP


packages/crypto-wallet-core/src/transactions/xrp/index.ts
```diff
@@ -1,64 +1,109 @@
...
+  getSignatureObject (params: { tx: string; key: Key; }) {
     const { tx, key } = params;
-    const txJSON = tx;
+    const txJSON = rippleBinaryCodec.decode(tx);
     let rippleAPI = new RippleAPI();
-    const signedTx = rippleAPI.sign(txJSON, {
-      privateKey: key.privKey,
-      publicKey: key.pubKey,
+    const signedTx = rippleAPI.sign(JSON.stringify(txJSON), {
+      privateKey: key.privKey.toUpperCase(),
```
micahriggan: why does this need to be uppercase?

justinkook: Ripple validation throws an error if not uppercase.

justinkook:
```
1) Transaction Sign
       should not fail to sign a XRP tx with lowercase key:
     Error: `e8` is not a valid hex representation of a byte
      at byteForHex (node_modules/ripple-binary-codec/distrib/npm/utils/bytes-utils.js:38:11)
      at parseBytes (node_modules/ripple-binary-codec/distrib/npm/utils/bytes-utils.js:52:18)
      at new Blob (node_modules/ripple-binary-codec/distrib/npm/types/blob.js:9:21)
      at Function.from (node_modules/ripple-binary-codec/distrib/npm/types/blob.js:22:14)
      at /Users/justin/bitpay/bitcore/packages/crypto-wallet-core/node_modules/ripple-binary-codec/distrib/npm/types/st-object.js:31:46
```

justinkook: If I remove `.toUpperCase()` this error returns.

matiu: Strange!

PR #2581, opened 2019-11-21
Feat/erc20 token recovery

matiu:
Adds support in preferences to store 'tokenAddresses'
with hint the server of token wallets paired to a ETH wallet.

BWC's `.serverAssistedImport` will return the tokenWallets ready to use.

### DIFF
--- a/packages/bitcore-wallet-client/src/lib/api.ts
+++ b/packages/bitcore-wallet-client/src/lib/api.ts
@@ -2544,6 +2544,10 @@ export class API extends EventEmitter {
                   cont = !!iclient;
                   if (iclient) {
                     clients.push(iclient);
+
+                    // expand credentials with tokens?
+                    
+
                   } else {
                     // we do not allow accounts nr gaps in BWS.
                     cont = false;
--- a/packages/bitcore-wallet-client/src/lib/api.ts
+++ b/packages/bitcore-wallet-client/src/lib/api.ts
@@ -12,6 +12,28 @@ import { PayProV2 } from './payproV2';
 import { Request } from './request';
 import { Verifier } from './verifier';
 
+// TODO put somewhere else
+export const TokenOpts = {
+  '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48': {
+    name: 'USD Coin',
+    symbol: 'USDC',
+    decimal: 6,
+    address: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
+  },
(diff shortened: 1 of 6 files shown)


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1203,7 +1206,16 @@ export class WalletService {
             _.isString(value) && _.includes(['btc', 'bit'], value.toLowerCase())
           );
         }
+      },
+      {
+        name: 'tokenAddresses',
+        isValid(value) {
+          return (
+            _.isArray(value) && value.every(x => x.length < 45 && x.indexOf('0x') == 0)
+          );
```
justinkook: Can use CWC ETH address validation here. All tokenAddresses are ETH addresses.

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -12,6 +12,28 @@ import { PayProV2 } from './payproV2';
...
+    address: '0x8e870d67f660d95d5be530380d0ec0bd388289e1'
+  },
+  '0x056fd409e1d7a124bd7017459dfea2f387b6d5cd': {
+    name: 'Gemini Dollar',
+    symbol: 'GUSD',
+    decimal: 2,
+    address: '0x056fd409e1d7a124bd7017459dfea2f387b6d5cd'
+  }
+};
+
```
justinkook: Move to `Constants.Common.Constants`

packages/bitcore-wallet-service/src/lib/server.ts
matiu: good idea, thanks!

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1211,7 +1211,18 @@ export class WalletService {
...
         isValid(value) {
           return (
-            _.isArray(value) && value.every(x => x.length < 45 && x.indexOf('0x') == 0)
+            _.isArray(value) && value.every(x => {
+              try {
+                Validation.validateAddress(
+                  'eth',
+                  'mainnet',
+                  x,
+                );
```
justinkook: This function will not throw if its just an invalid address, it only throws if theres a real error. 

return the function instead:
```
return Validation.validateAddress(
                  'eth',
                  'mainnet',
                  x,
                );
```

matiu: ok, then we have an error on _validateAddr, fixing.

PR #2587, opened 2019-11-22
Bitcore-Node: ETH Verify + Refactor

micahriggan:
Refactoring verification logic to be on VerificationPeers, so you can have different verification logic per chain.

Adding tests for verification


packages/bitcore-node/src/modules/ethereum/p2p/EthVerificationPeer.ts
```diff
@@ -57,4 +68,150 @@ export class EthVerificationPeer extends EthP2pWorker implements IVerificationPe
   async getBlockForNumber(blockNum: number) {
     return this.getBlock(blockNum);
   }
+
+  async validateDataForBlock(blockNum: number, tipHeight: number, log = false) {
+    let success = true;
```
micahriggan: duplicated from https://github.com/bitpay/bitcore/pull/2587/files#diff-93e5d9954cc14dd759f4d480c52effb3L48

Removed logic for coins

packages/bitcore-node/src/modules/bitcoin/VerificationPeer.ts
```diff
@@ -127,4 +144,257 @@ export class VerificationPeer extends BitcoinP2PWorker implements IVerificationP
       }
     }
   }
+
+  async validateDataForBlock(blockNum: number, tipHeight: number, log = false) {
+    let success = true;
```
micahriggan: Moved from 

https://github.com/bitpay/bitcore/pull/2587/files#diff-93e5d9954cc14dd759f4d480c52effb3L48

PR #2589, opened 2019-11-22
Should return undefined when a XRP tx cannot be found

micahriggan:


packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -178,8 +178,12 @@ export class RippleStateProvider extends InternalStateProvider implements CSP.IC
 
   async getTransaction(params: CSP.StreamTransactionParams) {
     const client = await this.getClient(params.network);
-    const tx = await client.getTransaction(params.txId);
-    return tx;
+    try {
+      const tx = await client.getTransaction(params.txId);
+      return tx;
+    } catch (e) {
+      return undefined;
```
micahriggan: This matches how internal.ts handles transactions that could not be found.

justinkook:
Tested on BWS:
returns
```
{"message":"16:00:26 GMT-0500 (EST) |       ::ffff:127.0.0.1 |      END |     348 ms |    GET |   404 | /api/XRP/testnet/tx/31FB88E20BB096E9E386B2CBA16E78AB8B53A696CA7031A392B70917B152AA36","level":"info"
}
```

PR #2593, opened 2019-11-25
Fix(BWS) - Adds ETH wallet balance check for ERC20 miner fee

justinkook:


matiu:
Please add tests.

AVGolden:
I appreciate the hard work you do and I no everyone else as well are approving of all this hard work you put into the matter

PR #2594, opened 2019-11-25
Feat/support

matiu:


packages/bitcore-wallet-service/src/scripts/v8tool-list.ts
```diff
@@ -6,10 +6,11 @@ const Bitcore = require('bitcore-lib');
 const requestStream = require('request');
 import { Client } from '../lib//blockchainexplorers/v8/client';
 
-const coin = 'BTC';
+const coin = 'ETH';
 console.log('COIN:', coin);
 const network = 'mainnet';
 const authKey = process.argv[2];
+const extra = process.argv[3] ||  '';
```
micahriggan: could we pass coin here, and then use the coin to select the baseUrl?

Not 100% sure of what this is used for, but it seems like you could make it work for all coins with that one change, if it doesn't already

matiu: yeah, it is just a script I use to determint is certain errors are on BWS or in  bitcore-node. Check balances and history mainly. needs  a lot of love yet :)

PR #2607, opened 2019-12-03
Prune mempool transactions from the database that are not found

micahriggan:


packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -146,6 +146,14 @@ export class ETHStateProvider extends InternalStateProvider implements CSP.IChai
           confirmations = tipHeight - found.blockHeight + 1;
         }
         const convertedTx = EthTransactionStorage._apiTransform(found, { object: true }) as EthTransactionJSON;
+        if (convertedTx.blockHeight === SpentHeightIndicators.pending) {
+          const web3 = await this.getWeb3(network);
+          const web3Tx = await web3.eth.getTransaction(convertedTx.txid);
```
matiu: What would happen is the node is temporarily out of service?

PR #2616, opened 2019-12-05
Broadcast should also send to infura and etherscan

micahriggan:


packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -168,6 +170,12 @@ export class ETHStateProvider extends InternalStateProvider implements CSP.IChai
           .on('transactionHash', resolve)
           .on('error', reject)
           .catch(e => reject(e));
+        if (Etherscan.supportsNetwork(network)) {
```
nitsujlangston: shouldnt these be wrapped in a catch?

micahriggan: I'm not sure, does request throw? I didn't think it did

matiu:
Is this one still active?

PR #2634, opened 2019-12-12
Fix/catch decode throws

micahriggan:


micahriggan:
I'm not sure what data triggered the throw, I have a test that catches garbage data, and it never threw, but some block in production contained a tx that caused it to throw.

micahriggan:
@matiu found some data that would cause this crash and added a test,

Thanks to #2636

PR #2635, opened 2019-12-12
Update web3 to 1.2.4 and fix compile errors

justinkook:
Removed unused dependencies from `bitcore-client`.

Bitcore-client was using `scrypt` from `web3`, so manually added `scrypt` and removed `web3`


packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -13,15 +13,24 @@ import { SpentHeightIndicators } from '../../../types/Coin';
 import { EthListTransactionsStream } from './transform';
 import { ERC20Abi } from '../abi/erc20';
 import { Transaction } from 'web3/eth/types';
-import { EventLog } from 'web3/types';
```
justinkook: `EventLog` type was removed in `web3@1.2.4`

packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -13,15 +13,24 @@ import { SpentHeightIndicators } from '../../../types/Coin';
...
+  blockHash: string;
+  blockNumber: number;
+  raw?: {data: string; topics: any[]};
+}
 interface ERC20Transfer extends EventLog {
   returnValues: {
-    _from: string;
-    _to: string;
-    _value: string;
+    [key: string]: string;
```
justinkook: `EventData[]` returnValues typing is:

```
  [key: string]: any;
```

Not compatible with:
```
  _from: string;
  _to: string;	
  _value: string;
```

packages/bitcore-node/src/modules/ethereum/p2p/parityRpc.ts
```diff
@@ -91,7 +91,7 @@ export class ParityRPC {
 
   public send<T>(data: JsonRPCRequest) {
     return new Promise<T>((resolve, reject) => {
-      this.web3.eth.currentProvider.send(data, function(err, data) {
+      this.web3.eth.currentProvider!['send'](data, function(err, data) {
```
justinkook: `this.web3.eth.currentProvider` type is `provider`

```
    | HttpProvider
    | IpcProvider
    | WebsocketProvider
    | string
    | null;
```

The `!` gets rid of `| null`

Each Provider has a `send()` property:
ex. `WebsocketProvider.send(), HttpProvider.send(), ...`

`string` does not have a send property so `['send']`

packages/crypto-wallet-core/src/derivation/eth/index.ts
```diff
@@ -28,7 +28,7 @@ export class EthDeriver implements IDeriver {
     const x = ecPoint.getX().toBuffer({ size: 32 });
     const y = ecPoint.getY().toBuffer({ size: 32 });
     const paddedBuffer = Buffer.concat([x, y]);
-    const address = utils.keccak256(paddedBuffer).slice(26);
+    const address = utils.keccak256(`0x${paddedBuffer.toString('hex')}`).slice(26);
```
justinkook: utils.keccak256 param takes in `typeof 'string | BN'`

Cannot pass in type `Buffer`

packages/bitcore-node/src/modules/ethereum/api/csp.ts
justinkook: https://github.com/ethereum/web3.js/blob/043cb81d40ede0f6ef1889222c5671f9678e409e/packages/web3-eth-contract/types/index.d.ts#L146

packages/bitcore-node/src/modules/ethereum/p2p/parityRpc.ts
justinkook: https://github.com/ethereum/web3.js/blob/043cb81d40ede0f6ef1889222c5671f9678e409e/packages/web3-core/types/index.d.ts#L413

packages/bitcore-node/src/modules/ethereum/api/csp.ts
justinkook: https://github.com/ethereum/web3.js/blob/043cb81d40ede0f6ef1889222c5671f9678e409e/packages/web3/types/index.d.ts

packages/crypto-wallet-core/src/derivation/eth/index.ts
justinkook: https://github.com/ethereum/web3.js/blob/a5e832fc20088cf4c4e916bc0ec3107ceb4428d0/packages/web3-utils/types/index.d.ts#L93

packages/bitcore-node/src/modules/ethereum/p2p/parityRpc.ts
micahriggan: lets do 
```
let provider = this.web3.eth.currentProvider as HttpProvider;
provider.send(data
```

packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -13,15 +13,24 @@ import { SpentHeightIndicators } from '../../../types/Coin';
 import { EthListTransactionsStream } from './transform';
 import { ERC20Abi } from '../abi/erc20';
 import { Transaction } from 'web3/eth/types';
-import { EventLog } from 'web3/types';
 import { partition } from '../../../utils/partition';
 import { WalletAddressStorage } from '../../../models/walletAddress';
-
+import { Contract } from 'web3-eth-contract';
+type ContractAbiInput = ConstructorParameters<typeof Contract>[0];
+interface EventLog {
```
micahriggan: interface EventLog<T> {
   ...,
   returnValues: T;
}

packages/bitcore-node/src/modules/ethereum/api/csp.ts
micahriggan:
```
interface ERC20Transfer extends EventLog<{
  _from: string;
  _to: string;	
  _value: string;
}>
```

packages/crypto-wallet-core/src/derivation/eth/index.ts
matiu: This change seems to be covered but test "should be able to generate a valid ETH address" and "should be able to generate a valid ETH address, privKey, pubKey".  I think we need to include more cases, including edge cases, on the test suite, on a future PR.

PR #2638, opened 2019-12-16
adding tag field to newTx

sbhatnagar30:


micahriggan:
what is the difference between tag and invoiceID?

justinkook:
> what is the difference between tag and invoiceID?

destination tag to send to exchange accounts

PR #2639, opened 2019-12-18
Generate Blockchain explorers urls from config

micahriggan:
Currently when developing, if you were to point BWS to your local bitcore-node, you'd need to change it in both config.ts, and the blockchain explorers file.

This pull request aims to remove the need to modify the blockchain explorers file.

### DIFF
--- a/packages/bitcore-wallet-service/src/config.ts
+++ b/packages/bitcore-wallet-service/src/config.ts
@@ -21,42 +21,45 @@ module.exports = {
 
   storageOpts: {
     mongoDb: {
-      uri: 'mongodb://localhost:27017/bws',
-    },
+      uri: 'mongodb://localhost:27017/bws'
+    }
   },
   messageBrokerOpts: {
     //  To use message broker server, uncomment this:
     messageBrokerServer: {
-      url: 'http://localhost:3380',
-    },
+      url: 'http://localhost:3380'
+    }
   },
   blockchainExplorerOpts: {
-    btc: {
-      livenet: {
-        url: 'https://api.bitcore.io',
+    chains: {
+      btc: {
+        livenet: {
+          url: 'https://api.bitcore.io'
+        },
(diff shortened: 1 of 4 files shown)


packages/bitcore-wallet-service/src/config.ts
```diff
@@ -21,58 +21,61 @@ module.exports = {
...
-      },
-      testnet: {
-        url: 'https://api.bitcore.io',
-        regtestEnabled: false
-      },
-    },
-    bch: {
-      livenet: {
-        url: 'https://api.bitcore.io',
+    chains: {
```
micahriggan: Added this so we can distinguish chains from the socket key.

packages/bitcore-wallet-service/src/lib/blockchainexplorer.ts
```diff
@@ -1,28 +1,27 @@
...
-    bch: {
-      livenet: 'https://api.bitpay.com',
-      testnet: 'https://api.bitpay.com'
-    },
-    eth: {
-      livenet: 'https://api-eth.bitcore.io',
-      testnet: 'https://api-eth.bitcore.io'
+const v8Providers = Object.keys(config.blockchainExplorerOpts.chains).reduce(
+  (obj, chain) => {
+    obj[chain] = {};
```
micahriggan: Generate the old form from the config

packages/bitcore-node/src/config.ts
```diff
@@ -79,7 +79,7 @@ const Config = function(): ConfigType {
       },
       p2p: {},
       socket: {
-        bwsKeys: []
+        bwsKeys: ['02064c30a69f2954fde1a32355c3769cd812ae4339d814835467d0aed398617244']
```
micahriggan: Adding a default key so bws and bitcore-node can talk to each other on clean installs

PR #2640, opened 2019-12-19
Feature(BWC) - Dynamically derive path using CWC pathFor()

justinkook:


packages/bitcore-wallet-client/src/lib/credentials.ts
```diff
@@ -162,24 +162,7 @@ export class Credentials {
           purpose = '48';
           break;
       }
-
-      var coin = '0';
-      if (this.network != 'livenet' && this.coin !== 'eth') {
-        coin = '1';
-      } else if (this.coin == 'bch') {
```
micahriggan: Let's leave this bch block

```
if(this.coin === 'bch' && !this.use145forBCH) {
  coin = '0'
} else {
   coin = Deriver.pathFor(this.coin, this.network, this.account).replace(/44/i, purpose);
}
```

packages/bitcore-wallet-client/src/lib/credentials.ts
```diff
@@ -162,24 +163,10 @@ export class Credentials {
...
-        coin = '0';
-      } else if (this.coin == 'eth') {
-        coin = '60';
-      } else {
-        throw new Error('unknown coin: ' + this.coin);
+      if (!this.use145forBCH && this.coin === 'bch') {
+        coin = 'btc';
       }
-      return 'm/' + purpose + "'/" + coin + "'/" + this.account + "'";
+      return Deriver.pathFor(coin, this.network, this.account).replace(/44/i, purpose);
```
matiu: if account is 44, this will be an issue.

matiu: when coin= BCH,  some wallets uses the `coin` path 0 and other 145. this doesnt seems to take that in account.

packages/bitcore-wallet-client/src/lib/credentials.ts
```diff
@@ -162,24 +163,10 @@ export class Credentials {
...
-        } else {
-          coin = '0';
-        }
-      } else if (this.coin == 'btc') {
-        coin = '0';
-      } else if (this.coin == 'eth') {
-        coin = '60';
-      } else {
-        throw new Error('unknown coin: ' + this.coin);
+      if (!this.use145forBCH && this.coin === 'bch') {
```
matiu: it dont think this is a good idea. sending the wrong coin just to get the right path can easily generate problems in future changes.

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -278,25 +278,12 @@ export class Key {
...
-      }
-    } else if (opts.coin == 'btc') {
-      coinCode = '0';
-    } else if (opts.coin == 'eth') {
-      coinCode = '60';
-    } else {
-      throw new Error('unknown coin: ' + opts.coin);
+    let coin = opts.coin;
+    const network = opts.network || 'livenet';
+    if (this.use0forBCH && opts.coin === 'bch') {
```
matiu: si comment before. I would prefer you send an extra `opts` in pathFor with the flag: `use0forBCH` or something like that.

matiu:
concept ACK. please see the comments.

PR #2642, opened 2019-12-19
BWS - Fee request spam

micahriggan:
Right now bitcore-node gets absolutely slammed by fee requests

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/common/defaults.ts
+++ b/packages/bitcore-wallet-service/src/lib/common/defaults.ts
@@ -159,7 +159,7 @@ module.exports = {
   BLOCKHEIGHT_CACHE_TIME: 30 * 60 * 1000,
 
   // Cache time fee levels (in ms)
-  FEE_LEVEL_CACHE_DURATION: 6 * 60 * 1000,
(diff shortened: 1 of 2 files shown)


packages/bitcore-wallet-service/src/lib/common/defaults.ts
```diff
@@ -159,7 +159,7 @@ module.exports = {
   BLOCKHEIGHT_CACHE_TIME: 30 * 60 * 1000,
 
   // Cache time fee levels (in ms)
-  FEE_LEVEL_CACHE_DURATION: 6 * 60 * 1000,
+  FEE_LEVEL_CACHE_DURATION: 1 * 60 * 1000,
```
matiu: why less time?

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2004,11 +2004,6 @@ export class WalletService {
             );
           }
 
-          if (failed > 0) {
```
matiu: this is meant to prefent storing failed requests in the cache.

PR #2643, opened 2019-12-19
Feature(BWS) - Add dev/prod versions of config.ts

justinkook:
with corresponding env:dev / env:prod scripts


micahriggan:
Great idea! What happens if you pull this and run npm start without running one of those scripts?
Let's make sure there is some default.

It would be nice to unify BWS's config with the bitcore.config.json file
That way we could have defaults, and over-ride various things with bitcore.config.json

justinkook:
We could commit an initial version of `src/config/index.ts` that is a duplicate of `src/config/prod.ts` then `.gitignore` it after the first initial commit.

justinkook:
Added `npm run env:prod` to `npm run start` script for default confg

micahriggan:
The current change means you can't use npm start with the dev env.

I think rather than use file system things to do this, we should modify the config.ts to export something different if the process.env.NODE_ENV equals dev.

Something like

```
if (process.env.NODE_ENV === "dev") {
  module.exports = require('./config.dev');
} else {
  module.exports = require('./config.prod')
}
```

matiu:
ACK. Will merge later since it requires change on the deployment script.

PR #2666, opened 2020-01-08
adding lastLedgerSequence to wallet newTx

sbhatnagar30:


micahriggan:
Do you want to add it to CWC in this PR? (Assuming it isn't already in there)

sbhatnagar30:
we don't need this anymore right? closing

PR #2671, opened 2020-01-09
Bitcore-Client: Fix Register to work with URL without /api

micahriggan:
We really should only need to pass in api.bitcore.io without /api


micahriggan:
@sbhatnagar30 what do you think of this

micahriggan:
Will need to fix the tests to not pass in /api

sbhatnagar30:
this would require a migration of all wallets right? maybe not a good idea

sbhatnagar30:
Actually wouldn't really need to migrate, we would just re register wallets, so this change could be good

PR #2676, opened 2020-01-13
Observed in fresh Server Installation

maektwain:
We need make, gcc and g++ if we are doing without any development tools pre-installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,10 @@
 
 - Trusted P2P Peer
 - MongoDB Server >= v3.4
+- make g++ gcc 
 
 ### Checkout the repo
+
 
 ```sh
 git clone git@github.com:bitpay/bitcore.git


justinkook:
Hmm do you mean `cmake`?

maektwain:
in case of ubuntu it failed because of the make only doing ``` apt install make gcc g++ ``` seems to be solving the problem.

maektwain:
Another thing I noticed is no clear mention of nodejs and npm version defined to make the Bitcore source code run, sometimes It seems to be failing and reverting to a specific nodejs and npm versions seems to be solving the issue.

justinkook:
Ah the recommended node/npm versions for development is:

```
"node": "8.15.0"
"npm": "6.4.1"
```

In the case of iOS, cmake is required as well for xcode. Ubuntu as you pointed out needs make.

maektwain:
Seems like we need a clear pre-requisite for development and as well as production, what other flavours of OS we need to support and I can try with these to verify and see how does the installation work and what are the packages with proper names and version if we can tag them with links.

Since you mentioned about the development version could you also share what is the production level nodejs and npm.

justinkook:
Production versions for node / npm should be same as development:
```
"node": "10.16.3"
"npm": "6.4.1"
```

PR #2677, opened 2020-01-13
CWC: Adding max ledger offset and max ledger version

micahriggan:
So we can have XRP transactions expire


packages/crypto-wallet-core/src/transactions/xrp/index.ts
```diff
@@ -48,7 +62,8 @@ export class XRPTxProvider {
     const instructions: Instructions = {
       fee: feeStr,
       sequence: nonce,
-      maxLedgerVersion: null,
+      maxLedgerVersion,
```
justinkook: delete `maxLedgerVersion` from instructions

PR #2683, opened 2020-01-15
DAI Token Added

maektwain:
DAI Token added in BWS and Client Service. Currently it supports testnet address.

Let me know if this is needed in the BWS as of now.


maektwain:
So, this seems like it has commits which I did not want to add, let me push one more time.

PR #2684, opened 2020-01-15
make wallet methods work for all currencies

sbhatnagar30:


packages/crypto-wallet-core/src/transactions/btc/index.ts
```diff
@@ -49,15 +56,16 @@ export class BTCTxProvider {
   }
 
   getSignature(params: { tx: string; keys: Array<Key>}) {
-    throw new Error('No implemented for UTXO coins');
+    throw new Error('function getSignature not implemented for UTXO coins');
```
matiu: thanks for the better message.

packages/crypto-wallet-core/src/transactions/btc/index.ts
```diff
@@ -49,15 +56,16 @@ export class BTCTxProvider {
...
   }
 
   applySignature(params: { tx: string; keys: Array<Key>}) {
-    throw new Error('No implemented for UTXO coins');
+    throw new Error('function applySignature not implemented for UTXO coins');
   }
 
   getHash(params: { tx: string}) {
-    throw new Error('No implemented for UTXO coins');
+    const bitcoreTx = new this.lib.Transaction(params.tx);
```
matiu: 👍

PR #2693, opened 2020-01-23
[FIX] add token address to ERC20 push notifications data

gabrielbazan7:


matiu:
thanks for the updated tests!

PR #2694, opened 2020-01-23
Fix/add timeout handling for bwc PayPro call

bsnowden3:
This PR should be together with 
https://github.com/bitpay/copay/pull/10528


packages/bitcore-wallet-client/src/lib/payproV2.ts
```diff
@@ -72,6 +72,8 @@ export class PayProV2 {
       if (requestOptions.args) {
         if (requestOptions.method == 'post' || requestOptions.method == 'put') {
           r.send(requestOptions.args);
+        } else if (requestOptions.method === 'get') {
```
micahriggan: What if the timeout occurs on one of the posts?

bsnowden3: Addressed the post, by also adding a 10 second timeout. An action sheet will be thrown on copay's side when caught.

micahriggan:
this looks good to me @matiu want to take a look?

PR #2697, opened 2020-01-27
[FEAT] adding eth fee error to spec

gabrielbazan7:


packages/bitcore-wallet-client/src/lib/errors/spec.ts
```diff
@@ -49,9 +49,15 @@ var errorSpec = [{
 }, {
   name: 'INSUFFICIENT_FUNDS_FOR_FEE',
   message: 'Insufficient funds for fee.'
+}, {
+  name: 'INSUFFICIENT_ETH_FEE',
```
matiu: Why not reusing https://github.com/bitpay/bitcore/pull/2697/files#diff-e57a24c0f3a56c58bf7457e8e42b712cR50
? Adding the same message, per coin, may not be a good idea.

packages/bitcore-wallet-client/src/lib/errors/spec.ts
```diff
@@ -49,9 +49,15 @@ var errorSpec = [{
...
+  message: 'Your linked ETH wallet does not have enough ETH for fee.'
 }, {
   name: 'LOCKED_FUNDS',
   message: 'Locked funds.'
+}, {
+  name: 'LOCKED_ETH_FEE',
```
matiu: we can reuse: https://github.com/bitpay/bitcore/pull/2697/files#diff-e57a24c0f3a56c58bf7457e8e42b712cR56

packages/bitcore-wallet-client/src/lib/errors/spec.ts
gabrielbazan7: Yes, maybe but not sure. I think the error is slightly different. This message are for ERC20 tokens when you don't have enough fee in your ETH linked wallets.

PR #2705, opened 2020-02-03
ERC-20 bitcore-client

SethMcGuire:


micahriggan:
Based of off https://github.com/bitpay/bitcore/pull/2684

I would prefer to not have token addresses stored on bitcore-node wallets.

I would rather the wallet-token file store the token information on the bitcore-client wallet, and not make the http call to update the wallet.

packages/bitcore-client/bin/wallet-token
```diff
@@ -0,0 +1,33 @@
...
+  .version(require('../package.json').version)
+  .option('--name <name>', 'REQUIRED - Wallet name')
+  .option('--path [path]', 'optional - Custom wallet storage path')
+  .option('--contractAddress <contractAddress>', 'REQUIRED - Token contract address')
+  .parse(process.argv);
+
+const main = async () => {
+  const { name, path, contractAddress } = program;
+  if (!contractAddress) {
+
```
matiu: ? maybe something is missing here?

packages/bitcore-node/src/routes/api/wallet.ts
```diff
@@ -134,6 +134,9 @@ router.get('/:pubKey/transactions', Auth.authenticateMiddleware, async (req: Aut
 
 router.get('/:pubKey/balance', Auth.authenticateMiddleware, async (req: AuthenticatedRequest, res) => {
   let { chain, network } = req.params;
+  if (req.body.tokenContractAddress) {
```
micahriggan: I see why this line, but I think sending a body with a GET is probably bad practice

https://github.com/bitpay/bitcore/pull/2705/files#diff-ae61c7683311dfd83594b08ca457ae70R63
This method should not be sending a body, token address is meant to be a queryParam

client should have something like this in getBalance

```
if(tokenAddress) {
  url += `?tokenAddress=${tokenContractAddress}`
}
```
or

```
if(payload) {
  url += '?' + qs.stringify(payload)
}
```

packages/bitcore-client/bin/wallet-paypro
```diff
@@ -80,56 +81,68 @@ const main = async () => {
...
-  //  Sign tx
-  const password = await promptly.password('Wallet Password:');
-  wallet = await wallet.unlock(password);
-  const signedTx = await wallet.signTx({ tx: unsignedTx, passphrase: password });
-  const txid = cwc.Transactions.getHash({chain, tx: signedTx});
-  console.log(`signedRawTx: ${signedTx}`);
-  console.log(`txid: ${txid}`);
+  let unsignedTransactions = [];
+  let signedTransactions = [];
+  nonce--;
```
micahriggan: seems like the nonce incrementing should be at the end of the loop, so you don't have to -- and then ++

PR #2707, opened 2020-02-04
Fix toLegacyAddress() method wrong

vanphandinh:
Change livenet network configs.


matiu:
Could you explain the rationale of this?

vanphandinh:
> Could you explain the rationale of this?

The Legacy Address of BCH is same with BTC Address, right?
With old livenet configs, when import CashAddress, eg: 

const test = Address.fromString('bitcoincash:qqrxa0h9jqnc7v4wmj9ysetsp3y7w9l36u8gnnjulq')
console.log(test.toLegacyAddress())
// result: CH3uHwLiKNWmDqY3aSoNMsv5jPASQhnAsT, it's not a valid BTC address
// correct answer: 1b1itzeSKYEKhdcthUSnNJ47Fx2U8Zwwn
You can see here: https://explorer.bitcoin.com/bch/address/1b1itzeSKYEKhdcthUSnNJ47Fx2U8Zwwn

It's because livenet configs is not correct.
I take the correctly configs from https://github.com/bitpay/bitcore/blob/master/packages/bitcore-lib/lib/networks.js to fix this file.

vanphandinh:
@matiu pls review.

matiu:
Hi, thanks for the explanation.

Copay/Bitcore stack never used the same address for BTC and BCH to avoid user's confusion. Insead we used C/H for p2pkh and p2sh respectively, so "legacy address" refers to this notation. 

You can read  more about this here.

https://support.bitpay.com/hc/en-us/articles/115004671663-BitPay-s-Adopted-Conventions-for-Bitcoin-Cash-Addresses-URIs-and-Payment-Requests

If you need the "BTC" type of address, you can instantiate `Address` from `bitcore-lib`. 

thanks!

PR #2708, opened 2020-02-04
Mempool txs shouldn't be cached

micahriggan:
5252525 - -1 is greater than 100 for sure


matiu:
On BWS we only cached "hard confirmed TXs" (ie, for BTC, confirmations > 25 or something) 

You may want to do this, for enhanced security.

PR #2714, opened 2020-02-06
Fix(BWS) - opts.from gets set by BWS instead of Copay

justinkook:
fix address validation for ETH
throw Address Errors on validation

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -2360,7 +2360,8 @@ export class WalletService {
       const output = opts.outputs[i];
       output.valid = false;
 
-      const addrErr = ChainService.validateAddress(wallet, output.toAddress, opts);
+      let addrErr = ChainService.validateAddress(wallet, output.toAddress, opts);
+      addrErr = ChainService.validateAddress(wallet, opts.from, opts);
       if (addrErr) return addrErr;
 
       if (!checkRequired(output, ['toAddress', 'amount'])) {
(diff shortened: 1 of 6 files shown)


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2608,6 +2612,13 @@ export class WalletService {
 
             async.series(
               [
+                (next) => {
+                  this.getMainAddresses({reverse: true, limit: 1}, (err, mainAddr) => {
```
matiu: lets do this only on the chain that need it. Many add `needsFromAddress` flag in chain Service or something?

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2360,8 +2360,12 @@ export class WalletService {
       const output = opts.outputs[i];
       output.valid = false;
 
-      const addrErr = ChainService.validateAddress(wallet, output.toAddress, opts);
-      if (addrErr) return addrErr;
+      try {
+        ChainService.validateAddress(wallet, output.toAddress, opts);
+        ChainService.validateAddress(wallet, opts.from, opts);
```
matiu: only on certain chains

justinkook: fixed to use opts third param for ETH/XRP only

packages/bitcore-wallet-service/src/lib/server.ts
justinkook: added `isUTXOCoin` check

PR #2732, opened 2020-02-21
[REF] use 6 decimals for xrp

gabrielbazan7:


justinkook:
copay related MR

https://github.com/bitpay/copay/pull/10658

PR #2737, opened 2020-02-26
fixing rounding for wallet-send usage

sbhatnagar30:


packages/bitcore-client/bin/wallet-send
```diff
@@ -29,12 +29,12 @@ const main = async () => {
   try {
     wallet = await Wallet.loadWallet({ name, path });
     const chain = wallet.chain;
-    const { decimals } = currencies.data.find(element => element.code === chain);
+    const { decimals } = currencies.data.find(element => element.code === (token || chain));
```
micahriggan: can you do 
```
const currency = token  || chain;
...
element.code === currency
```

Other than that, looks good.

sbhatnagar30: done

PR #2739, opened 2020-02-26
WIP: Add P2WPKH & P2WSH address derivation for CWC and Bitcore-client

justinkook:


packages/bitcore-client/bin/multi-sig/wallet-derive-address
```diff
@@ -37,20 +38,22 @@ const parseKeys = keys => {
...
-function getAddress(publicKeyRing, addressIndex, isChange, m, network) {
+function getAddress(publicKeyRing, addressIndex, isChange, m, network, useBech32) {
   const publicKeys = publicKeyRing.map(function(key) {
     const xpub = new bitcoreLib.HDPublicKey(key);
     const changeNum = isChange ? 1 : 0;
     const path = `m/${changeNum}/${addressIndex}`;
     return xpub.derive(path).publicKey;
   });
-  return bitcoreLib.Address.createMultisig(publicKeys, m, network);
+  const nestedWitness = false;
```
andrewtoth: Why not always use nested witness? It is P2SH so the user will not see the difference, but will save on transaction fees because it's using segwit underneath.

packages/crypto-wallet-core/src/derivation/btc/index.ts
```diff
@@ -3,20 +3,24 @@ import { IDeriver } from '..';
...
-    return this.bitcoreLib.Address(xpub.derive(path).publicKey, network).toString();
+    return this.bitcoreLib.Address(xpub.derive(path).publicKey, network, type).toString();
   }
 
-  derivePrivateKey(network, xPriv, addressIndex, isChange) {
+  derivePrivateKey(network, xPriv, addressIndex, options) {
+    const { isChange, isBech32 } = options;
     const xpriv = new this.bitcoreLib.HDPrivateKey(xPriv, network);
     const changeNum = isChange ? 1 : 0;
+    const type = isBech32 ? 'witnesspubkeyhash' : 'pubkeyhash';
```
andrewtoth: If the address is change, it should always be bech32. The user will not see the address since it's internal, but will save on fees because it is using segwit.

Sayrix:
Hi, why the PR was closed, the code seems good to me

Issue #2735, opened 2020-02-23
Cannot pass ETH ropsten #599275 blocks


rlaace423:
Hello, I'm syncing ethereum ropsten with bitcore.
I'm using websocket, and I'm stucked at 599275th block.

It seems that this block has 1077 transaction, so It is bigger than websocket's maximum frame size. 

here's log looks like.
![image](https://user-images.githubusercontent.com/5516480/75118580-84340c80-56be-11ea-91a3-34de2d0e49f8.png)

And bitcore process closed.
![image](https://user-images.githubusercontent.com/5516480/75118589-96ae4600-56be-11ea-8095-b0f930eae81d.png)

This happens with my own parity node and also infura node. 

FYI, I updated bitcore because of this (https://github.com/bitpay/bitcore/issues/2726).
And I'm curious how can it overcome this block when I was trying with older version.

**Is there any way to overcome it?**

Always, thanks @micahriggan

rlaace423:
For this issue, I tried changing my config from websocket to rpc.

protocol : 'ws' --> 'http'
port : 8546 --> 8545 (I used default ports)

but it produced another errors like below.
![image](https://user-images.githubusercontent.com/5516480/75118720-d88bbc00-56bf-11ea-9ffe-8c518d434805.png)

And here's error when I ran with debug (npm run debug)
![image](https://user-images.githubusercontent.com/5516480/75118749-0ec93b80-56c0-11ea-8452-e983ea7e226d.png)

rlaace423:
I found reason. 

Because of websocket's default frame & message size, web3 can't call rpc "tract_block" or "get_block" when data is too large.
For this issue, we can pass **clientConfig** when we use Web3.providers.WebsocketProvider.

I'll pull-request about this issue. @micahriggan

PR #2744, opened 2020-03-03
Allow web3 provider options for bitcore.config.js

rlaace423:
There's web3 provider issue when rpc result is too big it fails to keep connection (especially **WS provider**)
For example, 

- a lot of transactions (ex: ropsten 599275th block)
- a lot of trace_block

For this issue, we can pass Web3 constructor's options when initiate Web3.
I edited small amount of codes which can allow web3 can have options.
Now we can append options to Web3 like below.
image

FYI, here's my original issue #2735 
And here's web3 issue about clientConfig https://github.com/ethereum/web3.js/issues/1217#issuecomment-425956862

Always, thank you @micahriggan  :)

### DIFF
--- a/packages/bitcore-node/src/modules/ethereum/api/csp.ts
+++ b/packages/bitcore-node/src/modules/ethereum/api/csp.ts
@@ -69,6 +69,7 @@ export class ETHStateProvider extends InternalStateProvider implements IChainSta
       const protocol = provider.protocol || 'http';
       const portString = provider.port || '8545';
       const connUrl = `${protocol}://${host}:${portString}`;
+      const providerOptions = provider.options || {}; 
       let ProviderType;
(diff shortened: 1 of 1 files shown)


micahriggan:
Is the number 1073741824 based off of what you needed for the ropsten block to be processed, or can a smaller number be used?

If this is bytes, then that'd be a gig right?

rlaace423:
> Is the number 1073741824 based off of what you needed for the ropsten block to be processed, or can a smaller number be used?
> 
> If this is bytes, then that'd be a gig right?

That number is just for my test :) maxReceivedFrameSize and max ReceivedMessageSize are based off bytes.
Much smaller number can be used.

micahriggan:
okay, great thanks!

PR #2746, opened 2020-03-04
Gas estimation collides with fee estimates

micahriggan:


justinkook:
Need to change the BWS endpoint to hit the new /gas endpoint

PR #2758, opened 2020-03-18
Wallet lite

sbhatnagar30:


packages/bitcore-client/bin/wallet-storage
```diff
@@ -0,0 +1,81 @@
...
+const { Wallet } = require('../ts_build/wallet');
+const { Storage } = require('../ts_build/storage');
+const { StreamUtil } = require('../ts_build/stream-util');
+
+program
+  .version(require('../package.json').version)
+  .option('--name [name]', 'REQUIRED - Wallet Name')
+  .option('--path [path]', 'optional - Custom wallet storage path')
+  .option('--storageType [storageType]', 'optional - name of the database to use (Mongo or Level)')
+  .option('--exp [exp]', 'export an existing wallet into an encrypted json')
```
micahriggan: I would prefer the full word for exp and imp. You can add a short tag, like below if you want it to be shorter.
```
'-e, --export'
'-i, --import'
```

packages/bitcore-client/src/wallet.ts
```diff
@@ -420,6 +434,9 @@ export class Wallet {
   }
 
   async nextAddressPair(withChangeAddress?: boolean) {
+    if (this.lite) {
+      throw new Error('Cannot derive private keys for lite wallet');
```
micahriggan: couldn't you just call nextAddressPairLite() here?

packages/bitcore-client/bin/wallet-derive
```diff
@@ -28,8 +28,12 @@ const main = async () => {
     let index = wallet.addressIndex || 0;
 
     while (unusedAddressCounter < gap) {
-      const address = await wallet.nextAddressPair(change);
-
+      let address;
+      if (wallet.lite) {
+        address = await wallet.nextAddressPairLite(change);
```
micahriggan: I think if you made nextAddressPair do this switch, then you wouldn't have to change this part.

PR #2759, opened 2020-03-19
Fix schnorr verify

bsnowden3:
A quick fix that removes object spread syntax, so that bwc can compile in copay.


frankja75:
Jesse Frank

frankja75:
K

PR #2778, opened 2020-04-03
[FIX]: fix to handle error controls for payproV2

rastajpa:


packages/bitcore-wallet-client/src/lib/payproV2.ts
```diff
@@ -76,7 +76,7 @@ export class PayProV2 {
         if (err) {
           if (res && res.statusCode !== 200) {
             // some know codes
-            if (res.statusCode == 400 || res.statusCode == 422) {
+            if ((res.statusCode == 400 || res.statusCode == 422) && res.body && res.body.msg) {
```
cmgustavo, line 79: I think we should return some generic error if any condition

PR #2795, opened 2020-04-15
Adding mempool age pruning, and stream refactor

micahriggan:


packages/bitcore-node/src/services/pruning.ts
```diff
@@ -34,45 +36,101 @@ export class PruningService {
...
+        .pipe(
+          new Transform({
+            objectMode: true,
+            transform: async (data: any, _, cb) => {
+              if (this.stopping) {
+                return cb(new Error('Stopping'));
+              }
+              const tx = data as ITransaction;
+              const outputs = await this.transactionModel.findAllRelatedOutputs(tx.txid);
+              const invalid = outputs.find(c => c.mintHeight >= 0 || c.spentHeight >= 0);
```
matiu, line 67: Is this something that could happen? Should we log this issue?

micahriggan: I think it should crash if that happens, and log, I'll add an error there

PR #2802, opened 2020-04-23
Adding logging for txid to prune

micahriggan:


packages/bitcore-node/src/services/pruning.ts
```diff
@@ -34,10 +35,15 @@ export class PruningService {
   }
 
   async detectAndClear() {
-    for (let chainNetwork of Config.chainNetworks()) {
-      const { chain, network } = chainNetwork;
-      await this.processOldMempoolTxs(chain, network, MEMPOOL_AGE);
-      await this.processAllInvalidTxs(chain, network);
+    if (CHAIN && NETWORK) {
```
matiu, line 38: it would be nice to check that CHAIN and NETWORK exists on Config.chainNetworks() to avoid silent errors.

matiu:
In BTC, unconfirmed chains are limited to 25 deep. In BCH to 50.

Maybe we should add a check on the generator, so if it reaches more that levels, it throws.

PR #2808, opened 2020-04-28
Ref/push notification chainservice

matiu:


packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
```diff
@@ -489,7 +487,7 @@ export class V8 {
 
     blockSocket.on('connect', () => {
       log.info(`Connected to block ${this.getConnectionInfo()}`);
-      blockSocket.emit('room', `/${this.coin.toUpperCase()}/${this.v8network}/inv`);
+      blockSocket.emit('room', `/${this.chain}/${this.v8network}/inv`);
```
micahriggan, line 490: this.chain is lowercased up above right? I don't think that'd be a valid room for bitcore-node

packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
```diff
@@ -122,7 +120,7 @@ export class V8 {
       name: wallet.id,
       pubKey: wallet.beAuthPublicKey2,
       network: this.v8network,
-      chain: this.coin
+      chain: this.coin // lowerCase?
```
micahriggan, line 123: I think this should be this.chain

matiu: I agree. but it seems we are sending it lowercase now, so I didnt wanted to modify it.

Is it possible we have something on bitcore-node what allows chain to be lowerCased on wallet registration?

micahriggan:
```
  let { chain, network } = req.params;
  let { name, pubKey, path, singleAddress } = req.body;
```

Looks like it's coming from the URL, and not the request body

matiu: ic, ok, ill change it anyways so it is clear.

PR #2813, opened 2020-05-04
[Insight] FEAT: Global search. Home Page

rastajpa:
- Global search added
- Home page refactored 
- Global search page
- Charts related code was removed
- Fix lint && prettier
- Removed unused classes


rastajpa:
> ACK! Just one change I would make is when clicking 'Insight' on Navbar have it redirect to Global Blockchains Home Page

Done!

PR #2826, opened 2020-05-18
Refactor Key.ts

matiu:


bsnowden3 (dismissed):

bsnowden3:
ACK

gabrielbazan7:
ACK !

sondreb:
This recent refactoring of Key.ts results in an invalid type definition:

```
key.d.ts, line: 3
Invalid character.
```

image

This is related to fields on the Key class being named with a prefix of #. What is the reason for using # prefix? Never seen that used in TypeScript before.

image

PR #2842, opened 2020-06-02
[FEAT] ETH multisig

gabrielbazan7:
ETH MULTISIG changes for BWC/BWS/BN/CWC
Related Copay PR: 
https://github.com/bitpay/copay/pull/10960

Node configuration ( kovan testnet): 

./openethereum --chain kovan --base-path '/Users/gabrielbazan/Documents/blockchains/ethereum/prod' --db-compaction ssd --cache-size 8192 --jsonrpc-cors=all --ws-origins=all --ws-apis=all --tracing on

Needs `trace ` on for handling internal transactions.


packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -134,9 +165,12 @@ export class ETHStateProvider extends InternalStateProvider implements IChainSta
     const { chain, network, address } = params;
     const { web3 } = await this.getWeb3(network);
     const tokenAddress = params.args && params.args.tokenAddress;
+    const multisigContractAddress = params.args && params.args.multisigContractAddress;
     const addressLower = address.toLowerCase();
     const cacheKey = tokenAddress
       ? `getBalanceForAddress-${chain}-${network}-${addressLower}-${tokenAddress.toLowerCase()}`
+      : multisigContractAddress
+      ? `getBalanceForAddress-${chain}-${network}-${multisigContractAddress.toLowerCase()}`
```
micahriggan, line 173: Thinking we might not need this distinction. On the BWS side we should pass in the multisigContractAddress as the address

packages/bitcore-node/src/modules/ethereum/api/transform.ts
```diff
@@ -14,12 +14,13 @@ export class EthListTransactionsStream extends Transform {
...
     });
-    if (sending > 0) {
+
+    if (sending > 0 || transaction.from === this.multisigContractAddress) {
       const sendingToOurself = await WalletAddressStorage.collection.countDocuments({
         wallet: this.wallet._id,
         address: transaction.to
       });
-      if (!sendingToOurself) {
+      if (!sendingToOurself || transaction.to !== this.multisigContractAddress) {
```
micahriggan, line 23: If the address is the multisigContractAddress, this wouldn't need to be added.

packages/bitcore-node/src/modules/ethereum/models/transaction.ts
```diff
@@ -108,14 +124,19 @@ export class EthTransactionModel extends BaseTransaction<IEthTransaction> {
...
+      let batch = new Array<{ multisigContractAdress?: string; tokenAddress?: string; address: string }>();
       const { chain, network } = op.updateOne.filter;
       const { from, to, abiType } = op.updateOne.update.$set;
       batch = batch.concat([{ address: from }, { address: to }]);
-      if (abiType && abiType.params.length) {
+      if (abiType && abiType.type === 'ERC20' && abiType.params.length) {
         batch.push({ address: from, tokenAddress: to });
         batch.push({ address: abiType.params[0].value, tokenAddress: to });
       }
+
```
micahriggan, line 135: we should also push a batch for the internal transfers here

packages/bitcore-node/src/modules/ethereum/api/internalTxTransform.ts
```diff
@@ -0,0 +1,33 @@
...
+import { IEthTransaction } from '../types';
+
+export class InternalTxRelatedFilterTransform extends Transform {
+  constructor(private web3: Web3, private walletId) {
+    super({ objectMode: true });
+  }
+
+  async _transform(tx: MongoBound<IEthTransaction>, _, done) {
+    if (tx.internal && tx.internal.length > 0) {
+      const walletAddresses = await WalletAddressStorage.collection
```
micahriggan, line 14: should pull getting the wallet addresses into a method, and have it cache them, so you don't have to hit the db per tx


```
getWalletAddresses() {
 if(!this.walletAddresses.length){
    this.walletAddresses = await find()
  }
  return this.walletAddresses
}
```

gabrielbazan7: Done. Thanks !

packages/bitcore-node/src/modules/ethereum/models/transaction.ts
gabrielbazan7: Done. Thanks !

packages/bitcore-node/src/modules/ethereum/api/csp.ts
gabrielbazan7: We are treating multisig like erc20 wallets, so there is not wallet address data stored. 
```
let addresses = await this.getWalletAddresses(params.wallet._id);
```
^ since `params.wallet._id` is the id from the linked eth wallet to that multisig contract, this will get the linked wallet addresses. That's the reason why I need the `multisigContractAddress` variable in the params to get the correct balance.

packages/bitcore-node/src/modules/ethereum/api/transform.ts
gabrielbazan7: Same here... there is no wallet data stored for multisig. So, 
```
const sendingToOurself = await WalletAddressStorage.collection.countDocuments({
        wallet: this.wallet._id,
        address: transaction.to
      });
```
will not find any coincidence.

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -1089,6 +1089,8 @@ export class API extends EventEmitter {
   // * @param {Boolean} opts.twoStep[=false] - Optional: use 2-step balance computation for improved performance
   // * @param {Boolean} opts.includeExtendedInfo (optional: query extended status)
   // * @param {String} opts.tokenAddress (optional: ERC20 Token Contract Address)
+  // * @param {String} opts.multisigContractAddress (optional: MULTISIG ETH Contract Address)
+  // * @param {String} opts.network (optional: MULTISIG ETH Contract Address network)
```
matiu, line 1093: why we need network here? Should the multisig contract use the same network as the original wallet?

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -1111,6 +1113,11 @@ export class API extends EventEmitter {
       qs.push('tokenAddress=' + opts.tokenAddress);
     }
 
+    if (opts.multisigContractAddress) {
+      qs.push('multisigContractAddress=' + opts.multisigContractAddress);
+      qs.push('network=' + opts.network);
```
matiu, line 1118: (ie here:  change opts.network with:  this.credentials.network)

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -2116,6 +2129,36 @@ export class API extends EventEmitter {
     });
   }
 
+  // /**
+  // * Returns contract instantiation info.
```
matiu, line 2133: Could you explain more what information will be returned in this function?

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -2116,6 +2129,36 @@ export class API extends EventEmitter {
...
+  // * @param {string} opts.sender - sender eth wallet address
+  // * @param {string} opts.network - the network for this wallet
+  // * @param {string} opts.coin - the coin for this wallet
```
matiu, line 2136: isn't coin always ETH here?

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -2116,6 +2129,36 @@ export class API extends EventEmitter {
...
+  getMultisigContractInstantiationInfo(opts, cb) {
+    var url = '/v1/ethmultisig/';
+    this.request.post(url, opts, (err, contractInstantiationInfo) => {
+      if (err) return cb(err);
+      return cb(null, contractInstantiationInfo);
+    });
+  }
+
+  // /**
+  // * Returns contract info.
```
matiu, line 2148: please provide more details about the information that will be returned.

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -2116,6 +2129,36 @@ export class API extends EventEmitter {
...
+  // * @param {string} opts.multisigContractAddress - multisig contract address
+  // * @param {string} opts.network - the network for this wallet
+  // * @param {string} opts.coin - the coin for this wallet
```
matiu, line 2151: same as above, isn't coin always eth?

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -2386,6 +2429,22 @@ export class API extends EventEmitter {
               clients.push(tokenClient);
             });
           }
+          // Eth wallet with mulsig wallets?
+          const multisigEthInfo = status.preferences.multisigEthInfo;
+          if (!_.isEmpty(multisigEthInfo)) {
+            _.each(multisigEthInfo, info => {
+              log.info('Importing multisig wallet');
```
matiu, line 2436: add ` info.multisigContractAddress` to the log

packages/bitcore-wallet-service/src/lib/model/txproposal.ts
```diff
@@ -255,6 +259,8 @@ export class TxProposal {
     x.gasLimit = obj.gasLimit; // Backward compatibility for BWC <= 8.9.0
     x.data = obj.data; // Backward compatibility for BWC <= 8.9.0
     x.tokenAddress = obj.tokenAddress;
+    x.multisigContractAddress = obj.multisigContractAddress;
+    x.multisigTxId = obj.multisigTxId;
```
matiu, line 263: this is the TXID that created the contract? Why is needed?

packages/bitcore-wallet-service/src/lib/storage.ts
```diff
@@ -76,6 +76,10 @@ export class Storage {
     db.collection(collections.TXS).createIndex({
       txid: 1
     });
+    db.collection(collections.TXS).createIndex({
+      multisigContractAddress: 1,
```
matiu, line 80: this will be VERY slow to create, can we find an alternative way?

matiu:
I have reviewed bitcore-wallet-client / bitcore-wallet-server / cryto-wallet-client parts, great job! Left a few comment, the mayor blocker is:
```
    db.collection(collections.TXS).createIndex({
```
I dont think that is doable on production

packages/bitcore-wallet-service/src/lib/model/txproposal.ts
gabrielbazan7: That's the transaction id on the contract (  0, 1, 2, 3, etc ). Different from the transaction hash. Two transactions with different transactions hash have the same transaction id when they are trying to execute the contract. If I have a transaction that still need confirmations and I want to confirm one of them, I need the id to implement the confirm method.
```
  confirmTransactionEncodeData(params: { multisigContractAddress: string; transactionId: number }) {
    const { multisigContractAddress, transactionId } = params;
    const data = this.getMultisigContract(multisigContractAddress)
      .methods.confirmTransaction(transactionId)
      .encodeABI();
    return data;
  }
```

packages/bitcore-wallet-service/src/lib/storage.ts
gabrielbazan7: Done. I remove it. Not needed.

packages/bitcore-wallet-client/src/lib/api.ts
gabrielbazan7: Done. Thanks !

packages/bitcore-wallet-client/src/lib/api.ts
gabrielbazan7: Done. I use 'eth' for coin now

packages/bitcore-wallet-client/src/lib/api.ts
gabrielbazan7: Done. Thanks

packages/bitcore-node/src/modules/ethereum/models/transaction.ts
```diff
@@ -59,6 +68,13 @@ export class EthTransactionModel extends BaseTransaction<IEthTransaction> {
         partialFilterExpression: { chain: 'ETH', 'abiType.type': 'ERC20', 'abiType.name': 'transfer' }
       }
     );
+    this.collection.createIndex(
+      { chain: 1, network: 1, 'internal.action.to': 1 },
+      {
+        background: true,
```
micahriggan, line 74: needs sparse, don't need partialFilter

gabrielbazan7: Done. Thanks!

PR #2869, opened 2020-06-15
Adding verification worker to ripple to repair missing blocks

micahriggan:


packages/bitcore-node/test/verification/db-repair.ts
```diff
@@ -27,160 +27,164 @@ import { Verification } from '../../src/services/verification';
...
-          const tx = data.payload.tx;
-          const dupeTxs = await TransactionStorage.collection
-            .find({ chain: tx.chain, network: tx.network, txid: tx.txid })
-            .sort({ blockHeight: -1 })
-            .toArray();
-
-          if (dupeTxs.length < 2) {
-            console.log('No action required.', dupeTxs.length, 'transaction');
-            return;
+    try {
```
micahriggan, line 30: Added a try catch, diff looks worse than it is

PR #2871, opened 2020-06-17
BitcoreNode(XRP): Rewrite the transform to be more accurate

micahriggan:


packages/bitcore-node/src/modules/ripple/api/wallet-tx-transform.ts
```diff
@@ -36,75 +36,73 @@ export class RippleDbWalletTransactions extends Transform {
...
+            satoshis: -1 * Math.abs(output.value),
+            height: tx.blockHeight,
+            address: output.address,
+            outputIndex: output.mintIndex,
+            blockTime: tx.blockTimeNormalized
+          }) + '\n'
+        );
+      });
+
+      const moves = outputs.filter(o => o.address != tx.from && relevantAddresses.includes(o.address));
```
jguthrie7, line 58: Should do something to make sure that if an output registers as a `move` it won't also be a `send`

packages/bitcore-node/src/modules/ripple/api/wallet-tx-transform.ts
```diff
@@ -33,78 +33,76 @@ export class RippleDbWalletTransactions extends Transform {
...
-              outputIndex: output.mintIndex,
-              blockTime: tx.blockTimeNormalized
-            }) + '\n'
-          );
-        }
+    const walletAddressesObjs = await this.getAddresses();
+    const walletAddresses = walletAddressesObjs.map(w => w.address);
+    let sending = walletAddresses.includes(tx.from);
+    if (sending) {
+      const sends = outputs.filter(o => !walletAddresses.includes(o.address));
```
matiu, line 40: What you think about defining sends/moves/receives on a single loop on outputs, so we can optimize the flow a bit?

micahriggan: For this diff, it was originally a single loop, but the logic was confusing to think through.

I can rewrite it as a single loop again, and get rid of the filters and forEach

packages/bitcore-node/src/modules/ripple/api/wallet-tx-transform.ts
```diff
@@ -18,6 +18,25 @@ export class RippleDbWalletTransactions extends Transform {
...
+      JSON.stringify({
+        id: tx.txid,
+        txid: tx.txid,
+        fee: tx.fee,
+        size: 0,
+        category,
+        satoshis: value,
+        height: tx.blockHeight,
+        blockTime: tx.blockTimeNormalized,
+        ...(o && {
```
matiu, line 32: nice ;)

PR #2873, opened 2020-06-18
BitcoreNode(ETH): Add receipt to wallet tx history

micahriggan:
* Populate receipt on wallet txs that need it
 * Add gasPrice, gasLimit and receipt to the wallet tx transform


packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -176,19 +189,13 @@ export class ETHStateProvider extends InternalStateProvider implements IChainSta
...
           confirmations = tipHeight - found.blockHeight + 1;
         }
-        if (!found.receipt) {
-          const receipt = await this.getReceipt(network, found.txid);
-          if (receipt) {
-            await EthTransactionStorage.collection.updateOne({ _id: found._id }, { $set: { receipt } });
-            found.receipt = receipt;
-          }
-        }
+        found = await this.populateReceipt(found);
```
matiu, line 198: Do we need to return the `tx` here? Wouldn't be more clear to do just do:
```
   await this.populateReceipt(found);
```

micahriggan: That would work as well. I do this to signal that found is being changed.

kaycoinsUSA: Great work this is going to make everyone happy. If you are working diligently on any more projects I am doing please speak up about sponsoring. I don't know how that originates but I need as many of you on my team as possible. Kaycoins@outlook.com.

PR #2878, opened 2020-06-24
Chore/update mongo drive

matiu:
- updates mongoDriver
- rm morgan and npmlog + replace it for winston
- fix issue with isSupportStaff


packages/bitcore-wallet-service/src/lib/updatestats.ts
```diff
@@ -27,12 +28,13 @@ export class UpdateStats {
     }
     uri = uri + 'readPreference=secondaryPreferred';
     console.log('Connected to ', uri);
-    mongodb.MongoClient.connect(uri, (err, db) => {
+    mongodb.MongoClient.connect(uri, { useUnifiedTopology: true }, (err, client) => {
```
micahriggan, line 31: https://mongodb.github.io/node-mongodb-native/3.3/reference/unified-topology/

Nothing to change, just found out the concept of calling connect will go away soon. Was trying to research that flag

packages/bitcore-wallet-service/src/lib/updatestats.ts
```diff
@@ -202,7 +202,7 @@ export class UpdateStats {
     // Grab last run
     let cursor = await this.db
       .collection(coll)
-      .find({}, { _id: true })
+      .find({}) // , { _id: true }})  // not working on new mongo driver
```
micahriggan, line 205: I think _id comes back always on project

packages/bitcore-wallet-service/src/lib/updatestats.ts
matiu: yes, thanks for checking, will change that next.

PR #2888, opened 2020-07-03
[Fix] BCH legacy/base58 address version prefix

sagech:
The BCH networks list in bitcore-lib-cash uses a different version prefix for pubkeyhash and scripthash for mainnet. This results in incorrect addresses being generated for pubkeyhash and scripthash type addresses if we use the `Address.toLegacyAddress()` method

In this PR, the prefixes have been changed to follow the spec. Unit tests have also been updated to reflect the change in addresses generated. Reference:

- [BCHN](https://gitlab.com/bitcoin-cash-node/bitcoin-cash-node/-/blob/master/src/chainparams.cpp#L172)
- [Bitcoin ABC](https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/src/chainparams.cpp#L227)
- [Bitcoin Unlimited](https://github.com/BitcoinUnlimited/BitcoinUnlimited/blob/release/src/chainparams.cpp#L171)


matiu:
Copay used the prefix "C" and "H" for legacy addresses, back in the day, we never used the same addresses in BTC an BCH,  we wont change that to support our old clients.

You can access "BTC" style addresses simply by creating a new `Address` object from the library `bitcore-lib` using:

const bch = require('bitcore-lib-cash');
const btc = require('bitcore-lib');

const bitcoinCastAddress =  bch.Address.fromString(someParam);
const address = btc.Address.fromObject(bitcoinCashAddress.toObject());

matiu:
Reference: https://support.bitpay.com/hc/en-us/articles/115004671663-BitPay-s-Adopted-Conventions-for-Bitcoin-Cash-Addresses-URIs-and-Payment-Requests

PR #2889, opened 2020-07-06
coalesce CSP promises

nitsujlangston:


micahriggan:
Should include method name in the hashing, params may be the same across methods

nitsujlangston:
> Should include method name in the hashing, params may be the same across methods

done

matiu:
Code looks fine to me, but why we needed to add the caller to the key? Shouldn't we coalesce the same query even if it is called from different places?

PR #2893, opened 2020-07-13
[FIX] ERC20 history fix

gabrielbazan7:


bsnowden3:
The tests will pass after fixing Common = require('../common') to require('./common') in packages/bitcore-wallet-service/lib/storage.ts

bsnowden3:
Created a PR here for the fix. https://github.com/bitpay/bitcore/pull/2895

PR #2900, opened 2020-07-21
fix bwc base url

bsnowden3:


packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -30,7 +30,7 @@ var querystring = require('querystring');
 var log = require('./log');
 const Errors = require('./errors');
 
-var BASE_URL = 'http://localhost:3232/bws/api';
+var BASE_URL = 'https://bws.bitpay.com/bws/api';
```
matiu, line 33: Probably this is not needed, since you can pass `opts.baseUrl` on the constructor. Maybe is be to keep the default pointing to the local dev instance?

bsnowden3:
Yes thanks, thought I had closed it.

PR #2908, opened 2020-07-30
[FIX] ERC20 update wallet & stream wallet txs

gabrielbazan7:


packages/bitcore-node/src/modules/ethereum/models/transaction.ts
```diff
@@ -62,10 +62,10 @@ export class EthTransactionModel extends BaseTransaction<IEthTransaction> {
     this.collection.createIndex({ chain: 1, network: 1, from: 1 }, { background: true, sparse: true });
     this.collection.createIndex({ chain: 1, network: 1, from: 1, nonce: 1 }, { background: true, sparse: true });
     this.collection.createIndex(
-      { chain: 1, network: 1, 'abiType.params.0.value': 1, blockTimeNormalized: 1 },
```
micahriggan, line 65: Is this required? This index is already created, so changing this would create new indexes, and the previous indexes would be dropped.

packages/bitcore-node/src/modules/ethereum/api/csp.ts
```diff
@@ -494,7 +494,12 @@ export class ETHStateProvider extends InternalStateProvider implements IChainSta
           $or: [
             { chain, network, from: { $in: addressBatch } },
             { chain, network, to: { $in: addressBatch } },
-            { chain, network, 'internal.action.to': { $in: addressBatch } }
+            { chain, network, 'internal.action.to': { $in: addressBatch } },
+            {
+              chain,
+              network,
+              'abiType.params.0.value': { $in: addressBatch.map(address => address.toLowerCase()) }
```
micahriggan, line 501: which type of tx is this? Multi-sig, or ERC20?

gabrielbazan7: ERC20

micahriggan: This will tag any tx with an address as the first param, across all ABI types and all functions. Is that ok?

I'm getting the sense we should rethink ETH txs a bit, since each tx can have side-effects that affect other addresses, and that address could be in a different param index

It might be time to draw up a coin equivalent for ETH or something similar to track all the side-effects

Then a tx can be tagged with an aggregation of the side-effects wallets, just like bitcoin tx's are tagged with the aggregate of coin wallets

matiu:
needs rebase.

PR #2920, opened 2020-08-12
Feat/public cache

matiu:


packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -243,8 +243,16 @@ export class ExpressApp {
...
+    // Case: "▶Cache an asset with revalidation, but allow stale responses if origin server is unreachable"
+    function SetPublicCache(res: express.Response, seconds: number) {
+      res.setHeader('Cache-Control', `public, max-age=${seconds}, stale-if-error=${10 * seconds}`);
+    }
+
     // retrieve latest version of copay
     router.get('/latest-version', async (req, res) => {
+      SetPublicCache(res, 10 * ONE_MINUTE);
       try {
         res.setHeader('User-Agent', 'copay');
```
cmgustavo, line 257: This value will replace the previous one: https://nodejs.org/api/http.html#http_request_setheader_name_value

PR #2939, opened 2020-08-21
removing params from wallet object to save

sbhatnagar30:


packages/bitcore-client/src/wallet.ts
```diff
@@ -122,7 +122,12 @@ export class Wallet {
     if (alreadyExists) {
       throw new Error('Wallet already exists');
     }
-    const wallet = Object.assign(params, {
+    const wallet = Object.assign( {
```
micahriggan, line 125: We could remove Object.assign completely

sbhatnagar30: sorry I didn't see this until too late, maybe I can remove it in the future

PR #2940, opened 2020-08-25
Feat(PayPro) - Add address to selectPaymentOptions param

justinkook:
Reference to Copay MR:

https://github.com/bitpay/copay/pull/11095


packages/bitcore-wallet-client/src/lib/payproV2.ts
```diff
@@ -183,7 +183,8 @@ export class PayProV2 {
       },
       args: JSON.stringify({
         chain,
-        currency
+        currency,
+        address
```
micahriggan, line 187: shouldn't this be payload?

PR #2956, opened 2020-09-11
[FEAT] [BWS]: Clean fiat_rates2 periodically

rastajpa:


packages/bitcore-wallet-service/src/lib/cleanfiatrates.ts
```diff
@@ -0,0 +1,151 @@
...
+          console.log(`\tFiltering times for ${key.toUpperCase()}`);
+          let prevTime = null;
+          let isSameDay, isSameHour;
+
+          _.forEach(tsGroup, ts => {
+            if (prevTime) {
+              isSameDay = moment(prevTime).isSame(moment(ts), 'day');
+              isSameHour = moment(prevTime).isSame(moment(ts), 'hour');
+
+              if (!isSameDay || (isSameDay && !isSameHour)) {
```
matiu, line 120: if `results` are order in time `!isSameHour` should be enough, right?

packages/bitcore-wallet-service/src/lib/cleanfiatrates.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        // keep one date every hour for each coin
+        _.forEach(tsGruopedByCoin, (tsGroup, key) => {
+          console.log(`\tFiltering times for ${key.toUpperCase()}`);
+          let prevTime = null;
+          let isSameDay, isSameHour;
+
+          _.forEach(tsGroup, ts => {
+            if (prevTime) {
+              isSameDay = moment(prevTime).isSame(moment(ts), 'day');
```
matiu, line 117: maybe we should add `if (prevTime > ts) =>` ERROR, results not in order`

matiu: or if (prevTime > ts + 60*1000) (give it 60 secs JIC) => error

packages/bitcore-wallet-service/src/lib/cleanfiatrates.ts
rastajpa: yes! you're right!

packages/bitcore-wallet-service/src/lib/cleanfiatrates.ts
rastajpa: sounds good to me!

PR #2964, opened 2020-09-23
WIP bitcore docker

nitsujlangston:
`docker run bitcore node`
`docker run bitcore bws`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,7 @@ ADD . .
 RUN npm run compile
 
 EXPOSE 3000
+EXPOSE 3232
 
 ENTRYPOINT [ "npm", "run" ]
 CMD [ "node" ]
--- a/Dockerfile-ci
+++ b/Dockerfile-ci
@@ -1,5 +1,26 @@
 FROM node:10
 
+# Install Chrome
+
+RUN echo 'deb http://dl.google.com/linux/chrome/deb/ stable main' > /etc/apt/sources.list.d/chrome.list
+
+RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
+
+RUN set -x \
+    && apt-get update \
+    && apt-get install -y \
+        google-chrome-stable
+
+ENV CHROME_BIN /usr/bin/google-chrome
+
+# Log versions
+
+RUN set -x \
+    && node -v \
+    && npm -v \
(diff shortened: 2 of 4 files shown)


micahriggan:
Why do there need to be two Dockerfiles? The current one is capable of running the CI and bitcore node

nitsujlangston:
> Why do there need to be two Dockerfiles? The current one is capable of running the CI and bitcore node

@micahriggan !! Just a WIP, working on getting bitcore api/sync and bws api/other services all runnable from single container. Broke out the ci one to get the chrome stuff out of there, but open to suggestions on how to extend them from single dockerfile

nitsujlangston:
End goal being an image which can be run with different commands and arguments to create an entire kubernetes set up running all of it

micahriggan:
Nice!

https://github.com/bitpay/bitcore/pull/2706/files

I messed around a lil bit with that, I made additional Docker files for the specific services. 

So bitcore-node-api just extends the current Dockerfile

The chrome stuff was only in there for Insight afaik, it might not be needed anymore.

The thing I couldn't figure out was how to test the kubernetes files without having them mutate with machine specific info. Let me know if you get that figured out

sondreb:
> Nice!
> 
> https://github.com/bitpay/bitcore/pull/2706/files
> 
> I messed around a lil bit with that, I made additional Docker files for the specific services.
> 
> So bitcore-node-api just extends the current Dockerfile
> 
> The chrome stuff was only in there for Insight afaik, it might not be needed anymore.
> 
> The thing I couldn't figure out was how to test the kubernetes files without having them mutate with machine specific info. Let me know if you get that figured out

Is there a point in making additional dockerfiles for the different modues, e.g. you made "Dockerfile-bitcore-node-api" which does:

```
CMD ["node", "/bitcore/packages/bitcore-node/build/src/workers/api.js"]
```

If the main dockerfile only contains at the bottom:

```
ENTRYPOINT [ "npm", "run" ]
```

Then you can run the different npm tasks with command in your compose files for Kubernetes/Docker:

```
command: ["node"]
```

micahriggan:
@sondreb I think what you're suggesting makes more sense than the way I did it with multiple files.

I just didn't know how to pass something into the entry point like that.

 I think at the time I was thinking you could publish each one so people could run the specific service they wanted, but seems like that could be accomplished with what you're suggesting.

PR #2969, opened 2020-10-06
Feature/payId

kajoseph:


packages/bitcore-payid/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+class PayId {
+  constructor() {}
+
+  /**
+   * Sign a payId address with the given identityKey
+   * @param {string} payId
+   * @param {string} address Address to be signed
+   * @param {string} currency Currency ticker (e.g. BTC, ETH, XRP)
+   * @param {string} identityKey The key to be used for signing
```
matiu, line 16: maybe specify what type of key is expected? (ie:  ecdsa 32byte priv key) and an example of input

packages/bitcore-payid/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+} from '@payid-org/utils';
+import { JWS } from 'jose';
+import { IVerifyPayId } from './index.d';
+
+class PayId {
+  constructor() {}
+
+  /**
+   * Sign a payId address with the given identityKey
+   * @param {string} payId
```
matiu, line 13: would be nice to expecify examples for all parameters.

packages/bitcore-payid/src/index.ts
```diff
@@ -0,0 +1,56 @@
+import {
+  AddressDetailsType,
+  verifySignedAddress,
+} from '@payid-org/utils';
+import { JWS } from 'jose';
```
matiu, line 5: jose? Do we really need that new dep? Can't we create/verify the signature with bitcore-lib?

matiu (changes requested):

matiu:
great job on the first commit, loved the tests. Left some minor comments.

packages/bitcore-payid/src/index.ts
kajoseph: The jose lib is a pretty standard one across languages for cross-platform and cross-algorithm key conversion. I used it here for key conversion for the general case when someone might provide their own signing key and not one generated by bitcore-lib. However, it's not needed for verification which is what this MR pertains to and might even be over-kill for signing. I'm not 100% sure that it's needed just yet. Removed for now.

packages/bitcore-payid/src/index.ts
kajoseph: Done

packages/bitcore-payid/src/index.ts
kajoseph: Done, though let me know if you don't like what I added. PayId tries to be key agnostic, so it should be able to accept any key string.

packages/bitcore-payid/README.md
```diff
@@ -0,0 +1,22 @@
+# Bitcore PayID
+
```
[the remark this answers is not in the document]
kajoseph: Done. LMK if you'd like it to be more specific

packages/bitcore-payid/src/index.ts
kajoseph: I ended up leaving the jose dep in due to PayId's integral dependence on it. I'm still investigating ways to build the objects I need without relying on a whole package, though. You can see in the tests how I'm using it to transform a key to the JWK format that PayId expects.

As for creating and verifying with bitcore-lib, it's non-trivial if we want it to be interoperable with other PayId instances.

matiu (approved):

PR #2975, opened 2020-10-08
PayId signing implementation

kajoseph:
This a continuation of https://github.com/bitpay/bitcore/pull/2969

What was done:
- Implemented the `sign` function
- Refactored to remove `jose` dependency
- Added tests


packages/bitcore-payid/src/index.ts
```diff
@@ -1,24 +1,42 @@
...
-  sign(payId: string, address: string, currency: string, identityKey: string): GeneralJWS {
-    // TODO
-    return {} as GeneralJWS;
+  sign(payId: string, address: string, currency: string, identityKey: string | Buffer): GeneralJWS {
+    let jwk = this._convertIdentityKeyToJWK(identityKey);
+
+    const signingParams = new IdentityKeySigningParams(jwk, getDefaultAlgorithm(jwk));
+    const addy = {
+      paymentNetwork: currency,
+      environment: 'mainnet',
```
matiu, line 31: maybe add an optional param 'environment' which defaults to `mainnet`?  so we have the option in the future.

packages/bitcore-payid/test/integration/index.spec.ts
```diff
@@ -118,6 +215,133 @@ describe('PayId', () => {
...
+      const verified = PayId.verify(payId, signed);
+      expect(verified).be.true;
+    });
+
+    it('should work with Bitcore non-HD key', () => {
+      const signed = PayId.sign(payId, addressBTC.addressDetails.address, 'BTC', keys.bitcore.toString());
+      expect(signed).to.exist;
+
+      const verified = PayId.verify(payId, signed);
+      expect(verified).be.true;
```
matiu, line 234: what about adding a tests to PayId.verify with a different payId => verify should by false; with a different signature => verify should be false.

packages/bitcore-payid/src/index.ts
kajoseph: Done

packages/bitcore-payid/test/integration/index.spec.ts
kajoseph: Good idea

kajoseph: Done

PR #2978, opened 2020-10-12
skip prettier for now on bwc

matiu:


micahriggan:
https://palantir.github.io/tslint/rules/trailing-comma/

[It does look like bwc uses that rule](https://github.com/bitpay/bitcore/blob/f4bbc1fec0c7b0bf30af8c0d706e31e3be07aef2/packages/bitcore-wallet-client/tslint.json#L41)

Maybe adding `"esSpecCompliant": true` would be helpful

matiu:
actually, this is because prettier (even the latest version) does not seem to support 
"hard private" members

https://www.typescriptlang.org/docs/handbook/classes.html#ecmascript-private-fields

micahriggan:
Neat, didn't know about #private

PR #2989, opened 2020-10-21
Update check state messages

bsnowden3:


matiu (approved):

matiu:
Thanks for this! This will be very useful for debugging.

PR #2996, opened 2020-10-27
Ref/push Upgrade notification to httpv1 migr

bsnowden3:


packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
```diff
@@ -169,24 +171,26 @@ export class PushNotificationsService {
...
-                        data: {
-                          walletId: sjcl.codec.hex.fromBits(sjcl.hash.sha256.hash(notification.walletId)),
-                          tokenAddress,
-                          multisigContractAddress,
-                          copayerId: sjcl.codec.hex.fromBits(sjcl.hash.sha256.hash(recipient.copayerId)),
-                          title: content.plain.subject,
-                          body: content.plain.body,
-                          notification_type: notification.type
+                        message: {
+                          topic: sub.token,
```
matiu, line 175: So, the old `to` is now `topic`? Is that correct?

packages/bitcore-wallet-service/src/config.ts
```diff
@@ -72,7 +72,7 @@ module.exports = {
     defaultLanguage: 'en',
     defaultUnit: 'btc',
     subjectPrefix: '',
-    pushServerUrl: 'https://fcm.googleapis.com/fcm',
+    pushServerUrl: 'https://fcm.googleapis.com/v1/projects/bitpay-wallet/messages:send',
     authorizationKey: 'You_have_to_put_something_here'
```
matiu, line 76: `authorizationKey `  is now deprecated right? Can you remove this and move the fcm_google_credentials here?

matiu (approved):

PR #3000, opened 2020-10-29
[FIX] for fee and amount checks

gabrielbazan7:
related https://github.com/bitpay/copay/pull/11212

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
+++ b/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
@@ -284,17 +284,19 @@ export class EthChain implements IChain {
 
         /* If its paypro its an already created ERC20 transaction and we need to get the actual invoice value from the data
         invoice outputs example:
-        "outputs":[
-          {"amount":0,
-          "toAddress":"0x44d69d16C711BF966E3d00A46f96e02D16BDdf1f",
-          "message":null,
-          "data":"...",
-          "gasLimit":29041},
-          {"amount":0,
-          "toAddress":"0xc27eD3DF0DE776246cdAD5a052A9982473FceaB8",
-          "message":null,
-          "data":"...",
-          "gasLimit":200000
+        "outputs":[{
+            "amount":0,
+            "toAddress":"0x44d69d16C711BF966E3d00A46f96e02D16BDdf1f",
+            "message":null,
+            "data":"...",
+            "gasLimit":29041
+          },
+          {
+            "amount":0,
+            "toAddress":"0xc27eD3DF0DE776246cdAD5a052A9982473FceaB8",
+            "message":null,
+            "data":"...",
+            "gasLimit":200000
         }]
         */
         const txpTotalAmount =
--- a/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
+++ b/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
@@ -251,7 +251,27 @@ export class EthChain implements IChain {
 
         const getInvoiceValue = txp => {
           let totalAmount;
+
+          /* invoice outputs data example:
+          abiDecoder.decodeMethod(txp.outputs[0].data)
+          { name: 'approve',
+            params:
+            [ { name: '_spender',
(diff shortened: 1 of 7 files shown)


cmgustavo (dismissed):
ACK!

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -225,14 +245,30 @@ export class EthChain implements IChain {
...
+              totalAmount = decodedData.params[0].value;
+            }
+          });
+          return totalAmount;
+        };
+
         const { totalAmount, availableAmount } = balance;
-        if (totalAmount < txp.getTotalAmount()) {
+        const txpTotalAmount =
+          (opts.multisigContractAddress || opts.tokenAddress) && txp.payProUrl
```
matiu, line 261: why this is only when there is a payProUrl? if there is a tokenAddress txs, shouldn't we use ERC20 ABI for value

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -134,14 +155,13 @@ export class EthChain implements IChain {
           } catch (err) {
             output.gasLimit = Defaults.DEFAULT_GAS_LIMIT;
           }
+          if (_.isNumber(opts.fee)) {
+            // This is used for sendmax
+            gasPrice = feePerKb = Number((opts.fee / (inGasLimit || Defaults.DEFAULT_GAS_LIMIT)).toFixed());
+          }
+          gasLimit = inGasLimit || Defaults.DEFAULT_GAS_LIMIT;
+          fee += feePerKb * gasLimit;
```
matiu, line 163: Why this is inside the output loop? `          fee += feePerKb * gasLimit; `?

gabrielbazan7: We were comparing the incorrect amount of fee with the eth linked balance ( also showing the incorrect fee to pay on the confirm view on Copay ). There's two outputs involved (approve and pay) and we were only using the last one.

Example: Those are two invoices payments each of them has 2 fees related.
BitPay

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
gabrielbazan7: if its paypro, it comes back with data property, so its already created the ERC20 transaction, ( we skip it and go directly to ETH transaction create here: https://github.com/bitpay/bitcore/blob/e88001f0f71899c24b8704d6280f44f35d4b44d6/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts#L172 ). 

So outputs look like this: 

```
"outputs":[{"amount":0,"toAddress":"0x44d69d16C711BF966E3d00A46f96e02D16BDdf1f","message":null,"data":"...","gasLimit":29041},{"amount":0,"toAddress":"0xc27eD3DF0DE776246cdAD5a052A9982473FceaB8","message":null,"data":"...","gasLimit":200000}]
```

Both 0 amount ( In a common ERC20 transaction this does not happen we still need to replace the amount by 0 and create the decoded data ). So you would use the second this.output.data for paypro to get the actual value of the erc20 payment. This data uses a custom contract call `pay` (instead of the transfer ERC20 method) that was enconded with the invoice ABI thats why we cant use the ERC20 ABI to decode this data.

The data looks like this: 

``` 
abiDecoder.decodeMethod(data1)
{ name: 'approve',
  params:
   [ { name: '_spender',
       value: '0xc27ed3df0de776246cdad5a052a9982473fceab8',
       type: 'address' },
     { name: '_value', value: '1380623310000000', type: 'uint256' } ] }
> abiDecoder.decodeMethod(data2)
{ name: 'pay',
  params:
   [ { name: 'value', value: '1000000', type: 'uint256' },
     { name: 'gasPrice', value: '40000000000', type: 'uint256' },
     { name: 'expiration', value: '1604123733282', type: 'uint256' },
    ... ] }
```

matiu:
needs rebase.

matiu (approved):

PR #3007, opened 2020-11-11
Migrate Push notifications to httpv1

bsnowden3:
Ref/push Upgrade notification to httpv1 migr

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
+++ b/packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
@@ -16,6 +16,8 @@ const Utils = require('./common/utils');
 const Defaults = require('./common/defaults');
 const Constants = require('./common/constants');
 const sjcl = require('sjcl');
+const { google } = require('googleapis');
+const config = require('../config');
 
 const PUSHNOTIFICATIONS_TYPES = {
   NewCopayer: {
@@ -49,7 +51,6 @@ export interface IPushNotificationService {
   subjectPrefix: string;
   pushServerUrl: string;
   availableLanguages: string;
-  authorizationKey: string;
   messageBroker: any;
 }
 
@@ -61,7 +62,6 @@ export class PushNotificationsService {
   subjectPrefix: string;
   pushServerUrl: string;
   availableLanguages: string;
-  authorizationKey: string;
   storage: Storage;
   messageBroker: any;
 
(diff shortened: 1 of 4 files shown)


matiu (approved):

packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
```diff
@@ -443,19 +446,38 @@ export class PushNotificationsService {
...
+            'Content-Type': 'application/json',
+            Authorization: 'Bearer ' + access_token
+          },
+          body: opts
         },
-        body: opts
-      },
-      cb
-    );
+        cb
```
cmgustavo, line 479: I guess we're not returning `cb` callback to the main function `_sendPushNotifications`. It's important to return any error and handle it better.

My suggestion is to add `catch` and return `cb(err)` to catch error from the request or the access token 

```
this._getAccessToken().then(access_token => {
      this.request(
        {
          url: this.pushServerUrl,
          method: 'POST',
          json: true,
          headers: {
            'Content-Type': 'application/json',
            Authorization: 'Bearer ' + access_token
          },
          body: opts
       }, cb);
    }).catch(err => {
      return cb(err);
    });
```

packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
```diff
@@ -443,19 +446,38 @@ export class PushNotificationsService {
...
+  _getAccessToken() {
+    const MESSAGING_SCOPE = 'https://www.googleapis.com/auth/firebase.messaging';
+    const SCOPES = [MESSAGING_SCOPE];
+
+    return new Promise(function(resolve, reject) {
+      const key = require(config.pushNotificationsOpts.fcmGoogleCredentialsPath);
+      const jwtClient = new google.auth.JWT(key.client_email, null, key.private_key, SCOPES, null);
+      jwtClient.authorize(function(err, tokens) {
+        if (err) {
+          reject(err);
```
cmgustavo, line 458: Maybe in one line 
```
return reject(err);
```

packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
```diff
@@ -443,19 +446,38 @@ export class PushNotificationsService {
...
+          return;
+        }
+        resolve(tokens.access_token);
```
cmgustavo, line 461: Not necessary but a good practice: return resolve(xxx);

PR #3060, opened 2021-01-29
[FEAT][BWS]: endpoint to get rates of all coins

rastajpa:


packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -1179,7 +1197,7 @@ export class ExpressApp {
       } catch (ex) {
         return returnError(ex, res, req);
       }
-      server.getFiatRates(opts, (err, rates) => {
+      server.getFiatRatesByCoin(opts, (err, rates) => {
```
matiu, line 1200: nice.

packages/bitcore-wallet-service/src/lib/fiatrateservice.ts
```diff
@@ -168,6 +153,56 @@ export class FiatRateService {
...
+      },
+      (err, res: any) => {
+        if (err) return cb(err);
+        return cb(null, Object.assign({}, ...res));
+      }
+    );
+  }
+
+  getRatesByCoin(opts, cb) {
+    $.shouldBeFunction(cb, 'Failed state: type error (cb not a function) at <getRatesByCoin()>');
```
matiu, line 204: maybe add:

    if (!checkRequired(opts, ['coin'], cb)) return;
(so, coin is required always).

packages/bitcore-wallet-service/src/lib/fiatrateservice.ts
```diff
@@ -168,6 +153,56 @@ export class FiatRateService {
...
     $.shouldBeFunction(cb, 'Failed state: type error (cb not a function) at <getRates()>');
 
+    opts = opts || {};
+
+    const now = Date.now();
+    const ts = opts.ts ? opts.ts : now;
+    let fiatFiltered = [];
+    let rates = [];
+
+    if (opts.code) {
```
matiu, line 163: is code really optional? what is the expected behavior is not given?

matiu (changes requested):
great job! Left some minor comments.

packages/bitcore-wallet-service/src/lib/fiatrateservice.ts
rastajpa: this check is done in the previous step. Check server.ts

packages/bitcore-wallet-service/src/lib/fiatrateservice.ts
rastajpa: Yes, it is optional, if you don't pass it as a parameter then you get the rates in all fiat supported. if you pass it as a parameter you only get the rate of the specified fiat (code)

PR #3074, opened 2021-02-12
Doge support

kajoseph:


packages/bitcore-node/src/modules/dogecoin/index.ts
```diff
@@ -0,0 +1,14 @@
+import { BaseModule } from '..';
+import { DOGEStateProvider } from '../../providers/chain-state/doge/doge';
+import { DogecoinP2PWorker } from './p2p';
+import { VerificationPeer } from '../bitcoin/VerificationPeer';
+
+export default class DOGEModule extends BaseModule {
+  constructor(services) {
+    super(services);
+    services.Libs.register('DOGE', 'bitcore-lib-doge', 'bitcore-p2p');
+    services.P2P.register('DOGE', DogecoinP2PWorker);
```
micahriggan, line 10: I think you could do

```
 services.P2P.register('DOGE', BitcoinP2PWorker);
 ```
 
 And not need the doge p2p.ts

Comment (micahriggan) [reply packages/bitcore-node/src/modules/dogecoin/index.ts:10]: sourced from: https://github.com/bitpay/bitcore/blob/master/packages/bitcore-node/src/modules/bitcoin-cash/index.ts#L10

**Review by matiu — APPROVED** [review]

File: packages/bitcore-node/src/modules/dogecoin/index.ts (hunk 1 above)
Comment (kajoseph) [reply packages/bitcore-node/src/modules/dogecoin/index.ts:10]: Yeah, the BitcoinP2PWorker would just need some added logic for regtest and minimum dogecore version, so I thought it was a little cleaner to give it it's own p2p.

Comment (micahriggan) [reply packages/bitcore-node/src/modules/dogecoin/index.ts:10]: Gotcha, I see the diff now :D 

I think you could do something like this:
```
export class DogecoinP2PWorker extends BitcoinP2PWorker {
  constructor({ chain, network, chainConfig, blockModel = BitcoinBlockStorage }) {
    super({ chain, network, chainConfig, blockModel });
    if (this.network === 'regtest') {
      this.bitcoreLib.Networks.enableRegtest();
    }
    this.messages = new this.bitcoreP2p.Messages({
      protocolVersion: 70003,
      network: this.bitcoreLib.Networks.get(this.network)
    });
  }
}
```

kajoseph: Wouldn't that still add the message w/ `protocolVersion: 70002` in the super, tho?

micahriggan: Yeah, it would, and then it would overwrite it with your needed version here.

Since the constructor doesn't trigger any methods or lifecycle that accesses this.messages, I believe this should be fine to do.

You may want to re-create the pool as well, since I don't know how it uses the `this.messages` to know if changing the reference would be okay or not.

kajoseph: Ah ok, yeah I missed that it's reassigning `this.messages`. I was thinking it pushed into an array. Cool cool, I'll have to update that in the future. Much cleaner w/ the inheritance.

PR #3076, opened 2021-02-16
CWC doge

sbhatnagar30:


packages/crypto-wallet-core/src/derivation/doge/index.ts
```diff
@@ -0,0 +1,21 @@
+const BitcoreLibDoge = require('bitcore-lib-doge');
+import { AbstractBitcoreLibDeriver } from '../btc';
+export class DogeDeriver extends AbstractBitcoreLibDeriver {
```
micahriggan, line 3:
```
export class DogeDeriver extends AbstractBitcoreLibDeriver {
  bitcoreLib = BitcoreLibDoge;
}
```

packages/crypto-wallet-core/src/transactions/doge/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    const filteredUtxos = this.selectCoins(recipients, utxos, fee);
+    const btcUtxos = filteredUtxos.map(utxo => {
+      const btcUtxo = Object.assign({}, utxo, {
+        amount: utxo.value / 1e8,
+        txid: utxo.mintTxid,
+        outputIndex: utxo.mintIndex
+      });
+      return new this.lib.Transaction.UnspentOutput(btcUtxo);
+    });
+    let tx = new this.lib.Transaction().from(btcUtxos).feePerByte(Number(fee) + 2);
```
micahriggan, line 16: https://github.com/bitpay/bitcore/blob/master/packages/crypto-wallet-core/src/transactions/btc/index.ts

It looks like there's supposed to be a feeRate param for setting feePerByte, fee is supposed to set the total fee.

Does this not work?
```
export class DOGETxProvider extends BTCTxProvider {
  lib = require('bitcore-lib-doge');
}
```

KaycoinsInc (approved):

KaycoinsInc (approved):

KaycoinsInc (approved):

PR #3092, opened 2021-02-25
fix transaction signing and cleanup

sbhatnagar30:


packages/crypto-wallet-core/src/transactions/doge/index.ts
```diff
@@ -13,7 +13,13 @@ export class DOGETxProvider extends BTCTxProvider {
       });
       return new this.lib.Transaction.UnspentOutput(btcUtxo);
     });
-    let tx = new this.lib.Transaction().from(btcUtxos).feePerByte(Number(fee) + 2);
+    let tx = new this.lib.Transaction().from(btcUtxos).feePerKb(Number(fee) + 2);
+    if (fee) {
+      tx.fee(fee);
+    }
+    if (feeRate) {
+      tx.feePerKb(Number(feeRate) / 1000); // feeRate is in feePerByte
```
matiu, line 21: if feeRate is feePerByte, the feePerKb should be  `feePerKb  = feeRate * 1000` right?

matiu (dismissed):

PR #3111, opened 2021-03-04
Improve Insight Search Performance

unusualbob:
### Overview
Implements regex based smart searching to limit api calls to only valid paths based on what input is detected.
As txids and block ids are by far the most shared format, the current implementation of this code avoids querying eth if the 0x prefix is not present. If querying for eth without the 0x prefix is a common occurrence it can be included.

### Performance improvements with this merge for searches:

Input Type | Requests Sent in Master | Requests Sent with this change
------------|----------------------------|-----------------------------------
Garbage data | 24 | 0
Legacy bitcoin address (no bitcoin: prefix) | 24 | 2
Any address with a prefix | 24 | 1
Block or txid | 24 | 12
ETH tx/block ids | 24 | 4

On top of this, there's currently an issue where we **always retry requests**, so even 404's which are legitimate, or worse 429s, are immediately retried. So something like searching for 'test' on live insight results in 40+ requests, all of which fail.

### Additional improvements
* Search results screen is automatically skipped if only one result is returned
* No results search error message when a chain is selected is now more specific to let the user know they are only searching the current chain/network 
Screenshot from 2021-03-04 13-16-28

###  Other notes
As we add more currencies and there is even more overlap of txids and block ids, it may make sense to ask the user which currency they're querying for when their query would hit more than 2 chains/networks.

### Review notes
Please double check I haven't missed any strange edge cases in regards to address/block/txid formats.

### DIFF
--- a/packages/insight/src/providers/search/search.ts
+++ b/packages/insight/src/providers/search/search.ts
@@ -92,7 +92,8 @@ export class SearchProvider {
     },
     // Doge Address
     {
-      regexes: [/^(dogecoin:)?D[5-9A-HJ-NP-U][1-9A-HJ-NP-Za-km-z]{32}/],
+      regexes: [/^(dogecoin:)?(D[5-9A-HJ-NP-U][1-9A-HJ-NP-Za-km-z]{32})/],
+      dataIndex: 2,
       type: 'address',
       chainNetworks: [
         { chain: 'DOGE', network: 'mainnet' }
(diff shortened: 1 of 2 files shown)


matiu:
Nice idea!

packages/insight/src/providers/search/search.ts
```diff
@@ -5,61 +5,191 @@ import { Observable } from 'rxjs';
...
+    {
+      regexes: [/^[0-9]{1,9}$/],
+      type: 'block',
+      chainNetworks: [
+        { chain: 'BTC', network: 'mainnet' },
+        { chain: 'BCH', network: 'mainnet' },
+        { chain: 'DOGE', network: 'mainnet' },
+        { chain: 'BTC', network: 'testnet' },
+        { chain: 'BCH', network: 'testnet' },
+        { chain: 'DOGE', network: 'testnet' }
```
nitsujlangston, line 133: missing ETH block number

nitsujlangston (changes requested):

packages/insight/src/providers/search/search.ts
unusualbob: Fixed

nitsujlangston (approved):

PR #3124, opened 2021-03-23
Organize AuxPow; Add doge data to tests; Fix difficulty and valid PoW

kajoseph:
This PR:

- moves the Aux PoW functions to their own files and utilizes some existing functionality.
- Replaces BTC test data with DOGE test data
- Fixes the difficulty calculation
- Fixes the valid PoW calculation
- Adds Aux PoW serialization to block header serialization
- Points bitcore-node doge module to use p2p-doge


matiu:
great job!

PR #3126, opened 2021-03-25
[FEAT] [BWS][BWC]: new endpoint to get nonce

rastajpa:


matiu (approved):

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2105,6 +2105,20 @@ export class WalletService {
     });
   }
 
+  getNonce(opts) {
+    const bc = this._getBlockchainExplorer(opts.coin, opts.network);
```
matiu, line 2109: maybe add check that network = eth?

packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -1129,6 +1129,28 @@ export class ExpressApp {
       });
     });
 
+    router.get('/v1/nonce/:address', (req, res) => {
```
matiu, line 1132: This request will take a bit, and it is not cached and it is open to everyone. I suggest we do getAuthServer, so at least it has to be attached to a wallet.

matiu (changes requested):
see comments.

packages/bitcore-wallet-service/src/lib/expressapp.ts
rastajpa: good point. It is definitely better to use getAuthServer.

gabrielbazan7 (approved):

PR #3143, opened 2021-04-29
Modernize Litecoin lib

kajoseph:
This PR:

- Fixes wrong Litecoin chain-specific params
- Implements SegWit
- Implements bech32 addresses (Native SegWit)
- Implements multisig
- Fixes tests and replaces BTC data w/ LTC data (where applicable)
- Updates deprecated "new Buffer()" calls to "Buffer.from" or "Buffer.alloc"


nitsujlangston:
Nice work! :tada:

PR #3149, opened 2021-05-07
Fix Send Max for Doge: increase max fee per kb

cmgustavo:


bsnowden3:
ACK!

bsnowden3 (approved):
ACK!

bsnowden3 (commented):
Tested with copay and bws locally, successfully Sent Max amount

PR #3171, opened 2021-05-25
Remove P2P dependence on lib networks

kajoseph:


sbhatnagar30 (approved):
@nitsujlangston this is good to be merged too

PR #3181, opened 2021-06-02
isHexa -> isHexaOrEmpty

isghe:
So it's possible to

```
bitcore.crypto.Hash.sha256 (bitcore.util.buffer.hexToBuffer('')).toString ('hex');
```

that correctly produce
"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"

instead of an assertion error.

This small change is quite invasive anyway, so I think that before being merged, it must be well tested.


isghe:
https://github.com/bitpay/bitcore/pull/3182 makes this PR unnecessary.
So I am closing it.

PR #3182, opened 2021-06-02
hexToBuffer accept empty string

isghe:
So it's possible to:
```
bitcore.util.buffer.hexToBuffer('').toString('hex');
```

that produce an empty string as expected,
instead od an assertion error.

This commit makes https://github.com/bitpay/bitcore/pull/3181 unnecessary.


isghe:
and make it possible to:

```
bitcore.crypto.Hash.sha256 (bitcore.util.buffer.hexToBuffer('')).toString ('hex');
```

that produce `"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"` as expected,
instead of an assertion error.

PR #3184, opened 2021-06-03
Litecoin cwc

sbhatnagar30:


kajoseph (changes requested):
Looks good so far, needs just a couple of changes.

Looks like the CI failed due to external resources...should be good with the next commit.

kajoseph:
Also, it'd be great if you could add a few tests

kajoseph (approved):
Looks good

PR #3186, opened 2021-06-09
Fix ETH transaction query to use scoped index

kajoseph:


micahriggan (commented):
Might wanna check and see if this impacts eth multi sig

sbhatnagar30:
> Might wanna check and see if this impacts eth multi sig

hmm, you think we need to add another $or for abitype multisig? I'm not sure, I'd need to do some more testing around that

kajoseph:
@micahriggan The multisig transactions should still be returned by the query via the `internal.action.to` $or expression as long as the Parity node has tracing enabled

PR #3189, opened 2021-06-14
APP-277 BWC update to Credentials v3

gonzaloleon:
Add "chain" to the **credential**


packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -2716,22 +2754,101 @@ export class API extends EventEmitter {
...
+    // Try to migrate to new Credentials Version
+    _.each(oldCredentials, credential => {
+      let migrated;
+
+      if (!credential.version || credential.version < 2) {
+        log.info('About to migrate : ' + credential.walletId);
+        migrated = API.upgradeCredentialsV1(credential);
+      }
+
+      if (credential.version == 2) {
```
kajoseph, line 2835: Need to check `migrated.version` as well

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -416,6 +416,7 @@ export class Key {
 
     if (password) $.shouldBeString(password, 'provide password');
 
+    // this._checkCoin(opts.coin);
```
kajoseph, line 419: Is this supposed to be commented?

gonzaloleon: Yes, or removed. Coins (constant) contains a limited coins/token, and the goal is to add any token/chain in the future.

kajoseph (approved):
Looks good

cmgustavo:
I've linked the app against this new PR and I get this error. The app doesn't start:

Screen Shot 2021-07-19 at 10 18 19

PR #3196, opened 2021-06-23
[FIX] Handle ERC20 internal txs ( dex )

gabrielbazan7:


kajoseph (commented):
Since this is a transform function (i.e. is piped to), if it throws then it'll be uncaught. A quick fix would be to add a `try..catch`,  but it would be nice to have some tests around this as well.

packages/bitcore-node/src/modules/ethereum/api/erc20Transform.ts
```diff
@@ -24,6 +24,32 @@ export class Erc20RelatedFilterTransform extends Transform {
       tx.abiType.params[8].value.toLowerCase() === this.tokenAddress.toLowerCase()
     ) {
       tx.value = tx.abiType!.params[0].value as any;
+    } else if (tx.internal && tx.internal.length > 0) {
```
kajoseph, line 27: Last thing, could you add a try..catch inside this whole function?

packages/bitcore-node/src/modules/ethereum/api/erc20Transform.ts
```diff
@@ -24,6 +24,37 @@ export class Erc20RelatedFilterTransform extends Transform {
...
+      try {
+        const tokenRelatedIncomingInternalTxs = tx.internal.filter(
+          (internalTx: any) => this.tokenAddress.toLowerCase() === internalTx.action.to.toLowerCase()
```
kajoseph, line 30: There are some transactions with `internal.action.to` == undefined. Need to add a check for the to field

kajoseph (approved):
Looks good!

PR #3200, opened 2021-06-28
Fix uncaught error in XRP ledeger query

kajoseph:


packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -112,22 +112,32 @@ export class RippleStateProvider extends InternalStateProvider implements IChain
...
-          });
-        }
-      });
+      try {
+        const url = this.config[network].provider.dataHost + '/v2/ledgers/' + date;
+        request.get({ url, json: true }, (err, _, body) => {
+          try {
+            if (err) {
+              return reject(err);
+            } else if (body == null) {
```
nitsujlangston, line 121: Is it fully null? May want to do `if(!body.ledger)`. And would it make sense to 404 in that case? Is there an example of what is coming back when there is no ledger property? There may be a constructive error from ripple.

nitsujlangston:
this more condensed change will also fix it
```
else if (!body) {
  return resolve(body);
}
```
without the wrapped try catches

kajoseph:
> this more condensed change will also fix it
> 
> ```
> else if (!body) {
>   return resolve(body);
> }
> ```
> 
> without the wrapped try catches

True, but I didn't want to assume the body wouldn't be another falsey value (0, `false`). `body == null` limits the scope of the body to null or undefined. Anything else will be caught when `body.ledger.anything_else` fails.

Perhaps an even better solution would be to remove the `else if` and let the catch throw an `Unexpected body: ${body}` error so we actually capture the response instead of returning false positives or throwing ambiguous "cannot find property of undefined". Apparently there's nothing that prevents Ripple from changing the response in the future.

nitsujlangston (approved):

nitsujlangston (approved):

PR #3203, opened 2021-07-01
Auto load modules for configured chains

clabby:
Allows auto loading of modules for chains configured in `bitcore.config.json`. 

I've left the option to use `config.modules` intact in case anyone wants to load third party / modified modules as well. If the user does not need to load third-party modules, `config.modules` no longer needs to be present.


kajoseph (dismissed):
Looks good

PR #3205, opened 2021-07-08
[FIX] Point does not lie on the curve error

gabrielbazan7:


clabby:
Looks good, fixes the issue on my end. Nice find!

kajoseph (approved):
NIce

PR #3209, opened 2021-07-13
Insight/hashing assets

TavenShumaker:
**Description:**

This MR adds a node script which hashes img assets so CF cacheing doesn't lead to an out of date site. Refactored components to automatically use the updated assets.

**Repro steps:**

N/A

**Test cases:**

Pull locally, run `$ npm run build:prod` ensure that all .svg files in `insight/src/assets/img` are now hashed, and that the previous versions were deleted. Ensure that all .js and .css files in the build folder are also hashed. Build and serve the project and make sure that all icons and images remain present and the hashed file names are being properly referenced.


GUANLONGH (approved):
good

PR #3214, opened 2021-07-19
Bump web3 version from 1.2.11 -> 1.4.0

clabby:
Bumps web3 version from 1.2.11 -> 1.4.0 in CWC and bitcore-node.

### DIFF
--- a/packages/crypto-wallet-core/package.json
+++ b/packages/crypto-wallet-core/package.json
@@ -36,13 +36,13 @@
     "ripple-binary-codec": "0.2.6",
     "ripple-keypairs": "git+https://git@github.com/bitpay/ripple-keypairs.git#8d3a4643a8ddfce8222786e1e5a3e85a89a5b7f5",
     "ripple-lib": "^1.8.0",
-    "web3": "1.2.11"
+    "web3": "1.4.0"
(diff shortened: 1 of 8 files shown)


packages/crypto-wallet-core/package.json
```diff
@@ -36,13 +36,13 @@
...
     "ripple-lib": "^1.8.0",
-    "web3": "1.2.11"
+    "web3": "1.4.0"
   },
   "devDependencies": {
     "@types/chai": "4.1.7",
     "@types/mocha": "5.2.6",
     "@types/node": "^14.10.0",
-    "@types/web3": "^1.0.19",
+    "@types/web3": "^1.2.2",
```
micahriggan, line 45: https://www.npmjs.com/package/@types/web3

I think this module can be removed since the npm says 
```
web3 provides its own type definitions, so you do not need this installed.
```

clabby: https://github.com/ChainSafe/web3.js#usage-with-typescript

You're right, thanks. Removed the `@types/web3` dependency from CWC and bitcore-node.

kajoseph (approved):
LGTM

PR #3218, opened 2021-07-20
BWS persistent config

clabby:
Moves BWS' root config up from `./src/config.ts` -> `./bws.config.js` (a `.gitignore`d file).


micahriggan:
as a side note, the original goal of bitcore-config was that you could put the configs for bws in there too, just never got done.

That's why the bitcore-node config is nested under the bitcore-node key.

It might be too big of a config file, but thought I'd mention it, in case that was interesting

kajoseph:
> as a side note, the original goal of bitcore-config was that you could put the configs for bws in there too, just never got done.
> 
> That's why the bitcore-node config is nested under the bitcore-node key.
> 
> It might be too big of a config file, but thought I'd mention it, in case that was interesting to you

This is interesting. Thanks for the heads up! We may end up supporting both ways in the future to leave the user with options.

packages/bitcore-wallet-service/src/config.ts
```diff
@@ -1,188 +1,203 @@
-module.exports = {
-  basePath: '/bws/api',
-  disableLogs: false,
-  port: 3232,
+const { logger } = require('./lib/logger');
```
kajoseph, line 1: Should use the `import` syntax instead of `require`. This package is kind of weird in that it mixes ES6 import syntax with CommonJS module.exports, but we should stick to the ES6 syntax where possible.

packages/bitcore-wallet-service/src/config.ts
```diff
@@ -1,188 +1,203 @@
...
+    //
+    //
+    // //then add:
+    // mailer: sgMail,
+  };
+
+  // Override default values with bws.config.js' values, if present
+  try {
+    const bwsConfig = require('../bws.config');
+    defaultConfig = Object.assign({}, defaultConfig, bwsConfig);
```
kajoseph, line 196: Should use lodash's merge method to recursively merge. 

e.g.:
```
import * as _ from 'lodash';

...

defaultConfig =  _.merge(defaultConfig, bwsConfig);
```

packages/bitcore-wallet-service/src/config.ts
clabby: Updated

packages/bitcore-wallet-service/src/config.ts
clabby: Updated

kajoseph (approved):
LGTM

PR #3227, opened 2021-07-27
Exclude invalid transactions from Wallet.listTransactions

clabby:
Excludes transactions that have been invalidated from the Wallet.listTransactions query by default.

https://github.com/bitpay/bitcore/blob/master/packages/bitcore-node/src/types/Coin.ts#L43
https://github.com/bitpay/bitcore/blob/master/packages/bitcore-node/src/services/pruning.ts#L152-L166


micahriggan (approved):

kajoseph (approved):
Nice job

PR #3238, opened 2021-08-04
Add includeInvalidTxs parameter into bitcore-client & BWS query methods

clabby:
#3227

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
+++ b/packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
@@ -288,7 +288,7 @@ export class V8 {
       .catch(cb);
   }
 
-  getTransactions(wallet, startBlock, cb) {
+  getTransactions(wallet, startBlock, cb, includeInvalidTxs = false) {
     console.time('V8 getTxs');
     if (startBlock) {
       logger.debug(`getTxs: startBlock ${startBlock}`);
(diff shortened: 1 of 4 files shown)


packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
```diff
@@ -288,7 +288,7 @@ export class V8 {
       .catch(cb);
   }
 
-  getTransactions(wallet, startBlock, cb) {
+  getTransactions(wallet, startBlock, cb, includeInvalidTxs = false) {
```
kajoseph, line 291: Let's keep the callback as the last param

packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8/client.ts
```diff
@@ -126,25 +127,24 @@ export class Client {
...
+    const rawQuery = {
+      startBlock,
+      endBlock,
+      startDate,
+      endDate,
+      tokenAddress,
+      includeMempool: includeMempool && true,
+      includeInvalidTxs
+    };
+    const query = _.map(_.keys(rawQuery), key => rawQuery[key] && `${key}=${rawQuery[key]}`).join('&');
```
kajoseph, line 144: We shouldn't need lodash here

kajoseph (commented):
Overall looks good, just a couple of minor changes

packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
clabby: Updated

packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8/client.ts
clabby: Updated

PR #3249, opened 2021-08-20
[FIX] set default gas limit in the correct place

gabrielbazan7:
Was undefined in line 176

Slack___justtango___BitPay


Gamboster (approved):
ACK

kajoseph (approved):

PR #3252, opened 2021-08-25
[FIX] Server assisted import for custom tokens

gabrielbazan7:


packages/bitcore-wallet-client/test/api.test.js
```diff
@@ -6526,6 +6526,37 @@ describe('client API', function() {
...
+              Client.serverAssistedImport(
+              { words },
+              {
+                clientFactory: () => {
+                  return helpers.newClient(app);
+                }
+              },
+              (err, k, c) => {
+                // the eth wallet + 1 unknown token addresses on preferences.
+                c.length.should.equal(1);
```
kajoseph, line 6546: Should this be 2? The comment above makes me think it should be 2, but the oneInch API doesn't have a match for the tokenAddresses passed in preferences.

gabrielbazan7: No, one is ok. I'm just testing that if there is an address it does not exist in the 1inch api retrieves the eth wallet without issues

kajoseph (approved):
LGTM

Gamboster (commented):
ACK. Tested on my local BWS

PR #3256, opened 2021-08-31
[FIX] getChain for custom ERC20 tokens

Gamboster:


kajoseph (approved):
Good comment. Will definitely want to revisit this when we add another smart contract chain.

LGTM

PR #3278, opened 2021-10-13
Feat: check service availability

Gamboster:


gabrielbazan7 (dismissed):
ACK!

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -5029,6 +5029,32 @@ export class WalletService {
     });
   }
 
+  checkServiceAvailability(req): Promise<boolean> {
+    return new Promise((resolve, reject) => {
```
kajoseph, line 5033: I'd like to move away from this sort of verbosity. Could you instead just make this function an `async` function and throw/return below?

packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -1315,6 +1315,23 @@ export class ExpressApp {
       });
     });
 
+    router.post('/v1/service/checkAvailability', (req, res) => {
```
kajoseph, line 1318: I'd like to move away from promise chains and callbacks and instead use async/await syntax. Could you make this router function `async` and extend the try..catch below to encompass the whole function and `await server.checkServiceAvailability`?

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -5029,6 +5029,32 @@ export class WalletService {
...
+      if (!checkRequired(req.body, ['service', 'opts'])) {
+        return reject(new ClientError('checkServiceAvailability request missing arguments'));
+      }
+
+      let serviceEnabled: boolean;
+
+      switch (req.body.service) {
+        case '1inch':
+          if (req.body.opts && req.body.opts.country && req.body.opts.country.toUpperCase() == 'US') {
```
kajoseph, line 5042: This can be shortened to `req.body.opts?.country?.toUpperCase() === 'US'`

kajoseph (changes requested):

packages/bitcore-wallet-service/src/lib/server.ts
Gamboster: Done. Thanks

packages/bitcore-wallet-service/src/lib/expressapp.ts
Gamboster: Fixed. I avoided using promises for this particular function

packages/bitcore-wallet-service/src/lib/server.ts
Gamboster: Fixed. I avoided using promises for this particular function

kajoseph (approved):
Looks good! Thanks

PR #3300, opened 2021-11-08
Add RSK support to Bitcore Wallet Client

alepc253:


packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -2515,7 +2517,7 @@ export class API extends EventEmitter {
   // * @return {Callback} cb - Return error (if exists) and nonce
   // */
   getNonce(opts, cb) {
-    $.checkArgument(opts.coin == 'eth', 'Invalid coin: must be "eth"');
+    $.checkArgument(opts.coin == 'eth' || opts.coin == 'rsk' || opts.coin == 'rbtc', 'Invalid coin: must be "eth"');
```
micahriggan, line 2520: I think you don't want this line

alepc253: True, I can re-write this

alepc253: BTW, I'm using both RSK and RBTC because I detected in the frontend an assumption that "chain" and "coin" are the same

PR #3318, opened 2021-12-28
Implement taproot (BIP-0340 and BIP-0341)

kajoseph:


benjamintshi:
hi guys  , when this request will be passed ?

kajoseph:
Hi @benjamintshi, this is a work in progress that is still on my radar but unfortunately has fallen behind some other more pressing priorities at the moment. I'm hoping to return and finish this before the end of March :crossed_fingers:

benjamintshi:
> 

Hi kajoseph , thanks .

benjamintshi:
Hi @kajoseph  , Is there any new progress on this?

kajoseph:
Hey @benjamintshi , unfortunately not. It was overshadowed by our launch of lightning at first and now we're getting ready for the pending ETH merge.

This is still front of mind, though, and definitely something I want to get done.

packages/bitcore-lib/lib/crypto/schnorr.js
```diff
@@ -0,0 +1,138 @@
...
+  if (typeof aux === 'string' && JS.isHexaString(aux)) {
+    aux = Buffer.from(aux, 'hex')
+  }
+  $.checkArgument($.isType(aux, 'Buffer'), 'aux must be a hex string or buffer');
+
+  const G = Point.getG();
+  const n = Point.getN();
+
+  const dPrime = new BN(privateKey);
+  if (dPrime.eqn(0) || dPrime.gte(Point.getN())) {
```
MichaelAJay, line 48: Considering using constant `n`, assigned on line 45

packages/bitcore-lib/lib/crypto/schnorr.js
```diff
@@ -0,0 +1,138 @@
...
+ * @returns {Buffer}
+ * @link https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Default_Signing
+ */
+Schnorr.sign = function(privateKey, message, aux) {
+  privateKey = Buffer.isBuffer(privateKey) ? privateKey : privateKey.toBuffer();
+  if (privateKey.length !== 32) {
+    throw new Error('Private key should be 32 bytes for schnorr signatures');
+  }
+
+  if (typeof message === 'string' && JS.isHexaString(message)) {
```
MichaelAJay, line 31: What should happen if `typeof message === 'string'` is true, but JS.isHexaString(message) is false?

On the buffer concatenation of the assignment to "rand" on line 55, 'message' is not ensured to be a Buffer.

MichaelAJay: I see - the following line is ensuring that message is a Buffer further in the code. Here's a way that should perform nearly the same number of operations, but provides better IDE type help later in the code:
`  
  // if (typeof message === 'string' && JS.isHexaString(message)) {
  //   message = Buffer.from(message, 'hex')
  // }
  // $.checkArgument($.isType(message, 'Buffer'), 'message must be a hex string or buffer');
  if (typeof message === 'string') {
    $.checkArgument(JS.isHexaString(message), 'message must be a hex string or buffer');
    message = Buffer.from(message, 'hex')
  }
  $.checkArgument($.isType(message, 'Buffer'), 'message must be a hex string or buffer');
`

packages/bitcore-lib/lib/crypto/schnorr.js
```diff
@@ -0,0 +1,138 @@
...
+ */
+Schnorr.verify = function(publicKey, message, signature) {
+  if ($.isType(publicKey, 'PublicKey')) {
+    publicKey = publicKey.point.x.toBuffer();
+  }
+  if (publicKey.length !== 32) {
+    throw new Error('Public key should be 32 bytes for schnorr signatures');
+  }
+
+  if (typeof message === 'string' && JS.isHexaString(message)) {
```
MichaelAJay, line 89: See conversation in 'Schnorr.sign'.

packages/bitcore-lib/test/script/interpreter.js
```diff
@@ -404,9 +377,10 @@ describe('Interpreter', function() {
   describe('bitcoind transaction evaluation fixtures', function() {
     var test_txs = function(set, expected) {
       var c = 0;
-      set.forEach(function(vector) {
+      for (let i in set) {
```
MichaelAJay, line 380: Why not
```
for (let vector of set) {
   ...
}
```

You're intending to iterate over each element in 'set', right?

packages/bitcore-lib/test/script/interpreter.js
```diff
@@ -443,18 +418,101 @@ describe('Interpreter', function() {
...
+
+  describe('bitcoind script asset tests', function() {
+    this.timeout(60000); // Some tests run a little long
+
+    for (let i = 0; i < script_asset_tests.length; i++) {
+      let test = script_asset_tests[i];
+      it(`script asset test vector ${i}: ${test.comment}`, function() {
+        const tx =  new Transaction(test.tx);
+        const prevOuts = [];
+        for (let j in test.prevouts) {
```
MichaelAJay, line 467: Why not for of?
for (let prevout of test.prevouts)

packages/bitcore-lib/test/script/interpreter.js
```diff
@@ -443,18 +417,99 @@ describe('Interpreter', function() {
...
+
+  describe('bitcoind script asset tests', function() {
+    this.timeout(60000); // Some tests run a little long
+
+    for (let i = 0; i < script_asset_tests.length; i++) {
+      let test = script_asset_tests[i];
+      it(`script asset test vector ${i}: ${test.comment}`, function() {
+        const tx =  new Transaction(test.tx);
+        const prevOuts = [];
+        for (let prevout in test.prevouts) {
```
MichaelAJay, line 465: should be for of.

packages/bitcore-lib/test/script/interpreter.js
```diff
@@ -416,21 +389,22 @@ describe('Interpreter', function() {
...
               txoutnum = 0xffffffff; //bitcoind casts -1 to an unsigned int
             }
             map[txid + ':' + txoutnum] = Script.fromBitcoindString(scriptPubKeyStr);
-          });
+          };
 
           var tx = new Transaction(txhex);
           var allInputsVerified = true;
-          tx.inputs.forEach(function(txin, j) {
+          for (let j in tx.inputs) {
```
MichaelAJay, line 404: I think this should be for of again. Although you are passing j into verify, here:
`
var verified = interp.verify(scriptSig, scriptPubkey, tx, j, flags);
`

Are you meaning to pass string representations of tx.inputs' indices?

packages/bitcore-lib/test/script/interpreter.js
kajoseph: I reverted this back to a for..in loop since the index is needed in the test.

packages/bitcore-lib/test/script/interpreter.js
kajoseph: keeping this as a for..in because as you noted the index is used in the test case. Yeah, a string fine. `array[0] === array['0']`

packages/bitcore-lib/lib/transaction/taprootBuilder.js
MichaelAJay, line 1: No export. Keeping it in for future use?

kajoseph: removed. I think this snuck in as part of a WIP commit

packages/bitcore-lib/lib/transaction/sighashschnorr.js
```diff
@@ -0,0 +1,231 @@
...
+ * Returns a buffer of length 32 bytes with the hash that needs to be signed
+ * for witness v1 programs as defined by:
+ * https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki
+ *
+ * @name Signing.sighash
+ * @param {Transaction} transaction the transaction to sign
+ * @param {Number} sighashType the type of the hash
+ * @param {Number} inputNumber the input index for the signature
+ * @param {Number} sigversion Taproot or Tapscript version number
+ * @param {Object} execdata Object with directives and data for creating the signature hash
```
MichaelAJay, line 23: `@param {Record<string, any>} execdata...` gives a little better type help.

packages/bitcore-lib/lib/transaction/transaction.js
```diff
@@ -1232,31 +1236,33 @@ Transaction.prototype.removeInput = function(txId, outputIndex) {
...
-  sigtype = sigtype || Signature.SIGHASH_ALL;
-  var transaction = this;
-  var results = [];
-  var hashData = Hash.sha256ripemd160(privKey.publicKey.toBuffer());
-  for (let index = 0; index < this.inputs.length; index++) {
-    const input = this.inputs[index];
-    for (const signature of input.getSignatures(transaction, privKey, index, sigtype, hashData, signingMethod)) {
+  const results = [];
+  const hashData = Hash.sha256ripemd160(privKey.publicKey.toBuffer());
+  for (let i = 0; i < this.inputs.length; i++) {
```
MichaelAJay, line 1263: Consider flatMap & renaming 'results'
`
const hashData = Hash.sha256ripemd160(privKey.publicKey.toBuffer());
const results = this.inputs.flatMap((input, i) =>
  input.getSignatures(this, privKey, i, sigtype, hashData, signingMethod, merkleRoot)
);
`

kajoseph: I'm inclined to keep this as is. The flatMap usage looks clean, but in practice it's pretty common to want to step into getSignatures with a debugger and/or try to debug why a specific input isn't returning expected sigs and the existing syntax is better for that.

packages/bitcore-lib/lib/transaction/sighashschnorr.js
kajoseph: `Record` is a typescript primitive whereas this is a vanilla JS project.

packages/bitcore-lib/lib/transaction/sighashschnorr.js
```diff
@@ -0,0 +1,231 @@
...
+
+  const ss = TaggedHash.TAPSIGHASH;
+
+  // Epoch
+  ss.writeUInt8(0);
+
+  // Hash type
+  const outputType = (sighashType == Signature.SIGHASH_DEFAULT) ? Signature.SIGHASH_ALL : (sighashType & Signature.SIGHASH_OUTPUT_MASK); // Default (no sighash byte) is equivalent to SIGHASH_ALL
+  const inputType = sighashType & Signature.SIGHASH_INPUT_MASK;
+  if (!(sighashType <= 0x03 || (sighashType >= 0x81 && sighashType <= 0x83))) {
```
MichaelAJay, line 54: Feels like a comment would be useful above this conditional

packages/bitcore-lib/lib/transaction/input/taproot.js
```diff
@@ -0,0 +1,106 @@
...
+ * @param {number} signature.sigtype
+ * @return {TaprootInput} this, for chaining
+ */
+TaprootInput.prototype.addSignature = function(transaction, signature) {
+  if (this.isValidSignature(transaction, signature)) {
+    this.setWitnesses([
+      signature.signature.toBuffer(),
+    ]);
+  }
+  // else... do nothing?
```
MichaelAJay, line 96: Are you happy with the current implementation w/r to your notes here?

packages/bitcore-lib/lib/transaction/sighash.js
```diff
@@ -97,26 +99,12 @@ var sighash = function sighash(transaction, sighashType, inputNumber, subscript)
  * @param {number} sighash
  * @param {number} inputIndex
  * @param {Script} subscript
- * @param {String} signingMethod - method used to sign - 'ecdsa' or 'schnorr' (future signing method)
  * @return {Signature}
  */
-function sign(transaction, privateKey, sighashType, inputIndex, subscript, signingMethod) {
-  signingMethod = signingMethod || 'ecdsa';
-
+function sign(transaction, privateKey, sighashType, inputIndex, subscript) {
```
MichaelAJay, line 104: Consider updating method params to reflect new argument list:

- MultiSigInput.prototype.getSignatures
- MultiSigScriptHashInput.prototype.getSignatures
- PublicKeyInput.prototype.getSignatures
- PublicKeyHashInput.prototype.getSignatures

packages/bitcore-lib/lib/transaction/sighash.js
```diff
@@ -127,29 +115,15 @@ function sign(transaction, privateKey, sighashType, inputIndex, subscript, signi
  * @param {PublicKey} publicKey
  * @param {number} inputIndex
  * @param {Script} subscript
- * @param {String} signingMethod - method used to sign - 'ecdsa' or 'schnorr'
  * @return {boolean}
  */
-function verify(transaction, signature, publicKey, inputIndex, subscript, signingMethod) {
+function verify(transaction, signature, publicKey, inputIndex, subscript) {
```
MichaelAJay, line 120: Consider updating method call params to reflect new arguments:

- Input.prototype.isValidSignature
- MultiSigInput.prototype.isValidSignature
- MultiSigInput.normalizeSignatures
- MultiSigScriptHashInput.prototype.isValidSignature
- PublicKeyHashInput.prototype.isValidSignature

packages/bitcore-lib/lib/transaction/sighashschnorr.js
```diff
@@ -0,0 +1,231 @@
...
+    const spentScriptsBW = new BufferWriter();
+    const sequencesBW = new BufferWriter();
+
+    for(let vin of transaction.inputs) {
+      prevoutsBW.writeReverse(vin.prevTxId);
+      prevoutsBW.writeInt32LE(vin.outputIndex);
+
+      spentAmountsBW.writeUInt64LEBN(vin.output._satoshisBN);
+
+      spentScriptsBW.writeUInt8(vin.output.script.toBuffer().length);
```
MichaelAJay, line 74: Could assign vin.output.script.toBuffer() so you don't call toBuffer() twice.
`
const scriptBuffer = vin.output.script.toBuffer();
spentScriptsBW.writeUInt8(scriptBuffer.length);
spentScriptsBW.write(scriptBuffer);
`

packages/bitcore-lib/lib/transaction/sighashschnorr.js
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * Verify a Schnorr signature
+ *
+ * @name Signing.verify
+ * @param {Transaction} transaction
+ * @param {Signature} signature
+ * @param {PublicKey} publicKey
+ * @param {Number} inputIndex
+ * @param {object|Buffer} execdata Can be full execdata object or just the leafHash buffer
```
MichaelAJay, line 197: When called from taproot.js, execdata is null. Consider adding to JSDoc type

packages/bitcore-lib/lib/transaction/input/publickey.js
```diff
@@ -25,8 +25,10 @@ inherits(PublicKeyInput, Input);
...
  * @param {number} index - the index of the input in the transaction input vector
- * @param {number=} sigtype - the type of signature, defaults to Signature.SIGHASH_ALL
- * @param {String} signingMethod - method used to sign input - 'ecdsa' or 'schnorr' (future signing method)
+ * @param {number} sigtype - the type of signature, defaults to Signature.SIGHASH_ALL
+ * @param {Buffer} hashData - unused for this input type 
+ * @param {String} signingMethod - method used to sign input - 'ecdsa' or 'schnorr'
+ * @param {Buffer} merkleRoot - unused for this input type
  * @return {Array} of objects that can be
  */
 PublicKeyInput.prototype.getSignatures = function(transaction, privateKey, index, sigtype, hashData, signingMethod) {
```
MichaelAJay, line 34: merkleRoot is not included in the arguments

packages/bitcore-lib/lib/script/script.js
```diff
@@ -356,6 +366,8 @@ Script.prototype.getPublicKeyHash = function() {
     return this.chunks[2].buf;
   } else if (this.isWitnessPublicKeyHashOut()) {
     return this.chunks[1].buf;
+  } else if (this.isTaproot()) {
```
MichaelAJay, line 369: Could condense with OR of statement above

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -70,26 +72,133 @@ Interpreter.prototype.verifyWitnessProgram = function(version, program, witness,
...
+      // Drop annex (this is non-standard; see IsWitnessStandard)
+      const annex = stack.pop();
+      const annexWriter = new BufferWriter();
+      annexWriter.writeVarintNum(annex.length);
+      annexWriter.write(annex);
+      execdata.annexHash = Hash.sha256(annexWriter.toBuffer());
+      execdata.annexPresent = true;
+    }
+    execdata.annexInit = true;
+    if (stack.length === 1) {
```
MichaelAJay, line 101: This could simply be:
`
if (stack.length === 1) {
  return this.checkSchnorrSignature(...)
}
`

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -70,26 +72,133 @@ Interpreter.prototype.verifyWitnessProgram = function(version, program, witness,
...
+          // Note how this condition would not be reached if an unknown OP_SUCCESSx was found
+          this.errstr = 'SCRIPT_ERR_BAD_OPCODE';
+          return false;
+        }
+
+        execdata.validationWeightLeft = witnessSize + Script.VALIDATION_WEIGHT_OFFSET;
+        execdata.validationWeightLeftInit = true;
+        return this.executeWitnessScript(scriptPubKey, stack, Signature.Version.TAPSCRIPT, satoshis, flags, execdata);
+      }
+      if (flags & Interpreter.SCRIPT_VERIFY_DISCOURAGE_UPGRADABLE_TAPROOT_VERSION) {
```
MichaelAJay, line 152: Are there superceding cases above? Could there be? Or is this a condition which should supercede and return false?

If so, putting this simple check at the top could improve readability.

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -70,26 +72,133 @@ Interpreter.prototype.verifyWitnessProgram = function(version, program, witness,
...
     script: scriptPubKey,
     stack: stack,
-    sigversion: Interpreter.SIGVERSION_WITNESS_V0,
+    sigversion: sigversion,
     satoshis: satoshis,
     flags: flags,
+    execdata: execdata
   });
 
   if (!this.evaluate()) {
```
MichaelAJay, line 204: Proposed cleanup: add this.errstr for uniformity

MichaelAJay: Similar cases below

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -307,18 +421,19 @@ Interpreter.prototype.initialize = function(obj) {
...
   this.errstr = '';
   this.flags = 0;
+  this.execdata = {};
 };
 
 Interpreter.prototype.set = function(obj) {
   this.script = obj.script || this.script;
   this.tx = obj.tx || this.tx;
-  this.nin = typeof obj.nin !== 'undefined' ? obj.nin : this.nin;
+  this.nin = typeof obj.nin !== 'undefined' ? parseInt(obj.nin) : this.nin;
```
MichaelAJay, line 436: This reads as "If !A then B, else C"
Could simplify logic with "If A then C, else B"
this.nin = typeof obj.nin === 'undefined' ? this.nin : parseInt(obj.nin)

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -329,11 +444,14 @@ Interpreter.prototype.set = function(obj) {
   this.vfExec = obj.vfExec || this.vfExec;
   this.errstr = obj.errstr || this.errstr;
   this.flags = typeof obj.flags !== 'undefined' ? obj.flags : this.flags;
+  this.execdata = typeof obj.execdata !== 'undefined' ? (obj.execdata || {}) : this.execdata
```
MichaelAJay, line 447: Same as above

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -509,21 +661,230 @@ Interpreter.prototype.checkPubkeyEncoding = function(buf) {
...
+  $.checkArgument(pubkey && Buffer.isBuffer(pubkey), 'Missing pubkey');
+  $.checkArgument(sigversion, 'Missing sigversion');
+  $.checkArgument(execdata, 'Missing execdata');
+
+  $.checkArgument(pubkey.length === 32, 'Schnorr signatures have 32-byte public keys. The caller is responsible for enforcing this.');
+  // Note that in Tapscript evaluation, empty signatures are treated specially (invalid signature that does not
+  // abort script execution). This is implemented in EvalChecksigTapscript, which won't invoke
+  // CheckSchnorrSignature in that case. In other contexts, they are invalid like every other signature with
+  // size different from 64 or 65.
+  if (sig.length !== 64 && sig.length !== 65) {
```
MichaelAJay, line 722: `if (!(sig.length === 64 || sig.length === 65))` better expresses the spirit of the check

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -610,54 +965,104 @@ Interpreter.prototype.checkLockTime = function(nLockTime) {
...
+  if (!((txToSequenceMasked.lt(SEQUENCE_LOCKTIME_TYPE_FLAG_BN)  &&
+          nSequenceMasked.lt(SEQUENCE_LOCKTIME_TYPE_FLAG_BN)) ||
+        (txToSequenceMasked.gte(SEQUENCE_LOCKTIME_TYPE_FLAG_BN) &&
+          nSequenceMasked.gte(SEQUENCE_LOCKTIME_TYPE_FLAG_BN)))) {
+    return false;
+  }
+
+  // Now that we know we're comparing apples-to-apples, the comparison is a
+  // simple numeric one.
+  if (nSequenceMasked.gt(txToSequenceMasked)) {
```
MichaelAJay, line 1011: Instead of
`
  if (nSequenceMasked.gt(txToSequenceMasked)) {
    return false;
  }
  return true;
`

could just:
`return nSequenceMasked.lte(txToSequenceMasked)`

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -610,54 +965,104 @@ Interpreter.prototype.checkLockTime = function(nLockTime) {
...
+    //! The internal pubkey (x-only, so no Y coordinate parity).
+    const p = PublicKey.fromX(false, control.slice(1, Interpreter.TAPROOT_CONTROL_BASE_SIZE));
+    //! The output pubkey (taken from the scriptPubKey).
+    const q = PublicKey.fromX(false, program);
+    // Compute the Merkle root from the leaf and the provided path.
+    const merkleRoot = Interpreter.computeTaprootMerkleRoot(control, tapleafHash);
+    // Verify that the output pubkey matches the tweaked internal pubkey, after correcting for parity.
+    return q.checkTapTweak(p, merkleRoot, control);
+  } catch (err) {
+    // TODO: set this.errstr here?
```
MichaelAJay, line 1061: ping in case you want to address this TODO

packages/bitcore-lib/lib/opcode.js
```diff
@@ -75,6 +75,36 @@ Opcode.smallInt = function(n) {
...
+/**
+ * Converts OP_0 - OP_16 to an int
+ * @param {Number} opcode Opcode
+ * @returns {Number}
+ */
+Opcode.decodeOpN = function(opcode) {
+  if (opcode === Opcode.OP_0) {
+    return 0;
+  }
+  $.checkArgument(opcode >= Opcode.OP_1 && opcode <= Opcode.OP_16, new Error('Invalid opcode: ' + JSON.stringify(opcode)));
```
MichaelAJay, line 87: check usage (doesn't look like it's supposed to take an Error)

packages/bitcore-lib/lib/crypto/signature.js
```diff
@@ -26,9 +27,11 @@ Signature.prototype.set = function(obj) {
   this.r = obj.r || this.r || undefined;
   this.s = obj.s || this.s || undefined;
 
-  this.i = typeof obj.i !== 'undefined' ? obj.i : this.i; //public key recovery parameter in range [0, 3]
-  this.compressed = typeof obj.compressed !== 'undefined' ?
-    obj.compressed : this.compressed; //whether the recovered pubkey is compressed
+  // public key recovery parameter in range [0, 3]
+  this.i = typeof obj.i !== 'undefined' ? obj.i : this.i;
```
MichaelAJay, line 31: Recommend switching from "If not A then B, else C" to "If A then C, else B"
`this.i = typeof obj.i === 'undefined' ? this.i : obj.i`

Same as below

packages/bitcore-lib/lib/crypto/signature.js
```diff
@@ -61,9 +64,18 @@ Signature.fromCompact = function(buf) {
...
-  var obj = Signature.parseDER(buf, strict);
   var sig = new Signature();
 
+  // Schnorr Signatures use 65 byte for in tx r [len] 32 , s [len] 32, nhashtype
+  // NOTE: this check is not very reliable. You should use .fromSchnorr directly if you know it's a schnorr sig.
+  if((buf.length === 64 || buf.length === 65) && buf[0] != 0x30) {
+    return Signature.fromSchnorr(buf);
+  }
+  
+  $.checkArgument(!(buf.length === 64 && buf[0] === 0x30), new Error('64 DER (ecdsa) signatures not allowed'));
```
MichaelAJay, line 75: Check method signature

packages/bitcore-lib/lib/publickey.js
```diff
@@ -281,6 +296,115 @@ PublicKey.fromX = function(odd, x) {
   });
 };
 
+/**
+ * PublicKey instance from a Taproot (32-byte) public key
+ * @param {String|Buffer} hexBuf 
+ * @returns {PublicKey}
+ */
+PublicKey.fromTaproot = function(hexBuf) {
```
MichaelAJay, line 304: Consider refactoring fromTaproot and isValidTaproot to abstract common functionality, which:
- Updates param to Buffer IF param is hex string
- Ensures param is now 32 byte Buffer

packages/bitcore-lib/lib/publickey.js
```diff
@@ -281,6 +296,115 @@ PublicKey.fromX = function(odd, x) {
...
+PublicKey.prototype.checkTapTweak = function(p, merkleRoot, control) {
+  if (Buffer.isBuffer(p)) {
+    p = PublicKey.fromTaproot(p);
+  }
+  const tweak = p.computeTapTweakHash(merkleRoot);
+
+  const P = p.point.liftX();
+  const Q = P.add(this.point.curve.g.mul(BN.fromBuffer(tweak)));
+  
+  if (!this.point.x.eq(Q.x)) {
```
MichaelAJay, line 380: Instead of 380 - 386, you could:
`return this.point.x.eq(Q.x) && Q.y.mod(new BN(2)).eq(new BN(control[0] & 1)`

packages/bitcore-lib/lib/transaction/input/taproot.js
kajoseph: Per the rest of the comment, refactoring this would be major scope creep.

packages/bitcore-lib/lib/script/interpreter.js
kajoseph: this.errstr is set inside this.evaluate()

packages/bitcore-lib/lib/script/interpreter.js
kajoseph: The only place that uses this fn sets `this.errstr` if false. Removed comment.

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -1494,7 +1900,37 @@ Interpreter.prototype.step = function() {
...
+            return false;
+          }
+
+          // (sig num pubkey -- num)
+          if (this.stack.length < 3) {
+            this.errstr = 'SCRIPT_ERR_INVALID_STACK_OPERATION';
+            return false;
+          }
+
+          let sig = this.stack[this.stack.length - 3];
```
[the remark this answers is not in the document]
kajoseph: correct, there are cases where the stack needs to remain intact through the validation.

packages/bitcore-lib/lib/opcode.js
kajoseph: It's ok. I added some test cases for these 2 new methods, one of which tests for this.

packages/bitcore-lib/lib/script/interpreter.js
```diff
@@ -307,33 +419,37 @@ Interpreter.prototype.initialize = function(obj) {
...
+  this.execdata = {};
 };
 
 Interpreter.prototype.set = function(obj) {
   this.script = obj.script || this.script;
   this.tx = obj.tx || this.tx;
-  this.nin = typeof obj.nin !== 'undefined' ? obj.nin : this.nin;
+  this.nin = typeof obj.nin === 'undefined' ? this.nin : parseInt(obj.nin);
   this.stack = obj.stack || this.stack;
   this.altstack = obj.altack || this.altstack;
```
MichaelAJay, line 436: Is this a typo? 'obj.altack' - instead of 'obj.altstack'

packages/bitcore-lib/test/publickey.js
```diff
@@ -451,4 +451,36 @@ describe('PublicKey', function() {
...
+      const isValid = PublicKey.isValidTaproot('29b21ed3959615c97d0866a295486c25577aaa5ee18d463489d9dbe3cf0aaf5e');
+      isValid.should.equal(true);
+    });
+
+    it('should be true - buffer', function() {
+      const isValid = PublicKey.isValidTaproot(Buffer.from('29b21ed3959615c97d0866a295486c25577aaa5ee18d463489d9dbe3cf0aaf5e', 'hex'));
+      isValid.should.equal(true);
+    });
+
+    it('should be false - hex string', function() {
```
MichaelAJay, line 465: If you view tests as documentation, then either a more descriptive test name, or even just a comment in the test, would be useful. Why is this input invalid when the above input is valid?

packages/bitcore-lib/test/publickey.js
```diff
@@ -451,4 +451,36 @@ describe('PublicKey', function() {
 
   });
 
+  describe('#isValidTaproot', function() {
```
MichaelAJay, line 454: These are just some ideas. Consider replacing string values in 'isValidTaproot' inputs below with named constants defined in the describe block.
`
    const VALID_TAPROOT = '29b21ed3959615c97d0866a295486c25577aaa5ee18d463489d9dbe3cf0aaf5e';
    
    const INVALID_TAPROOT = 'bca10d4006dbfebe31eee345df1d18d738f72fcd62d6ccafa84f87f05bac3467'; // Why is this an invalid hex string?
    
    const INVALID_LENGTH = VALID_TAPROOT.slice(0, -1); // String length is different from valid
    
    const INVALID_HEX_DIGIT = 'g'; // Documents example of why a "hex string" may be invalid
    
    const INVALID_HEX = INVALID_LENGTH + INVALID_HEX_DIGIT;
    
`

MichaelAJay, line 454: Consider including Chai assertions that duplicate testing behavior (assert on length, assert on characters, etc)

MichaelAJay, line 454: Create a test which stubs fromTaproot and throws/handles an error, returning false.

packages/bitcore-lib/lib/crypto/signature.js
```diff
@@ -61,9 +64,18 @@ Signature.fromCompact = function(buf) {
 };
 
 Signature.fromDER = Signature.fromBuffer = function(buf, strict) {
-  var obj = Signature.parseDER(buf, strict);
   var sig = new Signature();
 
+  // Schnorr Signatures use 65 byte for in tx r [len] 32 , s [len] 32, nhashtype
+  // NOTE: this check is not very reliable. You should use .fromSchnorr directly if you know it's a schnorr sig.
+  if((buf.length === 64 || buf.length === 65) && buf[0] != 0x30) {
```
MichaelAJay, line 71: Is there any way that we can have greater reliability here? One thing that jumped out at me is that a Signature has 'isSchnorr' property on it.

It seems like using that property here would be better. And if it needs to be initialized better, then you can do so in just one place, and there won't be these dangling problems.

packages/bitcore-lib/lib/script/interpreter.js
kajoseph: Yes

packages/bitcore-lib/lib/crypto/signature.js
kajoseph: No. This is a `fromBuffer`/`fromDER` which is a defined encoded format. If all you have are a sequence of bytes, this is the best you can do to assert that it's a Schnorr sig or not. I could add a `isSchnorr` param to the function, but, like my comment says, if you know it's a Schnorr sig then just use .fromSchnorr()

packages/bitcore-lib/test/publickey.js
kajoseph: That's what the general failure does (invalid X)

packages/bitcore-lib/test/publickey.js
kajoseph: invalid X

MichaelAJay (approved):

nitsujlangston (approved):

PR #3322, opened 2022-01-20
multi-currency wallet-derive-address

SethMcGuire:


packages/bitcore-client/bin/multi-sig/wallet-derive-address
```diff
@@ -2,21 +2,32 @@
 
 const fs = require('fs');
 const bitcoreLib = require('bitcore-lib');
+const bitcoreLibCash = require('bitcore-lib-cash');
```
picnicbear, line 5: These requires weren't working for me on a fresh install. I think we need to refer to crypto wallet core instead. 

```
const CWC = require('crypto-wallet-core');
const bitcoreLibs = {
  BTC: CWC.BitcoreLib,
  DOGE: CWC.BitcoreLibDoge,
  LTC: CWC.BitcoreLibLtc,
  BCH: CWC.BitcoreLibCash
};
```

kajoseph (approved):

PR #3333, opened 2022-02-04
Add fiat rates to API for stablecoins

picnicbear:
- add `getRatesForStablecoin` to fiat rate service
- add USD stablecoin case to `getRatesByCoin`
- add `USD_STABLECOINS` to BWS constants


kajoseph (approved):
Nice!

PR #3335, opened 2022-02-14
[BWS] Allow a non-UTXO chain "re-scan"

kajoseph:
This allows for an ETH wallet to hit the `updateWallet` method in bitcore-node to capture missed past ETH events associated with the wallet


escottalexander (approved):
Looks good to me. 👍

nitsujlangston (approved):

Issue #3358, opened 2022-04-27
String deriviation paths cause invalid hardening value due to Lodash call


SmartArray:
We found a flaw with the string typed key derivation function at the following call:

https://github.com/bitpay/bitcore/blob/bed0702ff56ca207a09711fd2e2a2b75374b51ac/packages/bitcore-lib/lib/hdprivatekey.js#L74-L78

`_.every` calls the function `HDPrivateKey.isValidPath` with two args:
1. The first arg is the expected parsed number of the derivation path
2. The second (unexpected) arg is the index of the root array which toggles the hardening parameter of the function (`isValidPath`), which is clearly unintended

An easy fix is to introduce an anonymous function as shown in this PR: https://github.com/bitpay/bitcore/pull/3359

kajoseph:
Thanks for reporting and the PR, @SmartArray. I am looking at this now

kajoseph:
@SmartArray Just so I'm clear, you are not having any issues with the outcome of `isValidPath`, correct? The strict check for `hardened === true` in the line below ensures that passing in the array index via `_.every` doesn't have an effect.

https://github.com/bitpay/bitcore/blob/2694f72ca2b6c8a01d825b733fcc7f318389086d/packages/bitcore-lib/lib/hdprivatekey.js#L81

If this is "just" a code smell issue, then I'm inclined to punt on #3359 in favor of #3350

PR #3359, opened 2022-04-27
Fix key derivation when path is given as string

SmartArray:
As described in this issue: https://github.com/bitpay/bitcore/issues/3358

This PR fixes the derivation issues when trying to create a key with a string-typed derivation path such as `m/0'/1/2/3`


SmartArray:
@kajoseph Thank you for dealing with the issues from my other PR.
Can you please take a look into this one? I think it's quite a serious flaw and this could lead to many wrong implementations out there, if we don't fix it

PR #3360, opened 2022-04-27
Fix: Network Deletion fixed

SmartArray:
This PR prevents `undefined` to remain in the `networkMaps` array of `networks.js` by utilizing `.splice()` instead of `delete` operator.

The narrative of this PR is to be able to remove default networks, and add other networks, in combination with changing the default network.
However, since the implementation of `get()` always returns the first element of the the map.


kajoseph (changes requested):
Hey @SmartArray, thanks for the PR.

This is likely to break current implementations because it will change `networks.get('something_i_removed')` to return an empty array instead of undefined which is inconsistent with `networks.get('something_that_never_existed')` which returns undefined. Thus, the following piece of code would break:

```
networks.remove('customnet');
const myNetwork = networks.get('customnet');
// myNetwork == []
if (!myNetwork) {
  // ...do something
}
```

Having said that, I do agree that deleting the index which leaves behind an undefined value in the index's position and the array with a length of >0 is a bit smelly.

Therefore, I would suggest also changing the line below to remove the length check. Javascript is forgiving enough that giving an out-of-bounds index to an array will only result in `undefined`, which is what we want.
https://github.com/bitpay/bitcore/blob/2694f72ca2b6c8a01d825b733fcc7f318389086d/packages/bitcore-lib/lib/networks.js#L47

kajoseph:
Closing because this has been fixed in a subsequent PR

PR #3369, opened 2022-05-11
Batched calls to getStatusAll now works with token addresses

escottalexander:
There was an issue where the token address was not being passed to the server. That has been fixed.
This also makes it where you can send a single credential with multiple token addresses and it will return an array of responses from getStatusAll for the main wallet and every token you sent.


kajoseph (approved):
ACK

PR #3375, opened 2022-05-31
[FIX] Braze push notifications

gabrielbazan7:


cmgustavo (dismissed):
LGTM!

packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
```diff
@@ -238,22 +295,35 @@ export class PushNotificationsService {
...
+                        //                      logger.debug('Request message: ' + response.statusMessage);
+                        //                      logger.debug('Request body:  ' + response.request.body);
+                      }
+                      next();
+                    });
+
+                    // if external_user_ids is not set use old version firebase token
+                  } else if (notification && notification.to) {
+                    this._makeRequest(notification, (err, response) => {
+                      if (err) logger.error('ERROR:' + err);
```
kajoseph, line 318: I know this isn't originally your code, but cold we take this opportunity to make these error messages a little better?  This one and on line 306. Preferably some strings where we can uniquely identify where the error is originating instead of just `ERROR:`

kajoseph (changes requested):
Looks good - I just have one small request

kajoseph (approved):
Looks good!

PR #3377, opened 2022-06-01
Update circle ci pipeline image

kajoseph:
- Firstly, it updates the pipeline to use a ubuntu 20 image -- which was the main purpose of the PR, but it exposed some bugs which are fixed below.
- Secondly, it makes the `removeNetwork` function consistent across the libs, and also fixes one or two bugs with the network removal (not that anyone ever adds custom networks anyway, so it makes sense whey this hasn't really been addressed)
- Thirdly, it extends the timeout for some long-running tests that would sometimes cause the pipeline to fail
- Lastly, it fixes a bug where an XRP response would evaluate `body.result` before `body == null`...which inevitably throws if `body == null`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ jobs:
     machine:
       image: ubuntu-2004:202201-02
       docker_layer_caching: true
+      resource_class: large
     working_directory: ~/bitcore
     steps:
       - checkout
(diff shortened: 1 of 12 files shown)


Ellajoke (approved):

gabrielbazan7 (approved):
LGTM

PR #3379, opened 2022-06-06
modification for bitcore client not to use watch only address for tra…

rebrahimi1:
…nsaction input


packages/bitcore-client/bin/wallet-send
```diff
@@ -35,13 +35,13 @@ const main = async () => {
     const scale = Math.pow(10, decimals);
     const lastAddress = await wallet.deriveAddress(0);
     const recipients = [{ address: to, amount: amount * scale }];
-    const feerate = await wallet.getNetworkFee();
-    const feeRate = JSON.parse(feerate).feerate || 20 * scale;
+    const feeRateObject = await wallet.getNetworkFee();
+    const feeRate = JSON.parse(feeRateObject).feerate || 20;
```
kylexrose, line 39: Is the scale multiplier now unnecessary in the default feeRate here?

packages/bitcore-client/bin/wallet-send
```diff
@@ -50,11 +50,39 @@ const main = async () => {
...
+        if (utxosValueSum < outputAmount) {
+          utxosValueSum = utxosValueSum + utxo.value;
+          inputCount += 1;
+        }
+      }
+      if (utxosValueSum < outputAmount) {
+        console.error('wallet input was not sufficient to cover the output amount');
+        process.exit(1);
+      } else {
+        const estimatedBytes = txTemp.length / 2 + inputCount * (107 + 1);
```
kylexrose, line 69: For my own edification, where is the 107 + 1 coming from?

packages/bitcore-client/bin/wallet-send
```diff
@@ -50,11 +50,39 @@ const main = async () => {
       from: lastAddress,
       token
     };
-    const tx = await wallet.newTx(params);
+    let tx;
+    const txTemp = await wallet.newTx(params);
```
kylexrose, line 54: So if I'm reading this right, you aren't mutating this tx, why bother with the txTemp variable declaration rather than just give it a new value with line 78. Then you can drop the variable `txTemp` and the else statement altogether.

kylexrose (changes requested):

packages/bitcore-client/bin/wallet-send
rebrahimi1: it's bytes per input signature plus wiggle room byte. check bitpay/lib/opsUtils/bitcoinCash.js

kajoseph:
@rebrahimi1 Can you sync this branch with master? You tests are failing b/c of a fixed issue

PR #3382, opened 2022-06-22
Wallet derive address

rebrahimi1:
deriving addresses with change option

### DIFF
--- a/packages/bitcore-client/bin/wallet-derive-address
+++ b/packages/bitcore-client/bin/wallet-derive-address
@@ -1,0 +1,64 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const program = require('../ts_build/program');
+const { Wallet } = require('../ts_build/wallet');
+const promptly = require('promptly');
+
+program
+  .version(require('../package.json').version)
+  .option('--name <name>', 'REQUIRED - Wallet Name')
+  .option('--change [change]', 'optional - should derive change addresses')
+  .option('--addressIndexes [addressIndexes]', 'optional - takes one or more indexes')
+  .option('--path [path]', 'optional - Custom wallet storage path')
+  .option('--storageType [storageType]', 'optional - name of the database to use (default Level)')
+  .parse(process.argv);
+
+const main = async () => {
+  const { name, path, storageType, addressIndexes } = program;
+  let { change } = program;
+  change = (change === 'true');
+  let wallet;
+  let address;
+  let addressesAndIndex = new Object();
+  let indexes = [];
+  try {
+    const password = await promptly.password('Wallet Password:');
+    wallet = await Wallet.loadWallet({ name, path, storageType });
+    wallet = await wallet.unlock(password);
+    await wallet.register({ baseUrl: wallet.baseUrl });
+
+    if (addressIndexes) {
+      indexes = JSON.parse(addressIndexes).indexes;
+    }
+
+    for (let i = 0; i < indexes.length; i++) {
+      address = wallet.deriveAddress(indexes[i], change);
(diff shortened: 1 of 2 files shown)


packages/bitcore-client/bin/wallet-derive-address
```diff
@@ -0,0 +1,66 @@
...
+    wallet = await wallet.unlock(password);
+    await wallet.register({ baseUrl: wallet.baseUrl });
+
+    if (addressIndexes) {
+      indexes = JSON.parse(addressIndexes).indexes;
+    }
+
+    for (let i = 0; i < indexes.length; i++) {
+      address = wallet.deriveAddress(indexes[i], change);
+      if (address) {
```
kylexrose, line 38: If you're setting the address right before this than would you expect `address` to be defined? -- and if there is a problem deriving one address, is it skipping to line 61?

packages/bitcore-client/bin/wallet-derive-address
```diff
@@ -0,0 +1,66 @@
...
+    const listOfAddresses = Object.values(addressesAndIndex);
+    await wallet.client.importAddresses({ pubKey: wallet.authPubKey, payload: listOfAddresses });
+
+    if (change) {
+      console.log('Change Addresses & Indexes: ', addressesAndIndex);
+    } else {
+      console.log('Addresses & Indexes: ', addressesAndIndex);
+    }
+  } catch (e) {
+    console.error(e);
```
kylexrose, line 62: Maybe a clearer error to the user here/another try block somewhere since there is more code than in the original util.

packages/bitcore-client/bin/wallet-derive-address
```diff
@@ -0,0 +1,66 @@
...
+
+const program = require('../ts_build/program');
+const { Wallet } = require('../ts_build/wallet');
+const promptly = require('promptly');
+
+program
+  .version(require('../package.json').version)
+  .option('--name <name>', 'REQUIRED - Wallet Name')
+  .option('--change [change]', 'optional - should derive change addresses')
+  .option('--addressIndexes [addressIndexes]', 'optional - takes one or more indexes')
```
kylexrose, line 13: I believe the request was for a range (i.e. 4-20) vs an array (i.e. [4, 34, 40, 6]) I'm not sure the use case for deriving from an array of indexes but it's just a thought.

kylexrose (changes requested):

packages/bitcore-client/bin/wallet-derive-address
```diff
@@ -0,0 +1,66 @@
...
+  .option('--path [path]', 'optional - Custom wallet storage path')
+  .option('--storageType [storageType]', 'optional - name of the database to use (default Level)')
+  .parse(process.argv);
+
+const main = async () => {
+  const { name, path, storageType, addressIndexes } = program;
+  let { change } = program;
+  change = (change === 'true');
```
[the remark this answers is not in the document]
rebrahimi1: it looks like destructuring it will not result change to boolean true/false, which has it be boolean. Another way that I can think of is to check if change is defined to avoid the program to crash and do `JSON.parse(change)`.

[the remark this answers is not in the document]
Paulalobe: Can I have anyone interested in buying bitcore

escottalexander:
This has been replaced by https://github.com/bitpay/bitcore/pull/3460. It can be closed.

kajoseph:
replaced by https://github.com/bitpay/bitcore/pull/3460

Paulalobe:
Good day,Do you buy bitcore?

PR #3386, opened 2022-07-01
Reform eth transactions

escottalexander:
Utilize event logs for ERC20 transactions.


tonik-ru:
we also added logs to db in our fork. its impossible to track balances without logs.
but why are you doing getTransactionReceipt for each tx?
this will kill performance, making ~100 additional calls to parity for each block.
also in getPastBlocks you could filter by topics.
and what about erc1155?
also i would suggest adding "loadLogs" in config. because logs are space intensive.

escottalexander:
> 

I appreciate your thoughts. I went ahead and added the topic filter to getPastLogs as you suggested. 

As for using getTransactionReceipt, I agree that it's not optimal but any other option would take a lot of reengineering as the few places we are using it are places where the transactions are being handled individually (like in the event listener section looking for new txs). Anywhere where we have access to all the transaction in the block we are using the full logs for all txs in that block with getPastLogs. Feel free to clarify or submit a PR if you are seeing a way that I am not.

I like your idea about making the logs configurable. We might tackle that at a later date. Feel free to submit a PR.

Currently we are only decoding the logs with a handful of ABIs. I don't think we are even using the ERC721 and Multisig events. I don't see much of a point to adding ERC1155 but if someone needed that functionality then it should be pretty easy for them to add it by following the "template" laid out by the existing ABIs.

4player90 (approved):

4player90 (approved):

PR #3387, opened 2022-07-01
[FIX] getStatusAll - if error continue anyway getting next wallets st…

gabrielbazan7:
…atus


escottalexander (approved):
Looks good. You would know better than me but I wonder if there would ever be a need for it to fail. In that case we could pass our preference as a parameter but if we don't need that then I think this will suffice.

gabrielbazan7:
> Looks good. You would know better than me but I wonder if there would ever be a need for it to fail. In that case we could pass our preference as a parameter but if we don't need that then I think this will suffice.

Adding the error message to the result is enough we can use cached values in the app for that particular wallet. The idea is not to block the entire key if only one fails.

kajoseph (approved):

PR #3390, opened 2022-07-11
[REF] process status

gabrielbazan7:
process status to decrypt creator and copayers name.
might need a refactor to reuse `_processStatus` from api.ts.
might need more unit tests.

Screenshot_11_07_2022_18_49

I already tested in the app


kajoseph (approved):
ACK

PR #3394, opened 2022-07-18
Add polygon to bws, bwc, cwc

escottalexander:
This includes updates to CWC, BWC and BWS to support Matic and ERC20s on Matic.


packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
```diff
@@ -451,16 +451,30 @@ export class PushNotificationsService {
...
       doge: 'DOGE',
       ltc: 'LTC',
-      usdc: 'USDC',
-      pax: 'PAX',
-      gusd: 'GUSD',
-      busd: 'BUSD',
-      wbtc: 'WBTC',
-      dai: 'DAI',
-      shib: 'SHIB'
+      matic_e: 'MATIC',
```
gabrielbazan7, line 458: There is no need for adding new keys here

packages/bitcore-wallet-client/src/lib/common/constants.ts
```diff
@@ -21,23 +21,65 @@ export const Constants = {
     REQUEST_KEY_AUTH: 'm/2' // relative to BASE
   },
   BIP45_SHARED_INDEX: 0x80000000 - 1,
+
```
gabrielbazan7, line 24: I think it's safe to remove the COINS array https://github.com/bitpay/bitcore/pull/3400

packages/bitcore-wallet-service/src/lib/pushnotificationsservice.ts
escottalexander: Good point. I made the adjustment in my latest commit.

packages/bitcore-wallet-client/src/lib/common/constants.ts
escottalexander: When your PR is merged in I will pull in the latest and make revisions to this PR as needed.

escottalexander: I confirmed that the COINS array can be removed and I removed it.

packages/crypto-wallet-core/src/derivation/paths.ts
```diff
@@ -27,6 +27,11 @@ export const Paths = {
     livenet: "m/44'/2'/",
     testnet: "m/44'/2'/"
   },
+  MATIC: {
+    mainnet: "m/44'/966'/",
```
micahriggan, line 31: woah, didn't realize matic and other eth forks have their own derivation path, since metamask lets you switch networks.

Does the BWS side let you use an ethereum wallet on the matic network?

kajoseph: I haven't tested this, but I think you'd have to specify an ETH derivation path when importing from the wallet (from metamask, for example), but otherwise I don't see why that wouldn't be possible.

kajoseph: I wouldn't be surprised if there is a unique index on a pubkey, though. Something to check here

kajoseph (changes requested):

packages/crypto-wallet-core/src/constants/tokens.ts
```diff
@@ -41,16 +41,85 @@ export let TOKEN_OPTS = {
...
     decimal: 18,
+    name: 'SHIBA INU',
+    symbol: 'SHIB'
+  },
+  '0x8a037dbca8134ffc72c362e394e35e0cad618f85': {
     name: 'Euro Coin',
-    symbol: 'EUROC'
+    symbol: 'EUROC',
+    decimal: 6,
+    address: '0x8a037dbca8134ffc72c362e394e35e0cad618f85'
```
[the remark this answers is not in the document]
escottalexander: It's listed here: https://mapper.polygon.technology/ 
Samarthya told me everything on this site should be official. Please double check.

packages/crypto-wallet-core/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "./ts_build/src/index.js",
   "types": "./ts_build/src/index.d.ts",
   "scripts": {
+    "tsc": "npm run clean && tsc",
```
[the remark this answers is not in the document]
escottalexander: Accidentally left it in there after debugging some weirdness in the tests. I will remove.

packages/crypto-wallet-core/src/derivation/paths.ts
```diff
@@ -27,6 +27,11 @@ export const Paths = {
     livenet: "m/44'/2'/",
     testnet: "m/44'/2'/"
   },
+  MATIC: {
+    mainnet: "m/44'/60'/", // actual matic derivation path is 966 but users will expect address to be same as ETH
```
kajoseph, line 31: Maybe we should say "the official matic ..." instead of "actual". "Actual" makes it sound like `m/44'/60'/` will somehow be translated into `m/44'/966'/`

kajoseph (changes requested):
One last nitpick :)

packages/crypto-wallet-core/src/derivation/paths.ts
escottalexander: Fixed in my latest commit.

kajoseph (approved):
ACK

PR #3402, opened 2022-08-01
add ape and euroc

lanchana:


kajoseph (dismissed):
ACK

packages/bitcore-wallet-service/src/lib/common/constants.ts
```diff
@@ -40,8 +40,7 @@ module.exports = {
     DAI: 'dai',
     WBTC: 'wbtc',
     SHIB: 'shib',
-    APE: 'ape',
-    EUROC: 'euroc'
```
kajoseph, line 44: EUROC should stay in the ERC20 list

kajoseph (approved):

PR #3403, opened 2022-08-02
Fix amount BN issue

lanchana:


packages/crypto-wallet-core/src/transactions/erc20/index.ts
```diff
@@ -46,7 +46,7 @@ export class ERC20TxProvider extends ETHTxProvider {
       return multisendContract.methods.sendErc20(tokenAddress, addresses, amounts).encodeABI();
     } else {
       const [{ address, amount }] = params.recipients;
-      const amountStr = Number(amount).toLocaleString('en', { useGrouping: false });
+      const amountStr = toBN(amount);
```
kajoseph, line 49: should rename the var since it's not a str anymore

kajoseph (approved):
ACK

PR #3404, opened 2022-08-03
[REF] simplify adding new chains - bws

gabrielbazan7:
- stop using `coin` when it really means `chain`
- `chain` always lowercase
- REF `COINS`, `ERC20`, `USD_STABLECOINS` and `EUR_STABLECOINS`, constants -> replaced by `BITPAY_SUPPORTED_COINS`, `BITPAY_SUPPORTED_ERC20`, `BITPAY_USD_STABLECOINS` and `BITPAY_EUR_STABLECOINS` ( only used for getting bitpay rates )
- Always get bitcore libs using `chain` -> `Bitcore_[chain]`
- Keep `Utils` `getChain` function only for backwards compatibility
- Add `chain` to : address, copayer and wallet models 
- Use lowercase `chain` in txproposal model

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -68,6 +68,7 @@ let serviceVersion;
 
 interface IAddress {
   coin: string;
+  chain: string;
   network: string;
   address: string;
   hasActivity: boolean;
@@ -465,6 +466,7 @@ export class WalletService {
    * @param {string} opts.pubKey - Public key to verify copayers joining have access to the wallet secret.
    * @param {string} opts.singleAddress[=false] - The wallet will only ever have one address.
    * @param {string} opts.coin[='btc'] - The coin for this wallet (btc, bch, eth, doge, ltc).
+   * @param {string} opts.chain[='btc'] - The chain for this wallet (btc, bch, eth, doge, ltc).
    * @param {string} opts.network[='livenet'] - The Bitcoin network for this wallet.
    * @param {string} opts.account[=0] - BIP44 account number
    * @param {string} opts.usePurpose48 - for Multisig wallet, use purpose=48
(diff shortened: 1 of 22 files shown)


escottalexander (dismissed):
These changes look good. Nice work.

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -589,6 +598,9 @@ export class WalletService {
       // only for testing
       if (opts.doNotMigrate) return cb(null, wallet);
 
+      // backwards compatibility
+      if (!wallet.chain) wallet.chain = ChainService.getChain(wallet.coin);
```
kajoseph, line 602: TODO: we should probably create a migration script to add & populate the `chain` field for every wallet in the db

FYI, some very near future updates that are required for polygon will have breaking changes and require a major version bump. Introducing a breaking change and having a migration script to run pre-upgrade would be appropriate here.
Having said that, what you have here works.

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -3033,7 +3047,7 @@ export class WalletService {
               return txp.status == 'broadcasted';
             });
 
-            if (txps[0] && txps[0].coin == 'bch') {
+            if (txps[0] && txps[0].chain == 'bch') {
```
kajoseph, line 3050: Txp.chain is covered by TxProposal.fromObj()...nice

kajoseph (changes requested):
Overall looks great! Thanks for doing this. Left a few relatively minor comments/requests

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1019,7 +1032,8 @@ export class WalletService {
     if (_.isEmpty(opts.name)) return cb(new ClientError('Invalid copayer name'));
 
     opts.coin = opts.coin || Defaults.COIN;
-    if (!Utils.checkValueInCollection(opts.coin, Constants.COINS)) return cb(new ClientError('Invalid coin'));
+    // checking in chains for simplicity
+    if (!Utils.checkValueInCollection(opts.coin, Constants.CHAINS)) return cb(new ClientError('Invalid coin'));
```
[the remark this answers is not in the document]
gabrielbazan7: Added. I wasn't sure about changing createWallet and joinWallet since chain === coin when we are storing clients. But I decided to refactor both and I forgot about joinWallet

gabrielbazan7:
Done! thanks for reviewing

kajoseph (approved):
Super! :+1:

PR #3405, opened 2022-08-05
goerli added as default testnet

escottalexander:
This PR changes the default ETH testnet from Kovan to Goerli since the former is deprecated.This is likely dependent on changes to front end. @lanchana ?


kajoseph (approved):
ACK

PR #3411, opened 2022-08-15
Fix unspent inputs

escottalexander:
This updates pending inputs of invalid transactions back to an unspent status.

### DIFF
--- a/packages/bitcore-node/migrations/fixUnspentInputs.js
+++ b/packages/bitcore-node/migrations/fixUnspentInputs.js
@@ -1,65 +1,117 @@
+#!/usr/bin/env node
 /****************************************
-*** This migration script will fix historical instances 
-*** of an issue where inputs that were used in an RBF tx
-*** are left in a pending state if they are not used by 
-*** the new tx that is the replacement.
-********************************************/ 
+ *** This migration script will fix historical instances
+ *** of an issue where inputs that were used in an RBF tx
+ *** are left in a pending state if they are not used by
+ *** the new tx that is the replacement.
+ ***
+ *** This does a dry run by default. Use "--dryrun false" 
+ *** to execute outside of dry run.
+ ********************************************/
 
 const { TransactionModel, TransactionStorage } = require('../build/src/models/transaction');
 const { CoinModel, CoinStorage } = require('../build/src/models/coin');
+const fs = require('fs');
+const fsPromises = fs.promises;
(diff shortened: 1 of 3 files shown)


packages/bitcore-node/migrations/fixUnspentInputs.js
```diff
@@ -0,0 +1,65 @@
...
+            if (!Storage.connected) {
+                await Storage.start(storageArgs);
+                await wait(2000);
+              }
+        } catch (e) {
+            console.log(e);
+        }
+    }
+
+    async runScript() {
```
kajoseph, line 39: Let's add a param to do a dry-run that outputs the txids that will be changed. By default, it should do a dry-run and then have a cli param to run it for real

packages/bitcore-node/migrations/fixUnspentInputs.js
```diff
@@ -0,0 +1,65 @@
+/****************************************
```
kajoseph, line 1: Add a shebang to the top of the file `#!/usr/bin/env node` and give this file execution permissions `chmod +x fixUnspentInputs.js`

kajoseph (changes requested):
I think you're on the right track - thanks for doing this. Just some relatively minor things and a few best-practice items

packages/bitcore-node/migrations/fixUnspentInputs.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    async runScript() {
+        console.log('running script');
+            for await (const chain of RBFEnabledUTXOChains) {
+                for await (const network of networks) {
+                    // using Set so that we don't have duplicates
+                    console.log(`updating records for ${chain}:${network}`);
+                    // Get array of all invalid txs for chain and network
+                    const invalidTxs = await this.transactionModel.collection.find({ chain, network, blockHeight: -3 }).toArray();
```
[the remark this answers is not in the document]
kajoseph: Ah ok. Might be good to just make a local const then

mohammadfarari1360 (commented):
Okay

escottalexander:
> I think you're on the right track - thanks for doing this. Just some relatively minor things and a few best-practice items

Thanks for your input. I made the revisions you suggested. Let me know if you see anything else.

kajoseph (approved):
Looks good

mohammadfarari1360 (commented):
Okay good

PR #3425, opened 2022-09-06
[FEAT] emails notifications for erc20

gabrielbazan7:


packages/bitcore-wallet-service/src/lib/emailservice.ts
```diff
@@ -236,24 +246,59 @@ export class EmailService {
...
+          const tokenAddress = data.tokenAddress.toLowerCase();
+          if (Constants.TOKEN_OPTS[tokenAddress]) {
+            unit = Constants.TOKEN_OPTS[tokenAddress].symbol.toLowerCase();
+            label = UNIT_LABELS[unit];
+          } else {
+            let customTokensData;
+            try {
+              customTokensData = await this.getTokenData();
+            } catch (error) {
+              throw new Error('Could not get custom tokens data');
```
kajoseph, line 286: instead of throwing here, let's just `return cb(new Error(...));`

packages/bitcore-wallet-service/src/lib/emailservice.ts
```diff
@@ -236,24 +246,59 @@ export class EmailService {
...
+            if (customTokensData && customTokensData[tokenAddress]) {
+              unit = customTokensData[tokenAddress].symbol.toLowerCase();
+              label = unit.toUpperCase();
+              opts.toSatoshis = 10 ** customTokensData[tokenAddress].decimals;
+              opts.decimals = {
+                maxDecimals: 6,
+                minDecimals: 2
+              };
+            } else {
+              throw new Error('Notifications for unsupported token are not allowed');
```
kajoseph, line 297: same as above, return the cb;

Also it would be good to include the token in the error message, like `'Notifications for unsupported tokens are not allowed: ' + tokenAddress`

kajoseph (changes requested):

gabrielbazan7:
Done! thanks for reviewing

kajoseph (approved):
LGTM

PR #3426, opened 2022-09-12
Fix unspent inputs

escottalexander:
Updated migration script.

It has been converted to a stream to avoid memory over-utilization and it now calls the specified RPC to verify if a coin is spent and updates the record accordingly.

You now must specify the chain and network you want to run it against with --chain and --network flags. This should match the settings in bitcore.config.json.

It does a dryrun by default and copies the results to a log file. To override the dryrun use `--dryrun false`.


packages/bitcore-node/package.json
```diff
@@ -44,7 +44,8 @@
     "lint:style": "prettier --check 'src/**/*.ts'",
     "fix": "npm run fix:style && npm run fix:lint",
     "fix:lint": "tslint --fix -c ../../tslint.json 'src/**/*.ts'",
-    "fix:style": "prettier --write 'src/**/*.ts'"
+    "fix:style": "prettier --write 'src/**/*.ts'",
+    "migration": "cd migrations && node fixUnspentInputs.js --network regtest --dryrun false"
```
kajoseph, line 48: Given that this is a one-time run, we probably don't want to have this as a script

packages/bitcore-node/migrations/fixUnspentInputs.js
```diff
@@ -40,7 +42,9 @@ class Migration {
 
   processArgs(argv) {
     let defaults = {
-      dryrun: true
+      dryrun: true,
+      chain: 'BTC',
+      network: 'testnet'
```
kajoseph, line 47: If I run this with `--help`, it automatically starts it with BTC:testnet. I actually think it would be better if we don't have defaults for `chain` and `network`. In a migration script, forcing the user to be specific isn't a bad thing.

kajoseph (changes requested):
Looks better!

It would also be good to add a log about connecting to the db. When you first start the script, there's a pregnant pause before it outputs anything.

packages/bitcore-node/migrations/fixUnspentInputs.js
```diff
@@ -52,51 +56,106 @@ class Migration {
...
+        if (nextData) {
+          handleStream(nextData);
+        } else {
+          // End of data -
+          endProcess();
+        }
+      }
+    };
+    // Handle incoming stream data
+    handleStream(await stream.next());
```
kajoseph, line 140: need to handle if `await stream.next()` returns null here.

packages/bitcore-node/package.json
escottalexander: Haha my bad. Meant to remove this. Just used for development. I will remove. (gosh, I need to get better about this)

packages/bitcore-node/migrations/fixUnspentInputs.js
escottalexander: Good point. I updated it to have no defaults for chain and network.

packages/bitcore-node/migrations/fixUnspentInputs.js
escottalexander: Nice catch.

kajoseph (approved):

PR #3429, opened 2022-09-14
add check for  undefined value

escottalexander:
Making sure that `this.publicTxUrlTemplate[wallet.chain]` exists before attempting to get wallet.network.

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/emailservice.ts
+++ b/packages/bitcore-wallet-service/src/lib/emailservice.ts
@@ -326,7 +326,7 @@ export class EmailService {
       }
 
       if (_.includes(['NewIncomingTx', 'NewOutgoingTx'], notification.type) && data.txid) {
-        const urlTemplate = this.publicTxUrlTemplate[wallet.chain][wallet.network];
+        const urlTemplate = this.publicTxUrlTemplate[wallet.chain]?.[wallet.network];
         if (urlTemplate) {
           try {
             data.urlForTx = Mustache.render(urlTemplate, data);
--- a/packages/bitcore-wallet-service/src/lib/emailservice.ts
+++ b/packages/bitcore-wallet-service/src/lib/emailservice.ts
@@ -333,6 +333,8 @@ export class EmailService {
           } catch (ex) {
             logger.warn('Could not render public url for tx', ex);
           }
+        } else {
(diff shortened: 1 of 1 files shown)


packages/bitcore-wallet-service/src/lib/emailservice.ts
```diff
@@ -326,7 +326,7 @@ export class EmailService {
       }
 
       if (_.includes(['NewIncomingTx', 'NewOutgoingTx'], notification.type) && data.txid) {
-        const urlTemplate = this.publicTxUrlTemplate[wallet.chain][wallet.network];
+        const urlTemplate = this.publicTxUrlTemplate[wallet.chain]?.[wallet.network];
```
gabrielbazan7, line 329: what about doing it this way ? 

`const chain = wallet.chain || ChainService.getChain(wallet.coin); // getChain -> backwards compatibility
  const urlTemplate = this.publicTxUrlTemplate[chain]?.[wallet.network];`

gabrielbazan7: most wallet don't have the `chain` property

kajoseph: `this.storage.fetchWallet` already populates the chain field like that with `Wallet.fromObj`

gabrielbazan7: ooh ok nice..

gabrielbazan7 (dismissed):
ACK

packages/bitcore-wallet-service/src/lib/emailservice.ts
```diff
@@ -326,7 +326,7 @@ export class EmailService {
...
         if (urlTemplate) {
```
kajoseph, line 330: @gabrielbazan7 brought up a good point offline. Let's add an else after this if and log the chain/network when urlTemplate doesn't exist

kajoseph (changes requested):

packages/bitcore-wallet-service/src/lib/emailservice.ts
escottalexander: Good plan. Added the logging in the latest commit.

kajoseph (approved):
Thanks!

Issue #3430, opened 2022-09-14
Insight does not correctly present transaction b10c007c…792b4d41


Xekyo:
The first transaction that used P2TR outputs (notably before activation of the soft fork), [b10c007c60e14f9d087e0291d4d0c7869697c6681d979c6639dbd960792b4d41](https://bitpay.com/insight/BTC/mainnet/tx/b10c007c60e14f9d087e0291d4d0c7869697c6681d979c6639dbd960792b4d41) is shown to have one input and two outputs, although it actually has four inputs:

![image](https://user-images.githubusercontent.com/4060799/190226649-e00405be-a5f7-4c64-abd8-901f67376568.png)

For reference, [mempool.space](https://mempool.space/tx/b10c007c60e14f9d087e0291d4d0c7869697c6681d979c6639dbd960792b4d41) shows:
![image](https://user-images.githubusercontent.com/4060799/190226876-e0bb582c-7cc1-4bed-80e2-1cad3bcd1c6c.png)

It does look like bech32m addresses and P2TR outputs are correctly indexed starting from activation, as other transactions I've tested (e.g. [3777…35c8](https://bitpay.com/insight/BTC/mainnet/tx/37777defed8717c581b4c0509329550e344bdc14ac38f71fc050096887e535c8) look correct to me.

kajoseph:
Thanks for bringing this to our attention! We will look into this

kajoseph:
@Xekyo This fix has been merged and deployed.

![image](https://user-images.githubusercontent.com/32603939/193641687-7792990b-d4d5-4858-a534-9a95cd35211d.png)

Xekyo:
Hi @kajoseph, I was just looking at the transaction again. It's good that the addresses are now shown, but I notice that the last three inputs are compounded into one line, and other than I expected, even when expanding the details, the three inputs still are shown as one. While the latter three inputs all spend funds associated with the same address, this causes e.g. the arrow in the front to only link to the details of [input 1](https://bitpay.com/insight/BTC/mainnet/tx/b48a59fa9e036e997ba733904f631b1a64f5274be646698e49fd542141ca9404?detailsIdx=1&fromVout=false). 
![image](https://user-images.githubusercontent.com/4060799/193670464-b7cf8327-8f4f-4c89-bef0-8132bd87ef60.png)

I was wondering whether that is intentional, or whether the detailed view should reveal that there were four inputs even when some of them spend funds associated with the same address:
[...]

PR #3433, opened 2022-09-16
Try to parse unknown coin addresses

escottalexander:
This should fix the issue mentioned here: https://github.com/bitpay/bitcore/issues/3430

In the past, some addresses where not parsed correctly because BitcoreLib didn't know how to handle them. This remedy will attempt to parse each unknown address and update the db accordingly.

### DIFF
--- a/packages/bitcore-node/src/models/coin.ts
+++ b/packages/bitcore-node/src/models/coin.ts
@@ -190,7 +190,7 @@ export class CoinModel extends BaseModel<ICoin> {
 
   _apiTransform(coin: Partial<MongoBound<ICoin>>, options?: { object: boolean }): any {
     // try to parse coin.address if its 'false' and script exists
-    if (coin.address == 'false' && coin.script != undefined && coin.script != Buffer.alloc(0)) {
+    if (coin.address == 'false' && coin.script != undefined && coin.script.toString() != '') {
       try {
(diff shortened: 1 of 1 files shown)


packages/bitcore-node/src/models/coin.ts
```diff
@@ -187,6 +189,27 @@ export class CoinModel extends BaseModel<ICoin> {
   }
 
   _apiTransform(coin: Partial<MongoBound<ICoin>>, options?: { object: boolean }): any {
+    // try to parse coin.address if its 'false' and script exists
+    if (coin.address == 'false' && coin.script != undefined && coin.script != Buffer.alloc(0)) {
```
kajoseph, line 193: coin.script will always != Buffer.alloc(0) because they will be different instances of the prototypes. For example, `Buffer.alloc(0) == Buffer.alloc(0)` is false. 

I think this should be `coin.script.toString() != '')`

kajoseph (changes requested):
Thanks for jumping on this

packages/bitcore-node/src/models/coin.ts
escottalexander: Changed in latest commit.

kajoseph (approved):
ACK

PR #3376, opened 2022-05-31
modification for mongo client bug - Bitcore-client

rebrahimi1:

PR #3436, opened 2022-09-20
Test Mongo connection in Bitcore Client before we attempt to use it

escottalexander:
This replaces https://github.com/bitpay/bitcore/pull/3376.

### DIFF
--- a/packages/bitcore-client/src/storage.ts
+++ b/packages/bitcore-client/src/storage.ts
@@ -41,10 +41,21 @@ export class Storage {
     }
   }
 
+  async verifyDbs(dbs) {
+    // test mongo connection
+    const mongoDbIdx = dbs.findIndex(d => d instanceof Mongo);
+    if (!(await (dbs[mongoDbIdx] as Mongo).testConnection())) {
+      // remove mongo from array of dbs
+      dbs.splice(mongoDbIdx, 1);
+    }
+
+    return dbs;
+  }
+
   async loadWallet(params: { name: string }) {
     const { name } = params;
     let wallet;
-    for (let db of this.db) {
+    for (let db of await this.verifyDbs(this.db)) {
       try {
         wallet = await db.loadWallet({ name });
         if (wallet) {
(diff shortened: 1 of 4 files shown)


packages/bitcore-client/src/storage/mongo.ts
```diff
@@ -52,6 +52,16 @@ export class Mongo {
     await this.client.close();
   }
 
+  async testConnection() {
+    try {
+      this.client = new MongoClient(this.path, { reconnectTries: 3 });
```
kajoseph, line 57: `this.client` gets overwritten in the `init` function in this file, which leaves a zombie connection to the db. This results in a command forever hanging in the terminal (e.g. `bin/wallet-list`).

Furthermore, I'm getting a new error about using unified topology. Make sure you're passing the same client opts to MongoClient as the init

packages/bitcore-client/src/storage.ts
```diff
@@ -73,20 +84,22 @@ export class Storage {
 
   async listWallets() {
     let passThrough = new PassThrough();
-    for (let db of this.db) {
+    const dbs = await this.verifyDbs(this.db);
+    for (let db of dbs) {
       const listWalletStream = await db.listWallets();
       passThrough = listWalletStream.pipe(passThrough, { end: false });
-      listWalletStream.once('end', () => this.db.length-- === 0 && passThrough.end());
+      listWalletStream.once('end', () => dbs.length-- === 0 && passThrough.end());
```
kajoseph, line 91: this is a typo as `dbs.length === 0` will be evaluated before decrementing. It should be `--dbs.length` as seen below in `listKeys()`

packages/bitcore-client/src/storage.ts
```diff
@@ -41,10 +41,21 @@ export class Storage {
     }
   }
 
+  async verifyDbs(dbs) {
+    // test mongo connection
+    const mongoDbIdx = dbs.findIndex(d => d instanceof Mongo);
+    if (!(await (dbs[mongoDbIdx] as Mongo).testConnection())) {
```
kajoseph, line 47: A cleaner way of doing this would be to loop over every val in dbs and either: 
- add a `testConnection` to the other storage type classes (they could be stubbed to always return true) 
OR
- do `if (typeof db.testConnection === 'function') {...`

kajoseph (changes requested):

kajoseph (approved):
ACK

kajoseph (approved):

PR #3438, opened 2022-09-26
Fix: convert a scientific notation to Number if necessary

cmgustavo:
For some reason, BWS receives `outputs.amount` as string if it's a scientific notation (big number like 1e+21). 

This PR check if it's a string and try to convert to `Number`.

It doesn't happen on these versions of node running my own BWS local: v12.22.8 and v16.15.1. 

Error

```
{"message":"Client Err: 400 /v3/txproposals/ {\"code\":\"BADREQUEST\",\"message\":\"Invalid amount\",\"name\":\"BADREQUEST\"}","level":"info"}
```

Only happens on production.


cmgustavo:
Closed in favor of https://github.com/bitpay/bitcore/pull/3441

PR #3440, opened 2022-09-26
Fix: chain name for create TX

cmgustavo:


gabrielbazan7 (dismissed):

packages/bitcore-wallet-client/src/lib/common/utils.ts
```diff
@@ -484,9 +484,9 @@ export class Utils {
       const isMULTISIG = multisigContractAddress;
       const chainName = chain.toUpperCase();
       const _chain = isMULTISIG
-        ? chainName + 'MULTISIG'
+        ? chainName + 'ETHMULTISIG'
```
escottalexander, line 487: Not sure what the issue is that is requiring this change but the way this is supposed to work is 
chainName >> 'ETH'
so `chainName + 'MULTISIG'` would become 'ETHMULTISIG'. The way this is written I believe it will become 'ETHETHMULTISIG'. The reason this logic was added was to account for other chains such as MATIC.

I wonder if there is a better solution to the issue you are solving for.

escottalexander (changes requested):

packages/bitcore-wallet-client/src/lib/common/utils.ts
cmgustavo: Oh, I did not realize. It was another issue. 
Thanks for taking a look!

cmgustavo: I'll close this PR.

cmgustavo:
Closing this due to I didn't realize it was a redundant mistake (didn't realize of `chainName` value)

PR #3441, opened 2022-09-28
Fix: validate TX amount if BN for ETH only

cmgustavo:
This PR fix issue when sending amount with more than 18 digits in BN format.

https://github.com/bitpay/bitpay-app/pull/425

```
{"message":"Client Err: 400 /v3/txproposals/ {\"code\":\"BADREQUEST\",\"message\":\"Invalid amount\",\"name\":\"BADREQUEST\"}","level":"info"}
```


kajoseph (dismissed):
ACK. Tested and works. Thanks @cmgustavo

kajoseph (approved):

PR #3444, opened 2022-10-04
[REF] supported matic tokens updates

rastajpa:
updates are based on this list:
Captura de Pantalla 2022-10-04 a la(s) 15 11 05

### DIFF
--- a/packages/bitcore-wallet-client/src/lib/common/constants.ts
+++ b/packages/bitcore-wallet-client/src/lib/common/constants.ts
@@ -55,16 +55,12 @@ export const Constants = {
   ],
 
   BITPAY_SUPPORTED_MATIC_ERC20: [
-    'eth_m',
     'usdc_m',
-    'pax_m',
-    'gusd_m',
     'busd_m',
     'dai_m',
     'wbtc_m',
     'shib_m',
-    'ape_m',
-    'euroc_m'
+    'ape_m'
   ],
 
   CHAINS: ['btc', 'bch', 'eth', 'matic', 'xrp', 'doge', 'ltc'],
--- a/packages/crypto-wallet-core/src/constants/tokens.ts
+++ b/packages/crypto-wallet-core/src/constants/tokens.ts
@@ -2,124 +2,100 @@ export let ETH_TOKEN_OPTS = {
   '0x4fabb145d64652a948d72533023f6e7a623c7c53': {
     name: 'Binance USD Coin',
     symbol: 'BUSD',
-    decimal: 18,
+    decimals: 18,
     address: '0x4fabb145d64652a948d72533023f6e7a623c7c53'
   },
   '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48': {
     name: 'USD Coin',
     symbol: 'USDC',
-    decimal: 6,
+    decimals: 6,
     address: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
   },
   '0x8e870d67f660d95d5be530380d0ec0bd388289e1': {
     name: 'Paxos Dollar',
     symbol: 'USDP',
-    decimal: 18,
+    decimals: 18,
     address: '0x8e870d67f660d95d5be530380d0ec0bd388289e1'
   },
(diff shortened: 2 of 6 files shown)


packages/bitcore-wallet-client/src/lib/common/constants.ts
```diff
@@ -55,16 +55,13 @@ export const Constants = {
...
     'usdc_m',
-    'pax_m',
-    'gusd_m',
     'busd_m',
     'dai_m',
     'wbtc_m',
     'shib_m',
     'ape_m',
-    'euroc_m'
+    'matic_m'
```
escottalexander, line 64: I guess this is wrapped MATIC?

packages/crypto-wallet-core/src/constants/tokens.ts
```diff
@@ -32,7 +32,7 @@ export let ETH_TOKEN_OPTS = {
   '0x2260fac5e5542a773aa44fbcfedf7c193bc2c599': {
     name: 'Wrapped Bitcoin',
     symbol: 'WBTC',
-    decimal: 9,
+    decimal: 8,
```
escottalexander, line 35: Good catch on the decimal number. Shouldn't this be `decimals` per your other changes?

escottalexander (changes requested):
These changes look good to me except for the one thing I mentioned.

packages/bitcore-wallet-client/src/lib/common/constants.ts
rastajpa: MATIC Token is its name

packages/crypto-wallet-core/src/constants/tokens.ts
rastajpa: yes, you're right! ty!

packages/bitcore-wallet-client/src/lib/common/constants.ts
escottalexander: In that case I don't think this is needed. MATIC is the base asset on the Polygon chain similar to how ETH is the base asset on the Ethereum chain. We don't need to add it as a token here.

packages/crypto-wallet-core/src/constants/tokens.ts
```diff
@@ -2,124 +2,106 @@ export let ETH_TOKEN_OPTS = {
...
-  '0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619': {
-    name: 'Wrapped Ethereum',
-    symbol: 'ETH',
-    decimal: 18,
-    address: '0x7ceB23fD6bC0adD59E62ac25578270cFf1b9f619'
+  '0x0000000000000000000000000000000000001010': {
+    name: 'Matic Token',
+    symbol: 'MATIC',
+    decimals: 18,
+    address: '0x0000000000000000000000000000000000001010'
```
escottalexander, line 105: Per my above comment we don't need this either.

packages/bitcore-wallet-client/src/lib/common/constants.ts
rastajpa: in the pr description image is there a list of all the tokens that we support, `matic_m` is simply a token in the polygon chain called "MATIC Token" different from the MATIC coin that would be the equivalent of ETH in its own chain

escottalexander (approved):
These look good to me.

rastajpa:
> These look good to me.

I apreciate it. I did a master rebase to see if the checks pass

rastajpa:
@escottalexander I wanted to let you know that I temporarily disabled the Matic token as you suggested earlier because this was only causing errors. For example, in the app when adding that token we obtained the same balance as in the Matic coin, all this speaking only of the Polygon network... a bit confusing, it is also very difficult to differentiate the token from the coin because the coin and chain parameters they are exactly the same, the only difference is the tokenaddress in the token.
if in the future it is necessary to add this confusing token we will see how to solve it from the bitcore side and from the app side

kajoseph (approved):

kajoseph (dismissed):

PR #3449, opened 2022-10-05
[FIX] missing coins in rates list

gabrielbazan7:
Ape and Shib rates are missing in the app.

`BITPAY_SUPPORTED_COINS` constant should include all coins from `fiatrateservice.ts` line 60 list ( those are the actual coins we store using bitpay rates endpoint )

`const coins = ['btc', 'bch', 'eth', 'matic', 'xrp', 'doge', 'ltc', 'shib', 'ape'];`


kajoseph:
These are already in the array as shib_e, ape_e, shib_m, and ape_m. How does adding these to the array fix your rate pulling?

kajoseph (approved):
ACK

kajoseph (approved):

PR #3451, opened 2022-10-10
Bug fix for non-archive node block syncing

escottalexander:
This logs a debug-level error when we get the 'The method debug_traceBlockByNumber does not exist/is not available' error from a non-archive node.

It returns a noisier error if the method fails for some other reason.

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/evm/p2p/rpcs/gethRpc.ts
+++ b/packages/bitcore-node/src/providers/chain-state/evm/p2p/rpcs/gethRpc.ts
@@ -36,7 +36,7 @@ export class GethRPC implements IRpc {
   }
 
   public getBlock(blockNumber: number): Promise<GethBlock> {
-    return this.web3.eth.getBlock(blockNumber, true) as unknown as Promise<GethBlock>;
+    return (this.web3.eth.getBlock(blockNumber, true) as unknown) as Promise<GethBlock>;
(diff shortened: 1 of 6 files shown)


packages/bitcore-node/src/providers/chain-state/evm/p2p/rpcs/gethRpc.ts
```diff
@@ -44,20 +45,29 @@ export class GethRPC implements IRpc {
       params: [this.web3.utils.toHex(blockNumber), { tracer: 'callTracer' }],
       jsonrpc: '2.0',
       id: 1
+    }).catch(err => {
```
kajoseph, line 48: Let's put this in a try..catch instead of doing a .catch here. For consistency, I'd like to stick to one pattern of error handling. The only time we'd use a .catch is if we aren't awaiting the promise.

kajoseph (changes requested):

packages/bitcore-node/src/providers/chain-state/evm/p2p/rpcs/gethRpc.ts
escottalexander: I have corrected this. Thanks for the input.

kajoseph (approved):

kajoseph (approved):

PR #3456, opened 2022-10-13
[REF] default fee matic

gabrielbazan7:
Since https://api-matic.bitcore.io/api/MATIC/mainnet/fee/1 is not currently working, we need a more reliable default fee. Numbers are base on this https://polygonscan.com/chart/gasprice


escottalexander (approved):
This seems good to me.

JohnathanWhite (approved):

PR #3458, opened 2022-10-17
Incorrect spentHeight UTXO migration script updates

escottalexander:
Added logging when rpc call fails for any reason other than the output not being found.

Also protocol defaults to 'http' when it isn't defined in config. This avoids a known error when calling the rpc from the script.


kajoseph (approved):

curtcurt871:
2996210c23eff794bf87d2d93f4e3497096b6e41

PR #3460, opened 2022-10-18
Wallet derive address

escottalexander:
Made some tweaks to Reza's PR so this replaces https://github.com/bitpay/bitcore/pull/3382.

@kylexrose It looks like you were the one reviewing his PR, perhaps you could review this one.


kajoseph:
This appears to be duplicating `derive` with `derive-address` to some extent. It's not clear what problem this is trying to fix, but I think the existing `derive` util could be reworked to be better and probable address whatever this is trying to accomplish.

PR #3461, opened 2022-10-18
Fix/matic erc20

escottalexander:
@gabrielbazan7 pointed out that the way we were inheriting the ERC20TxProvider in the MATICERC20Provider was giving us the wrong chain id as ERC20TxProvider is inheriting from ETHTxProvider. The problem was the same with the Multisig class for Matic. The changes here should remedy the issue.
@kajoseph

### DIFF
--- a/packages/bitcore-client/src/wallet.ts
+++ b/packages/bitcore-client/src/wallet.ts
@@ -328,7 +328,7 @@ export class Wallet {
     gasPrice?: number;
     contractAddress?: string;
   }) {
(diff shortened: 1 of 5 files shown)


gabrielbazan7 (approved):
ACK! tested in the app 
https://polygonscan.com/tx/0x60ba09868979c8c9ae3753acf2b1c4db248d8c8d69ee35fcd5d0345ec6fe7aa0

kajoseph (approved):

PR #3477, opened 2022-11-02
BCMonitor Updates

escottalexander:
Removed some reliance on Lodash.
Refactored fetchActiveTxConfirmationSubs to use a stream instead of callbacks. 
Also added txid to index to avoid lots of in-memory processing.
Added tests for new block notification throttling and set defaults for each chain. MATIC at 12 seconds < please advise if less time is favored.


kajoseph (approved):
tACK

PR #3481, opened 2022-11-08
Insight fix

escottalexander:
Added a check based on the spentHeight to correctly show an unspent tx as unspent. Currently the way Bitcore Node handles coins that are spent by an invalid transaction is by moving them back to an unspent status (spentHeight = -2) but it does not remove the spentTxid which points to the invalid tx. Bitcore Node may be reworked at some point to better handle this event.

@lanchana Let me know if this solution may be problematic.


twiztid89 (approved):

kajoseph (approved):

kajoseph:
tACK

PR #3482, opened 2022-11-08
Coin integration tests

escottalexander:
Added a couple tests to cover coins that are related to RBFed transactions.

### DIFF
--- a/packages/bitcore-node/test/integration/models/coin.spec.ts
+++ b/packages/bitcore-node/test/integration/models/coin.spec.ts
@@ -1,7 +1,230 @@
-import {expect} from 'chai';
+import { expect } from 'chai';
+import * as crypto from 'crypto';
+import { BitcoreLib } from 'crypto-wallet-core';
+import { CoinStorage, ICoin } from '../../../src/models/coin';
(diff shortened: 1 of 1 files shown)


packages/bitcore-node/test/integration/models/coin.spec.ts
```diff
@@ -1,7 +1,230 @@
...
+        mintHeight: -1,
+        mintTxid: mempoolTxid,
+        mintIndex: i,
+        spentHeight: SpentHeightIndicators.unspent
+      } as ICoin;
+    });
+    await addTx(mempoolTx, mempoolOutputs);
+    
+    // update existing outputs to be spent by mempool tx
+    CoinStorage.collection.updateMany(
```
kajoseph, line 188: Need to await?

kajoseph (changes requested):

packages/bitcore-node/test/integration/models/coin.spec.ts
escottalexander: Good catch. Added await in the latest commit.

kajoseph (approved):

PR #3490, opened 2022-11-15
[Bitcore-Node] Standardize EVM Traces

escottalexander:
This PR standardizes the way we record transaction trace data in Bitcore DB. Instead of handling it two ways (seperate Erigon and Geth classes) we transform the Erigon traces to be like Geth. 

There are three parts to this:
1. Now when Erigon blocks are synced they are transformed to be identical to blocks synced from Geth (the only difference was the way we were storing `internal` for Erigon and `calls` for Geth)
2. Added a couple places where we handle old db entries to transform them before continuing ( if we ever chose to update old entries then we could remove these places, they have been commented on accordingly)
3. Removed several places where we had chunks of code handling Erigon and then handling Geth, now it just expects the Geth classes to be the standard


escottalexander:
Closing since these changes are very out of date and will be managed a different way with new chain additions.

PR #3491, opened 2022-11-16
Insight: Remove output selected pill

lanchana:


kajoseph (changes requested):
This doesn't appear to show the `Selected` pill on any inputs. 

image
image

This should be easily reproducible on any tx w/ spent outputs, but the tx shown above is 7641c08f4bd299abfef26dcc6b477938f4a6c2eed2f224d1f5c1c86b4e09739d

kajoseph:
I think we'll need another query param like `voutIndex`

If you click on an **output** -> then it should go to the txid with `?fromVout=true&voutIndex=<output index that was clicked>`
If you click on an **input**   -> then it should go to the txid with `?fromVout=false&voutIndex=<mintIndex>`

kajoseph:
I also noticed that erroneous query params show up when I click on a txid link

image
image

kajoseph (approved):
tACK

Thanks!

PR #3493, opened 2022-11-18
[Bitcore-Node][BWS] Fix for wallet balance issue

escottalexander:
It now checks spentHeight instead of spentTxid in BWS.
In Bitcore-Node when a mempool tx is invalidated, in addition to resetting the coin to unspent status we set the spentTxid to ''.


kajoseph (approved):
tACK

PR #3495, opened 2022-11-18
Bws internal fix

escottalexander:
This checks if internal array exists before using it. Same with the calls array. There does exist a way for this to have neither if blocks were synced by non-archive node.

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
+++ b/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
@@ -468,8 +468,8 @@ export class EthChain implements IChain {
       amount = tx.abiType.params[1].value;
     } else if (tx.abiType && tx.abiType.type === 'MULTISIG' && tx.abiType.name === 'confirmTransaction') {
       multisigContractAddress = tx.to;
-      address = Web3.utils.toChecksumAddress(tx.internal[0].action.to);
-      amount = tx.internal[0].action.value;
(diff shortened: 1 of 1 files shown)


kajoseph (approved):

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -467,9 +467,16 @@ export class EthChain implements IChain {
       address = Web3.utils.toChecksumAddress(tx.abiType.params[0].value);
       amount = tx.abiType.params[1].value;
     } else if (tx.abiType && tx.abiType.type === 'MULTISIG' && tx.abiType.name === 'confirmTransaction') {
+      let amount = 0,
```
kajoseph, line 470: amount and address are being redefined here, and vscode is saying these definition vars are unused...

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -467,9 +467,16 @@ export class EthChain implements IChain {
...
+        address = '0x0';
       multisigContractAddress = tx.to;
-      address = Web3.utils.toChecksumAddress(tx.internal[0].action.to);
-      amount = tx.internal[0].action.value;
+      if (tx.internal) {
```
kajoseph, line 473: I believe tx.internal is a pre-set empty array, so this is truthy. Should be `if (tx.internal && tx.internal.length > 0)`

kajoseph (changes requested):

kajoseph:
@escottalexander Just needed a re-run. :shrug:

kajoseph (approved):

PR #3499, opened 2022-11-28
[Bitcore-Node] ETH Syncing Updates

escottalexander:
This fixes a couple issues found in the multi-threaded ETH syncing functionality.

1. A index was not being used properly so the mongo instance threw `MongoError: errmsg: "Sort operation used more than the maximum 33554432 bytes of RAM. Add an index, or specify a smaller limit."`. This would force the sync process to exit without performing the verification step. If later restarted without multithreaded syncing enabled, it would begin to rollback ~100 blocks at a time because periodic blocks lacked the nextBlockHash property (the getBlockSyncGaps method is what throws the mongo error and later in that method it sets the nextBlockHash).
2. During the multithreaded sync, after all the blocks had been obtained, when any block gaps were returned it would not reliably sync the returned blocks (due to a race condition?)

I also added some logging around the process of filling block gaps.

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/evm/models/block.ts
+++ b/packages/bitcore-node/src/providers/chain-state/evm/models/block.ts
@@ -189,62 +189,47 @@ export class EVMBlockModel extends BaseBlock<IEVMBlock> {
     return new Promise(async (resolve, reject) => {
       let timeout;
       try {
-        const maxBlock = await self.collection.findOne(
-          {
-            chain,
-            network,
-            // height and processed are here to ensure the `chain_1_network_1_processed_1_height_-1` index is hit
-            height: { $exists: true },
-            $or: [{ processed: { $exists: true } }, { processed: { $exists: false } }]
-          },
-          {
-            // height all that matters in the sort. The rest is there to ensure the `chain_1_network_1_processed_1_height_-1` index is hit
-            sort: { height: -1, chain: 1, network: 1, processed: 1 },
-            projection: { height: 1 }
-          }
-        );
-        if (!maxBlock) {
-          return resolve([]);
-        }
-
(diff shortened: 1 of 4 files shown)


packages/bitcore-node/src/providers/chain-state/evm/models/block.ts
```diff
@@ -189,62 +189,47 @@ export class EVMBlockModel extends BaseBlock<IEVMBlock> {
...
         let block = (await stream.next()) as IEVMBlock;
+        const maxBlock = block;
+        if (!maxBlock) {
+          return resolve([]);
+        }
+        const maxHeight = maxBlock.height;
         let prevBlock: IEVMBlock | undefined;
         const outOfSync: number[] = [];
         timeout = setInterval(
           () => logger.info(`${chain}:${network} Block verification height: ${block.height}`),
```
kajoseph, line 210: Since block.height is descending now, let's change this to:
```
        timeout = setInterval(
          () => logger.info(`${chain}:${network} Block verification progress: ${((maxHeight - block.height) / (maxHeight - startHeight) * 100).toFixed(1)}%`),
          1000 * 2
        );
```

kajoseph (changes requested):
Looks good and fixed it for me.

One requested change

packages/bitcore-node/src/providers/chain-state/evm/models/block.ts
escottalexander: Good idea. I implemented that.

Zxvbnm111 (changes requested):

kajoseph (approved):
tACK

WojasKrk (approved):

PR #3507, opened 2022-12-27
[Bitcore-Node] Fix Unspent Coins with spentTxid

escottalexander:
This PR adds a new migration script that checks all unspent coins that have a spentTxid and sets the spentTxid to an empty string. It also checks to verify that the coin is unspent before updating.

This script is mostly a copy of the original (now renamed) migration script.

Location of changes are: 

- (99) Where the stream query is defined
- (136) Catch for when a coin is actually spent - updated to log any records that catch here as this is not expected
- (179) Write to file if any "actually spent" coins showed up from our query

Also updated the original migration script to match a uniform naming scheme for the migration scripts and to set the spentTxid when it updates a record. This would be useful to anyone who runs the script in the future though unnecessary if the second migration script is used.

### DIFF
--- a/packages/bitcore-node/scripts/fixOldSpentTxid.js
+++ b/packages/bitcore-node/scripts/fixOldSpentTxid.js
@@ -1,9 +1,9 @@
 #!/usr/bin/env node
 /****************************************
  *** This migration script will fix historical instances
- *** of an issue where inputs that were used in an RBF tx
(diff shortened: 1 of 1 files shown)


kajoseph (approved):

u353y:
<details><summary>Details</summary>
<p>


Trezor
</p>
</details>

PR #3510, opened 2023-01-04
Feat: disable Wyre

Gamboster:
This change is required to temporarily disable Wyre from the Bitpay app.
The "message" attribute will be displayed to users in the app, and they will not be able to continue the purchase process with this particular exchange.
In the future we will implement a remote config to disable/remove this type of integration directly from the config file, without having to wait for app releases.


kajoseph (approved):
ACK

PR #3512, opened 2023-01-09
Fix: Wyre disabled message

Gamboster:


Gamboster:
It was recently decided that the Wyre integration will continue to work in our app, since it will continue to provide its services

PR #3514, opened 2023-01-16
Revert "Disable Wyre"

Gamboster:
It was recently decided that the Wyre integration will continue to work in our app, since it will continue to provide its services


kajoseph (approved):
ACK

PR #3516, opened 2023-01-20
Update README.md with the correct link to Bitpay wallet

lateminer:
The link to Bitpay wallet in README.md points to the old Copay, not Bitpay Wallet. This PR fixes that.


kajoseph (approved):
ACK

PR #3540, opened 2023-02-14
[BWS] Make common.ts a module

kajoseph:
This makes the common.ts a module which allows for a much better dev experience to be able to easily reference commonly used constants, util functions, etc.

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -88,11 +88,12 @@ export interface IWalletService {
   walletId: string;
   copayerId: string;
   appName: string;
-  appVersion: string;
-  parsedClientVersion: { agent: number; major: number; minor: number };
+  appVersion: { agent?: string; major?: number; minor?: number };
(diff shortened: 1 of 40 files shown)


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -113,8 +113,8 @@ export class WalletService {
   walletId: string;
   copayerId: string;
   appName: string;
-  appVersion: string;
-  parsedClientVersion: { agent: string; major: number; minor: number };
+  appVersion: { agent?: string; major?: number; minor?: number };
+  parsedClientVersion: { agent?: string; major?: number; minor?: number };
```
escottalexander, line 117: Should the IWalletService interface on line 79 be updated as well?

escottalexander (approved):
Just left a small question on one of the changes.

packages/bitcore-wallet-service/src/lib/server.ts
kajoseph: Updated. It doesn't look like IWalletService is used anywhere...so I added an `implements` to the class.

escottalexander (approved):

PR #3543, opened 2023-02-17
Ref: Changelly block NY users

Gamboster:
Workaround to block users from NY. This is a temporary solution until we can migrate to Changelly API v2


Gamboster:
I close this PR since I implemented a better solution.

PR #3551, opened 2023-03-07
Ref: Simplex get keys web support

Gamboster:


JohnathanWhite (approved):

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -4865,11 +4865,22 @@ export class WalletService {
...
+        env = 'productionWeb';
+      } else {
+        env = 'sandboxWeb';
+      }
+    } else {
+      if (req.body.env && req.body.env === 'production') {
+        env = 'production';
+      } else {
+        env = 'sandbox';
+      }
```
kajoseph, line 4880: This can be simplified into:

```
    // production | sandbox | productionWeb | sandboxWeb
    let env = req.body.env === 'production' ? 'production' : 'sandbox';
    if (req.body.context === 'web') {
      env += 'Web';
    }
```

kajoseph (changes requested):
It works :+1: , could just be cleaned up a bit

PR #3553, opened 2023-03-08
EVM nonce management

leolambo:
EVM nonce is set at transaction signing unless it was specified during transaction creation.
A transactions nonce is equal to the last used nonce for a given address plus the number of signed unbroadcasted transactions for the given address.


packages/bitcore-wallet-client/src/lib/common/utils.ts
```diff
@@ -505,7 +505,7 @@ export class Utils {
             ...recipients[index],
             tag: destinationTag ? Number(destinationTag) : undefined,
             chain: _chain,
-            nonce: Number(txp.nonce) + Number(index),
+            nonce: txp.nonce ? Number(txp.nonce) + Number(index): null,
```
kajoseph, line 508: This and above should be `txp.nonce != null ? ...` so that if txp.nonce == 0 it doesn't set it to null.

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -264,7 +283,7 @@ export class EthChain implements IChain {
       for (let index = 0; index < recipients.length; index++) {
         let params = {
           ...recipients[index],
-          nonce: Number(txp.nonce) + Number(index),
+          nonce: txp.nonce ? Number(txp.nonce) + Number(index) : null,
```
kajoseph, line 286: This and above should be `txp.nonce != null ? ...` so that if txp.nonce == 0 it doesn't set it to null.

packages/bitcore-wallet-service/src/lib/chain/matic/index.ts
```diff
@@ -39,7 +39,7 @@ export class MaticChain extends EthChain {
         ...txp,
         ...recipients[index],
         chain,
-        nonce: Number(txp.nonce) + Number(index),
+        nonce: txp.nonce ? Number(txp.nonce) + Number(index) : null,
```
kajoseph, line 42: txp.nonce != null

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2419,7 +2419,8 @@ export class WalletService {
                   next();
                 },
                 async next => {
-                  if (!opts.nonce) {
+                  // If an EVM TX nonce is not given, we will set it during TX signing
+                  if (!opts.nonce && !ChainService.isEVMChain(wallet.chain)) {
```
kajoseph, line 2423: opts.nonce != null

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2804,78 +2805,96 @@ export class WalletService {
...
+        });
+        if (action) return cb(Errors.COPAYER_VOTED);
+        if (!txp.isPending()) return cb(Errors.TX_NOT_PENDING);
 
-          const copayer = wallet.getCopayer(this.copayerId);
+        if (txp.signingMethod === 'schnorr' && !opts.supportBchSchnorr) return cb(Errors.UPGRADE_NEEDED);
+
+        const copayer = wallet.getCopayer(this.copayerId);
 
+        if (!txp.nonce && ChainService.isEVMChain(wallet.chain)) {
```
kajoseph, line 2845: txp.nonce != null

kajoseph (changes requested):

leolambo:
bitpay wallet changes that uses setNonce: https://github.com/leolambo/bitpay-app/pull/1/files

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -388,6 +388,15 @@ export class WalletService implements IWalletService {
 
     this.lock.runLocked(this.walletId, { waitTime }, cb, task);
   }
+
+  _releaseLock(lockName){
+    return new Promise((resolve, reject) => {
+      this.lock.releaseLock(lockName, err => {
```
kajoseph, line 394: this.lock.releaseLock is undefined

kajoseph (changes requested):

PR #3554, opened 2023-03-09
Ref: Changelly to v2

Gamboster:
As the first step for this migration, I made a refactor that will only use the `V2` credentials if we pass the `useV2` parameter. In this way users will continue to use `V1` until we perform the necessary tests. Once we are sure that `V2` works correctly, I'll make the necessary changes so that `V2` is usable by everyone.

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -30,6 +30,7 @@ import {
 import { Storage } from './storage';
 
 const config = require('../config');
+const crypto = require('crypto');
 const Uuid = require('uuid');
 const $ = require('preconditions').singleton();
 const deprecatedServerMessage = require('../deprecated-serverMessages');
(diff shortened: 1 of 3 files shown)


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -5153,9 +5178,39 @@ export class WalletService {
...
 
+
+  changellySignRequestsV2(message, secret: string) {
+    if (!message || !secret) throw new Error('Missing parameters to sign Changelly v2 request');
+
+    const privateKey = crypto.createPrivateKey({
+      key: secret,
+      format: 'der',
+      type: 'pkcs8',
+      encoding: 'hex'
```
kajoseph, line 5189: I don't think encoding is a valid param (anymore? maybe removed in node v18). For hex, it should be `key: Buffer.from(secret, 'hex'),`

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -5153,9 +5178,39 @@ export class WalletService {
...
+    
+    const publicKey = crypto.createPublicKey(privateKey).export({
+        type: 'pkcs1',
+        format: 'der'
+    });
+
+    const signature = crypto.sign('sha256', Buffer.from(JSON.stringify(message)), {
+      key: privateKey,
+      type: 'pkcs8',
+      format: 'der'
```
kajoseph, line 5200: This key object is unnecessary. You can just pass privateKey by itself:
```
crypto.sign('sha256', ...., privateKey);
```

kajoseph (changes requested):
Small changes requested. Overall looks good. :+1: 

It doesn't look like we use `apiKeyBase64` or `publicKey` in the v2 credentials. Instead, it looks like you just use the `secret` to derive the apiKey and pub key. Not really an action item. I just thought it was odd that they're in the config.

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -28,6 +28,7 @@ import {
 import { Storage } from './storage';
 
 const config = require('../config');
+const crypto = require('crypto');
```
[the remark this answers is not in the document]
Gamboster: Done

packages/bitcore-wallet-service/src/lib/server.ts
Gamboster: Done! Ty!

packages/bitcore-wallet-service/src/lib/server.ts
Gamboster: I changed it to use Buffer. Thank you!

PR #3556, opened 2023-03-13
Process evm txs

escottalexander:
This introduces some changes to the tx stream in Bitcore-Node primarily for its use in the evm export util. 

- It now checks for refunding internal txs on every tx
- We now track ERC20 transfers in their own property in the response so that they can be easily parsed


escottalexander:
Closing this as we adjusted the script so that we don't need this PR.

PR #3558, opened 2023-03-17
Ref: Swap banned US States added

Gamboster:


packages/bitcore-wallet-service/test/integration/server.js
```diff
@@ -10757,19 +10757,22 @@ describe('Wallet service', function() {
...
-            currentLocationState: 'NY',
-          };
-    
-          server.getServicesData(opts, (err, config) => {
-            should.not.exist(err);
-            should.exist(config.swapCrypto);
-            config.swapCrypto.disabled.should.equal(true);
-            config.swapCrypto.disabledMessage.should.equal('Swaps are currently unavailable in your area.');
+        const usaBannedStates = ['HI', 'LA', 'NY'];
+        usaBannedStates.forEach(bannedState => {
```
kajoseph, line 10761: Please use `for (const bannedState in usaBannedStates) {` instead of .forEach. forEach is slow and a for loop is easier to step through with a debugger

kajoseph (changes requested):

kajoseph (approved):

PR #3570, opened 2023-04-05
Prevent signing EVM transctions with incorrect nonce 

leolambo:


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2848,6 +2848,15 @@ export class WalletService implements IWalletService {
           if (!txp.isPending()) return cb(Errors.TX_NOT_PENDING);
 
           if (txp.signingMethod === 'schnorr' && !opts.supportBchSchnorr) return cb(Errors.UPGRADE_NEEDED);
+          
+          if (Constants.EVM_CHAINS[wallet.chain.toUpperCase()]) {
+            try {
+              const nonce = await this.getNonce({address : txp.from, chain: wallet.chain, network: wallet.network});
```
kajoseph, line 2854: lil bit o lint here

kajoseph (changes requested):

PR #3573, opened 2023-04-10
Feat: Changelly getTransactions method implemented

Gamboster:


packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -1707,6 +1707,23 @@ export class ExpressApp {
...
+      } catch (ex) {
+        return returnError(ex, res, req);
+      }
+      server
+        .changellyGetTransactions(req)
+        .then(response => {
+          res.json(response);
+        })
+        .catch(err => {
+          if (err) return returnError(err, res, req);
```
kajoseph, line 1723: Looks like this is a copy n paste from elsewhere, but when would `!err` be true? And if it were, we'd just never respond to the request? 

I think we should get rid of the `if (err)` condition and just `return returnError(...);` .... maybe pass in `err || 'unknown'` if we're concerned about `!err` being true.

kajoseph (changes requested):
Looks good overall. Just a note about the possibility of never responding to the request... which applies to other exiting endpoints as well.

packages/bitcore-wallet-service/src/lib/expressapp.ts
Gamboster: Done. Ty!
I only changed the response for this particular request. What do you think if I make the changes for all the pending requests later in a separate PR?

kajoseph (approved):
ACK

PR #3574, opened 2023-04-10
[BCN] Optimize pruneMempool

kajoseph:
This PR optimizes TransactionStorage.pruneMempool by streamifying the data instead of storing all relevant txs and coins in memory.

I've also snuck in a logging fix for BWS. For high call volume, consecutive calls to `console.time()` with the same keys before `console.timeEnd()` would throw noisy errors in the logs such as:
```
(node:25) Warning: Label 'V8getUtxos' already exists for console.time()
```
which then cause
```
(node:25) Warning: No such label 'V8getUtxos' for console.timeEnd()
```

I fixed this by putting randomly generated identifiers in the key strings to prevent consecutive call conflicts.

### DIFF
--- a/packages/bitcore-node/src/models/transaction.ts
+++ b/packages/bitcore-node/src/models/transaction.ts
@@ -610,7 +610,7 @@ export class TransactionModel extends BaseTransaction<IBtcTransaction> {
     return allRelatedCoins;
   }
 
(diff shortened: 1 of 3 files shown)


escottalexander (approved):
These changes look great.

PR #3578, opened 2023-04-14
Add script to sync wallets in coins and txns

kajoseph:
There was an problem where the `wallets` array on Transactions were not being populated with the wallet ObjectIds, but the Coin's `wallets` were. The end result of this problem is that wallet users do not see some transactions in their wallet's tx history, but the balance and list of UTXOs is correct.

This PR adds a script which uses the Coin's `wallets` to populate the Txns' withing a block window.


packages/bitcore-node/scripts/syncTxnAndCoinsWallets.js
```diff
@@ -0,0 +1,119 @@
...
+const { Storage } = require('../build/src/services/storage');
+const { TransactionStorage } = require('../build/src/models/transaction');
+const { CoinStorage } = require('../build/src/models/coin');
+
+function usage(err) {
+  if (err) {
+    console.log(err);
+    console.log('');
+  }
+  console.log('Usage: ./fixTxnToWalletMarriage.js [options] <chain> <network> <startBlockHeight> <endBlockHeight>');
```
escottalexander, line 18: We should change this to the actual name 'fixTxnToWalletMarriage.js' => 'syncTxnAndCoinsWallets.js'

packages/bitcore-node/scripts/syncTxnAndCoinsWallets.js
```diff
@@ -0,0 +1,119 @@
...
+  console.log('RUNNING FOR REAL. DATA WILL BE MODIFIED IN THE DATABASE');
+}
+
+console.log('Connecting to database...');
+
+Storage.start({ dbName: 'bitcore_mocha_scripts' })
+  .catch(console.error)
+  .then(async () => {
+    try {
+      const logFileName = `fixTxnsToWalletMarriageOutput-${chain}-${network}-${new Date().toISOString()}.log`;
```
escottalexander, line 66: Change name here too for consistency.

packages/bitcore-node/scripts/syncTxnAndCoinsWallets.js
```diff
@@ -0,0 +1,119 @@
...
+  usage('Invalid endBlockHeight: ' + endBlockHeight);
+}
+
+if (!dryRun) {
+  console.log('RUNNING FOR REAL. DATA WILL BE MODIFIED IN THE DATABASE');
+}
+
+console.log('Connecting to database...');
+
+Storage.start({ dbName: 'bitcore_mocha_scripts' })
```
escottalexander, line 62: I see the point in leaving the test db here for when the test runs but I am curious if there is a better way to handle this. I think at a minimum you should leave a comment at the top of the script indicating that the db name needs to be changed to run against prod. You could also make this a variable defined up top so that it is easily changed without having to dive into the script.

packages/bitcore-node/scripts/test/syncTxnAndCoinsWallets.spec.js
```diff
@@ -0,0 +1,715 @@
...
+        "mintHeight": 784431,
+        "coinbase": false,
+        "value": 546,
+        "script": "BinData(0,\"USAhNi1VGtNAPap2ZvEHSBd1cRyyQ9tIggIIfx7ZUGihdw==\")",
+        "spentHeight": -2,
+        "wallets": [wallets[1]]
+      }
+    ]
+  }
+]
```
escottalexander, line 715: Add semicolon. This is partially my bad. When I removed prettier awhile back, things like this stopped getting enforced on commit. We need to fix that before things get too messy.

escottalexander (commented):
Just a few minor suggestions. Works great.

packages/bitcore-node/scripts/test/syncTxnAndCoinsWallets.spec.js
kajoseph: Are you seeing a lint squiggly? If so, which linter is catching it? I think tslint is what's working in this repo for me, ATM. We need to unify this monorepo to use eslint.

packages/bitcore-node/scripts/syncTxnAndCoinsWallets.js
kajoseph: D'oh! I kept seeing this while I was testing and reminding myself to remove it before committing. ...I forgot :-1:

escottalexander (approved):
LGTM

Issue #3576, opened 2023-04-11
Q: Should we filter out `conflicting` coins when fetching unspent outputs ? 


peng-huang-ch:
Use the `Get Transaction Outputs by Address` 

`https://****/api/DOGE/mainnet/address/DS2cpSDvZXHzhZEyBEbTNuQTgonoHkFM4n/?unspent=true`

the result contains `conflicting` coins.

```sh
[
	{
		"_id": "6435026ee26b85d3ccb3bfa0",
		"chain": "DOGE",
		"network": "mainnet",
		"coinbase": false,
		"mintIndex": 1,
		"spentTxid": "",
		"mintTxid": "7d256872edff56b65d1c444f467b3bbb5bf766e7b3ef25452f24aa12af306ca0",
		"mintHeight": 4671846,
		"spentHeight": -2,
		"address": "DS2cpSDvZXHzhZEyBEbTNuQTgonoHkFM4n",
		"script": "76a914e521015b004d2d5023d5cd6bcd14ff297697d5f388ac",
		"value": 7738340,
		"confirmations": -1
	},
        ...
	{
		"_id": "63f091852e2b2be930f32466",
		"chain": "DOGE",
		"network": "mainnet",
		"coinbase": false,
		"mintIndex": 0,
		"spentTxid": "",
		"mintTxid": "006140516e2133d4fee801d99f08b5097f64246313fa8619ee9440a62b5b6d5d",
		"mintHeight": -3,
		"spentHeight": -2,
		"address": "DS2cpSDvZXHzhZEyBEbTNuQTgonoHkFM4n",
		"script": "76a914e521015b004d2d5023d5cd6bcd14ff297697d5f388ac",
		"value": 20000000,
		"confirmations": -1
	}
]
```

Expect:

remove the `conflicting` coins.

The [getBalanceForAddress](https://github.com/bitpay/bitcore/blob/98611e6bb8be12be6667babfe857f73b8b0095c6/packages/bitcore-node/src/providers/chain-state/internal/internal.ts#L96) will filter out the `conflicting` coins. 

There should be the same logic for [getting `unspent` outputs](https://github.com/bitpay/bitcore/blob/98611e6bb8be12be6667babfe857f73b8b0095c6/packages/bitcore-node/src/providers/chain-state/internal/internal.ts#L69)?

```js
if (args.unspent) {
      query.spentHeight = { $lt: SpentHeightIndicators.minimum };
[...]

PR #3591, opened 2023-05-16
Filter out conflicting coins when fetching unspent outputs.

peng-huang-ch:
Filter out conflicting coins when fetching unspent outputs

[#3576](https://github.com/bitpay/bitcore/issues/3576)


kajoseph (approved):
ACK

Thanks!

kajoseph:
@peng-huang-ch You'll need to sign your commits before I can merge this in.

image

PR #3592, opened 2023-05-17
[CWC] Handle numbers given in scientific notation form

escottalexander:
The crux of the issue is that very large numbers are returned from mongo with scientific notation (ie 1.1573971113625423e+25). The Web3.utils.toBN method does not know what to do with scientific notation when given as a number or even as a string. The solution is to parse all numbers as a BigInt but then return the number as a string (which toBN is ready to handle).

Once this is is merged we need to publish and update BitPay and the Wallet with the new changes. Will need to get the new version of BWS and BWC as well as CWC.

### DIFF
--- a/packages/crypto-wallet-core/src/transactions/erc20/index.ts
+++ b/packages/crypto-wallet-core/src/transactions/erc20/index.ts
@@ -40,13 +40,13 @@ export class ERC20TxProvider extends ETHTxProvider {
       const amounts = [];
       for (let recipient of recipients) {
         addresses.push(recipient.address);
-        amounts.push(toBN(recipient.amount));
+        amounts.push(toBN(BigInt(recipient.amount).toString()));
       }
       const multisendContract = this.getMultiSendContract(contractAddress);
       return multisendContract.methods.sendErc20(tokenAddress, addresses, amounts).encodeABI();
     } else {
       const [{ address, amount }] = params.recipients;
-      const amountBN = toBN(amount);
+      const amountBN = toBN(BigInt(amount).toString());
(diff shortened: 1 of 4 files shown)


packages/crypto-wallet-core/src/transactions/erc20/index.ts
```diff
@@ -40,13 +40,13 @@ export class ERC20TxProvider extends ETHTxProvider {
...
         addresses.push(recipient.address);
-        amounts.push(toBN(recipient.amount));
+        amounts.push(toBN(BigInt(recipient.amount).toString()));
       }
       const multisendContract = this.getMultiSendContract(contractAddress);
       return multisendContract.methods.sendErc20(tokenAddress, addresses, amounts).encodeABI();
     } else {
       const [{ address, amount }] = params.recipients;
-      const amountBN = toBN(amount);
+      const amountBN = toBN(BigInt(amount).toString());
```
leolambo, line 49: getting the following error here

`Error: [number-to-bn] while converting number 1.15207373271889e+26 to BN.js instance, error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported. Given value: "1.15207373271889e+26"`

weird considering the use of the similar changes in `checkValidTxAmount` no longer throw an error. assuming this error will occur on the same `index.ts` changes for polygon and native eth

leolambo (approved):

leolambo:
wait to merge

leolambo:
good to go

leolambo (approved):

kajoseph (approved):

PR #3598, opened 2023-06-06
WIP: [BWS] Record confirmations and reverts on EVM transaction proposals

leolambo:
We want to be able to know whether or not an Ethereum transaction in confirmed in order to determine if it should be considered pending.

We also want to track when a transaction is reverted.


leolambo:
Adding functionality to track failed evm transactions (reverted, dropped and replaced)

PR #3615, opened 2023-07-27
Ref: "Buy crypto" rule added to services data

Gamboster:
With this PR:

The blocking rule will be as follows:
- If the user is logged in but the account address is NY, "Buy crypto" will be blocked, regardless of current location
- If the user is logged out, the block will be based on the current location (IP). If it is in NY then it is blocked.

The message “This service is currently unavailable in your area.“ will be shown into a modal in the buy crypto root page for NY users, and they won't be able to use this feature.


kajoseph (approved):
ACK

PR #3618, opened 2023-08-10
[BCN] Add replacedByTxid to RBF'd txs

kajoseph:
Currently, there is no good way to identify which transaction replaced an existing one in the mempool. If I have a system that is monitoring a tx and that tx is RBF'd, all our API endpoint returns is a blockHeight of -3 (conflicting) without a pointer to the conflicting tx.

This PR adds the `replacedByTxid` field to the transaction schema and records it when a tx is RBF'd. This required a refactor of the `pruneMempool` method to reliably record the `replacedByTxid` where appropriate. The result is a simpler, more elegant `pruneMempool` method.

### DIFF
--- a/packages/bitcore-node/src/models/baseTransaction.ts
+++ b/packages/bitcore-node/src/models/baseTransaction.ts
@@ -15,6 +15,7 @@ export interface ITransaction {
   fee: number;
   value: number;
   wallets: ObjectID[];
+  replacedByTxid?: string;
 }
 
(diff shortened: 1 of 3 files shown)


escottalexander (approved):
This works perfect. Much cleaner than it used to be.

nitsujlangston (approved):

Khademr7 (approved):

PR #3620, opened 2023-08-14
[BWS] Multisend gasLimit buffer

leolambo:
Adds a 10 percent gas limit buffer on multisend transactions

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
+++ b/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
@@ -215,7 +215,7 @@ export class EthChain implements IChain {
               data,
               gasPrice
             });
-            gasLimit = gasLimit + gasLimit * Defaults.MS_GAS_LIMIT_BUFFER_PERCENT;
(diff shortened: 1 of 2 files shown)


packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -215,6 +215,7 @@ export class EthChain implements IChain {
               data,
               gasPrice
             });
+            gasLimit = gasLimit + gasLimit * Defaults.MS_GAS_LIMIT_BUFFER_PERCENT;
```
kajoseph, line 218: Need to convert this to an int

```
FpNUAGe6RUZ79o3FXeJeXB: Error: Error: [number-to-bn] while converting number 148679.3 to BN.js instance, error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported. Given value: "148679.3"
```

This can also be cleaned up a little to 
```
gasLimit += gasLimit * Defaults.MS_GAS_LIMIT_BUFFER_PERCENT;
```

kajoseph (changes requested):

kajoseph (approved):

PR #3621, opened 2023-08-14
[REF] oneinch and coingecko api

gabrielbazan7:


Gamboster (approved):
LGTM

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -6015,35 +6016,56 @@ export class WalletService implements IWalletService {
...
+
+        this.request.get(
+          URL,
+          {
+            headers,
+            json: true
+          },
+          (err, data) => {
+            if (err) {
+              return reject(err.body ?? err);
```
kajoseph, line 6051: Should we throw or log and return oldValues?

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -6015,35 +6016,56 @@ export class WalletService implements IWalletService {
 
   oneInchGetTokens(req): Promise<any> {
     return new Promise((resolve, reject) => {
+
       const credentials = this.oneInchGetCredentials();
+      const chain = req.params?.['chain'];
```
kajoseph, line 6021: Fallback to 'eth'? Actually, I would recommend falling back in the route handler since the chain param is optional (i assume for backwards compatibility from before polygon)

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -6197,8 +6219,66 @@ export class WalletService implements IWalletService {
...
+
+    return credentials;
+  }
+
+  coinGeckoGetRates(req): Promise<any> {
+    return new Promise((resolve, reject) => {
+  
+      const credentials = this.coinGeckoGetCredentials();
+      const chain = req.params?.['chain'];
+      const cacheKey = `coinGeckoRates:${chain}`;
```
kajoseph, line 6247: This cacheKey is incomplete since the API results depend on contractAddresses, and altCurrencies. But if we're wanting to maximize API usage, can we instead always call the API for a static set of contractAddresses and altCurrencies and then prune the cache retval to fit the query params?

gabrielbazan7:
Done. Thanks

kajoseph (changes requested):

packages/bitcore-wallet-service/src/lib/server.ts
gabrielbazan7: Hey, thanks for reviewing. That's not possible since `contract_addresses` is required and could be any token address that we support in the app. I removed the cache for this request ( this request is only for users that have tokens that are not supported by bitpay invoices )

kajoseph (approved):
tACK

PR #3622, opened 2023-08-16
[BCN] Add replacedByTxid to EVM

kajoseph:


escottalexander:
I was able to simulate this locally and it worked as expected.

escottalexander (approved):
This works as expected.

nitsujlangston (approved):

PR #3628, opened 2023-08-29
[BWS] WalletService type in routes

kajoseph:
This adds the WalletService type to the expressapp.ts routes file to allow easy mapping into the `server` var that's prevalent throughout the file.


escottalexander (approved):
This is gold. Love it when we add types.

nitsujlangston (approved):

PR #3632, opened 2023-09-13
[BWS] Cleanup fiat rates service

kajoseph:
This turns the providers files into ES6 modules and adds typing. Also removes lodash in places and cleans up syntax with null conditional operator ("?." vs ternary).


escottalexander (approved):
This looks good. Nice clean up.

nitsujlangston (approved):

PR #3634, opened 2023-09-14
[Bitcore-Client] Add option to use existing xpriv to create wallet

escottalexander:
This enables creating a wallet from an existing extended private key.

### DIFF
--- a/packages/bitcore-client/src/wallet.ts
+++ b/packages/bitcore-client/src/wallet.ts
@@ -162,7 +162,9 @@ export class Wallet {
     });
 
     if (!xpriv) {
-    console.log(mnemonic.toString());
+      console.log(mnemonic.toString());
+    } else {
+      console.log(hdPrivKey.toString());
     }
 
     await loadedWallet.register().catch(e => {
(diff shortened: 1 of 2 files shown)


kajoseph (changes requested):

packages/bitcore-client/src/wallet.ts
```diff
@@ -82,14 +83,20 @@ export class Wallet {
...
     let { storageType, storage } = params;
     if (!chain || !network || !name) {
       throw new Error('Missing required parameter');
     }
     // Generate wallet private keys
-    const mnemonic = new Mnemonic(phrase);
-    const hdPrivKey = mnemonic.toHDPrivateKey('', network).derive(Deriver.pathFor(chain, network));
+    let hdPrivKey;
+    let mnemonic;
+    if (xpriv) {
```
kajoseph, line 94: Just thinking about edge cases - we should check for mutual exclusivity between the phrase and xpriv params

kajoseph (changes requested):

packages/bitcore-client/src/wallet.ts
escottalexander: Good call. Added in latest commit.

kajoseph (approved):
tACK

PR #3638, opened 2023-10-03
[Client] Add --addressType to wallet creation; Squish all the bugs! :bug: 

kajoseph:
Currently, bitcore-client wallets are pubkeyhash wallets only. This adds the ability to create wallets using other address types.

Note, taproot addresses are not fully implemented in bitcore-lib yet and don't work (hence the bypass in the tests)

Also, this fixes several bugs:
* `bitcore-lib-cash` and `bitcore-lib-doge` disregarded the specified address type when computing an address from a public key
* `wallet-send` was all kinds of messed up due to the fee calculation.
  * There was a fallback of "20" which didn't make any sense so I added a `--feeRate` optional param.
  * The `selectCoins` method required a `fee` param, but none was given, so it produced a NaN, causing the tx to not have _any_ inputs. So, I added a `|| 0` in the reduce.
  * The `feeRate` in `wallet-send` did a multiplication by `scale`, but the `feerate` value returned from bitcore-node was in a different scale for EVM and UTXO chains (i.e. EVM returned Wei, UTXO returned BTC). In both cases, this resulted in the tx having an exorbitant feeRate. I fixed this to calculate sats/KB for UTXO and Wei for EVM, considering if the user specified the `--feeRate` param or not.
* The change was also somewhat messed up in that there was no way to *not* have a change output. However, the `wallet.deriveAddress` fallback when constructing a tx does NOT *save* the address anywhere. Thus, the wallet would essentially lose those funds since it would have no awareness of the generated change address.
  * This PR removes the fallback which properly puts the onus of keeping track of addresses back on the upstream logic, and adds change generation logic to `Wallet.newTx
* `wallet storage --import` was broken because it would detect `--name` as a function, so the imported wallet would have a `name` of undefined, but in order to use the wallet you needed to pass in `--name "function(str) {
  if (arguments.length === 0) return this._name;
  this._name = str;
  return this;
}"` :grimacing:
* Parameter validation was broken. I upgraded commander, removed `program.ts` (which unsuccessfully tried to do some param validation), and changed the param formatting to leverage commander's built-in param validation.


packages/bitcore-client/tsconfig.json
escottalexander, line 1: Something about adding ./test/* makes it so that when it is compiled it puts files in ts_build/src (or test) whereas prior to this change everything existed at the top level of ts_build.

It breaks several places throughout the wallet utils where files are obtained without any reference to the src directory. For instance `const { Storage } = require('../ts_build/storage');`

I would just change the references in those files to match the new directory structure.

escottalexander (changes requested):
One change that is pretty crucial but an easy fix. 

By the time you are done with these changes bitcore-client will be brand new. 😂

packages/bitcore-client/tsconfig.json
escottalexander: Also, pertaining to the error in the pipeline, when I ran `npm i` from the top level of bitcore it showed three package-lock files. Maybe if you commit those it will fix the error.

kajoseph:
@escottalexander phew! pipeline is passing again. NPM really didn't like me manually adding those dev dependencies.

escottalexander:
@kajoseph Still need to address this: https://github.com/bitpay/bitcore/pull/3638#discussion_r1346338177
I think it got lost when you pushed changes to fix the pipeline.

packages/bitcore-client/src/wallet.ts
```diff
@@ -360,16 +360,21 @@ export class Wallet {
       }
       tokenContractAddress = tokenObj.address;
     }
+    let change = params.change;
+    if (change === 'miner') {
+      change = undefined; // no change
+    } else if (!change) {
+      change = await this._getChangeAddress();
```
escottalexander, line 367: This change causes the wallet to need to be unlocked before the newTx method is called. Otherwise it gives an exception. In the case of wallet-send you just need to move the unlock statement up above the call to newTx. Looks like wallet-bump-tx, wallet-payrpo and wallet-tx also use this method.

escottalexander: The reason why is because the call to derivePrivateKey uses `this.unlocked.masterKey` and `this.unlocked` is undefined before unlocking the wallet.

escottalexander (changes requested):
One issue that should hopefully be easy to fix.

escottalexander (approved):
At long last, LGTM!

packages/bitcore-client/bin/wallet-create
```diff
@@ -10,11 +10,7 @@ program
   .version(require('../package.json').version)
   .option('--name <name>', 'REQUIRED - Wallet Name')
   .option('--chain <chain>', 'REQUIRED - Chain (BTC/BCH/ETH, etc)')
-  .option(
-    '--network <network>',
-    'REQUIRED - Network (livenet/testnet/regtest, etc)'
-  )
-  .option('--addressType [type]', 'optional - Specify the address type (default: pubkeyhash)')
```
escottalexander: It seems that addressType was accidentally removed.

kajoseph: good catch

escottalexander (changes requested):
This looks great! Just a couple very vary small changes.

escottalexander (approved):
LGTM!

nitsujlangston (approved):

PR #3643, opened 2023-10-27
Fix indexes

escottalexander:
We found some issues with the indexes I had put together. This should fix it.

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/evm/models/transaction.ts
+++ b/packages/bitcore-node/src/providers/chain-state/evm/models/transaction.ts
@@ -102,14 +102,14 @@ export class EVMTransactionModel extends BaseTransaction<IEVMTransaction> {
       { chain: 1, network: 1, 'effects.to': 1, blockTimeNormalized: 1 },
       {
         background: true,
-        sparse: true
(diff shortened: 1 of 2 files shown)


packages/bitcore-node/src/providers/chain-state/evm/models/transaction.ts
```diff
@@ -102,14 +102,14 @@ export class EVMTransactionModel extends BaseTransaction<IEVMTransaction> {
...
-        sparse: true
+        partialFilterExpression: { 'effects.0': { $exists: true } }
       }
     );
     this.collection.createIndex(
-      { chain: 1, network: 1, 'effects.amount': 1, blockTimeNormalized: 1 },
+      { chain: 1, network: 1, 'effects.from': 1, blockTimeNormalized: 1 },
       {
         background: true,
-        sparse: true
```
kajoseph, line 112: actually, the partial filters should be `partialFilterExpression: { 'effects.to': { $exists: true } }` and `partialFilterExpression: { 'effects.from': { $exists: true } }`

kajoseph (changes requested):

kajoseph (approved):
tACK

PR #3647, opened 2023-11-01
[BCC] Fix XRP issues with bitcore-client

kajoseph:
This [removes the semvar](https://github.com/bitpay/bitcore/pull/3647/files#diff-e37da4f64183a7da37dc38b44dd7e35033afeaa4780fbfd4f1b0ef7c2b698f85R34) from the `ripple-lib` dep in CWC. `ripple-lib` v1.10.1 was being installed in other environments which caused the following error when signing an XRP tx
```TypeError: Cannot read properties of undefined (reading 'encodeForSigning')```

Additionally, I've cleaned up the check for needing change addresses and effectively excluded XRP wallets from creating change addresses as well.


escottalexander (approved):
This change looks good and works as expected.

PR #3648, opened 2023-11-01
[CWC] Replace ripple-lib with xrpl

kajoseph:


Dopeso (commented):
Wallet with cash balance review

escottalexander (approved):
This looks good and works locally.

PR #3649, opened 2023-11-07
[BCN] Better error handling for getEffects

escottalexander:
This PR checks for each param before attempting to parse unexpectedly similar methods and also adds a try/catch to make sure any edge cases are covered and reported.


kajoseph (approved):
tACK

PR #3650, opened 2023-11-08
[CWC] Transactions - Support evm type 2

leolambo:
Allows type 2 tx fields used in transaction creation. maxPriorityFeePerGas and maxFeePerGasare set on the [unsigned transaction](https://github.com/ethers-io/ethers.js/blob/master/packages/transactions/src.ts/index.ts#L32)


packages/crypto-wallet-core/src/transactions/eth/index.ts
```diff
@@ -41,15 +43,21 @@ export class ETHTxProvider {
...
-      gasPrice: utils.toHex(gasPrice),
       to,
       data,
       value: utils.toHex(amount),
       chainId
     };
+    if (maxGasFee) {
+      txData.maxFeePerGas = utils.toHex(maxGasFee);
+      txData.maxPriorityFeePerGas = utils.toHex(priorityGasFee || 0);
+    } else {
```
kajoseph, line 57: Need to add `txData.type = 2;`

packages/crypto-wallet-core/src/transactions/eth/index.ts
```diff
@@ -41,15 +43,21 @@ export class ETHTxProvider {
     }
     let { chainId } = params;
     chainId = chainId || this.getChainId(network);
-    const txData = {
+    let txData: any = {
       nonce: utils.toHex(nonce),
       gasLimit: utils.toHex(gasLimit),
```
kajoseph, line 48: Should we move gasLimit to inside the else statement below?

kajoseph (changes requested):
Let's add a few test cases to this as well

packages/crypto-wallet-core/src/transactions/eth/index.ts
kajoseph: Nvm

kajoseph (approved):
tACK

Nice work!

PR #3654, opened 2023-11-10
[BCC] Fix adding like token to wallet

kajoseph:
The current behavior is that it just keeps appending to `wallet.tokens`. The problem is that there is a) no check to see if the token already exists (to prevent duplication) and querying balance only grabs the first entry in the array.

This PR checks for duplicity and prompts the user to replace the existing token. This is most helpful in dev/test environments when you might want to update the token address.


escottalexander (approved):
Nice workflow improvement. This looks good to me.

nitsujlangston (approved):

PR #3657, opened 2023-11-15
[FEAT] add native usdc

gabrielbazan7:


packages/crypto-wallet-core/src/constants/tokens.ts
```diff
@@ -75,7 +75,7 @@ export let ETH_TOKEN_OPTS = {
 
 export let MATIC_TOKEN_OPTS = {
   '0x2791bca1f2de4661ed88a30c99a7a9449aa84174': {
-    name: 'USD Coin (PoS)',
+    name: 'USDC',
     symbol: 'USDC',
```
leolambo, line 79: name and symbol should be USDC.e

kajoseph (approved):
ACK

PR #3658, opened 2023-11-16
[BCN] Add invoice payment effect

escottalexander:
This adds a new effect to each transaction that calls our invoice contracts pay method.


escottalexander:
Converted this to a draft since the need for the changes has been addressed a different way.

PR #3659, opened 2023-11-27
[BWS-BWC] Remove square brackets from query params

escottalexander:
TLDR We need to remove the use of square brackets in our getStatusAll endpoint.

From Gabriel:
We’ve encountered an issue with the getStatusAll method on iOS 17 devices. It seems that we are using an invalid character in the GET request ( https://www.ietf.org/rfc/rfc3986.txt ) :
```
A host identified by an Internet Protocol literal address, version 6 [RFC3513] or later, is distinguished by enclosing the IP literal within square brackets ("[" and "]"). This is the only place where square bracket characters are allowed in the URI syntax.
[1:14](https://bitpay.slack.com/archives/D03C0QN831U/p1701108841501659)
```
this is the error on the app side:
```
Task <CED3943C-443E-4F48-869D-648126EBC4DF>.<18> finished with error [18,446,744,073,709,550,607] Error Domain=NSURLErrorDomain Code=-1009 "URL not allowed" UserInfo={_NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalDataTask <CED3943C-443E-4F48-869D-648126EBC4DF>.<18>"
), NSLocalizedDescription=URL not allowed, _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <CED3943C-443E-4F48-869D-648126EBC4DF>.<18>}
```


gabrielbazan7 (approved):
ACK ! tested on device

kajoseph (dismissed):

gabrielbazan7:
Something I noticed is that we are not handling the old params for users using old versions

gabrielbazan7 (changes requested):
Something I noticed is that we are not handling the old params in bws for users using old versions of bwc

gabrielbazan7 (approved):
ACK

kajoseph (approved):

PR #3660, opened 2023-11-28
[Bitcore] Transactions - Supprt EVM type 2 tx

leolambo:


packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
```diff
@@ -517,6 +517,30 @@ export class V8 {
...
   }
 
+  estimateFeeV2(opts, cb) {
+    const nbBlocks = opts.nbBlocks || [1, 2, 6, 24];
+    const txType = opts.txType;
+    if(!txType || txType.toString() !== '2') {
+     return this.estimateFee(nbBlocks, cb);
+    }
+    let result;
+    const url = this.baseUrl + `/fee/10/${txType}`;
```
kajoseph, line 527: i think you mean `${nbBlocks}` instead of `10`

packages/bitcore-node/src/routes/api/fee.ts
```diff
@@ -5,18 +5,22 @@ import { CacheMiddleware } from '../middleware';
...
+  if (
+    cachedFee 
+    && cachedFee.date > Date.now() - 10 * 1000
+    && (!txType || txType.toString() !== '2')
+    ) {
     return res.json(cachedFee.fee);
   }
   try {
-    let fee = await ChainStateProvider.getFee({ chain, network, target: targetNum });
+    let fee = await ChainStateProvider.getFee({ chain, network, target: targetNum, txType});
```
kajoseph, line 23: lint: add space in params

packages/bitcore-node/src/routes/api/fee.ts
```diff
@@ -30,6 +34,14 @@ router.get('/:target', CacheMiddleware(CacheTimes.Second), async (req: Request,
   } catch (err) {
     return res.status(500).send('Error getting fee from RPC');
   }
+}
+
+router.get('/:target', CacheMiddleware(CacheTimes.Second), async (req: Request, res: Response) => {
+  getFee(req, res);
+});
+
+router.get('/:target/:txType', CacheMiddleware(CacheTimes.Second), async (req: Request, res: Response) => {
```
kajoseph, line 43: this is minor, but i think it would be a little cleaner and scalable to add a `txType` query param to the existing route rather making txType a path param. (side note: Express does have syntax for optional path params, but i would recommend not using those except in rare cases.)

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
```diff
@@ -106,12 +106,16 @@ export class BaseEVMStateProvider extends InternalStateProvider implements IChai
...
     if (network === 'livenet') {
       network = 'mainnet';
     }
-
+    if (txType?.toString() === '2') {
+      const { rpc } = await this.getWeb3(network);
+      const feerate = await rpc.estimateFee({ nBlocks: target, txType });
+      return { feerate, blocks: target };
+    }
     const cacheKey = `getFee-${chain}-${network}-${target}`;
```
kajoseph, line 119: should add txType to the cacheKey. Also tweak the test that checks that it's not cached.

kajoseph (changes requested):

PR #3661, opened 2023-11-28
[BCN] Add configurable fee mode

kajoseph:


packages/bitcore-node/src/routes/api/fee.ts
```diff
@@ -1,22 +1,45 @@
...
 const feeCache = {};
 
+const feeModes = {
+  BTC: ['CONSERVATIVE', 'ECONOMICAL'],
+  LTC: ['CONSERVATIVE', 'ECONOMICAL']
+};
+
 router.get('/:target', CacheMiddleware(CacheTimes.Second), async (req: Request, res: Response) => {
   let { target, chain, network } = req.params;
+  let { mode } = req.query as any;
```
escottalexander, line 18: Specifying 'any' here is redundant as req.query is already any. Not sure if you included for some linting error but it works fine for me.

escottalexander (approved):
This looks great and works as expected. Left one small comment on a type specification but otherwise this is good to go. I am approving just in case you have some good reason I missed that would lead to us disregarding my comment.

packages/bitcore-node/src/routes/api/fee.ts
kajoseph: removing `as any` implicitly sets mode to a string which doesn't jive with the `'CONSERVATIVE' | 'ECONOMICAL' | undefined` type in the rest of the code.

escottalexander (approved):
LGTM!

packages/bitcore-node/src/routes/api/fee.ts
kajoseph: After a closer look, `let { mode} = req.query;` sets mode to any. I had likely been seeing an error that was left in my linter after moving mode from req.params to req.query.

nitsujlangston (approved):

PR #3668, opened 2023-12-18
[BWS] Prevent wrapped address types for non-segwit chain wallets

kajoseph:
There's a bug in the wallet that passes in `useNativeSegwit` even for BCH and DOGE wallets. The wallet bug is being fixed, but this will prevent older wallet versions from continuing to create wallets with segwit addresses for non-segwit chains.


leolambo:
LGTM!

leolambo (approved):

nitsujlangston (approved):

PR #3669, opened 2023-12-19
[BCN] Add rpc check to pruning

kajoseph:


escottalexander (approved):
Everything looks good and works as expected. Nice work.

nitsujlangston (approved):

PR #3672, opened 2023-12-28
[BCC] Support fee bump for UTXOs

kajoseph:


packages/bitcore-client/src/wallet.ts
```diff
@@ -413,7 +430,10 @@ export class Wallet {
       gasLimit: params.gasLimit || 200000,
       data: params.data,
       tokenAddress: tokenContractAddress,
-      contractAddress: params.contractAddress
+      contractAddress: params.contractAddress,
+      replaceByFee: params.replaceByFee,
```
escottalexander, line 434: This is nice to add but it appears CWC isn't ready to handle RBF. Fine to just keep it in preparation for adding full support.

escottalexander (changes requested):
Left a few comments on things that need to be addressed. Mostly small stuff. Nice work. Love the revisions made.

packages/bitcore-client/bin/wallet-bump-tx
```diff
@@ -5,69 +5,121 @@
...
+      if (lib.ethers) {
+        existingTx = lib.ethers.utils.parseTransaction(rawTx);
+      } else {
+        const tx = new lib.Transaction(rawTx);
+        txid = tx.id;
+      }
+    }
     if (txid) {
-      existingTx = JSON.parse(await wallet.getTransactionByTxid({ txid }));
-    } else if (rawtx) {
```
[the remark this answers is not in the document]
kajoseph: I think you're looking at the deleted code. What you suggested is close to what it is now.

packages/bitcore-client/src/wallet.ts
kajoseph: I changed CWC to support this locally but forgot to commit. Pushed now.

packages/bitcore-client/bin/wallet-send
```diff
@@ -40,14 +43,21 @@ const main = async () => {
...
     if (!feeRate) {
       const feerate = await wallet.getNetworkFee();
-      feeRate = JSON.parse(feerate).feerate;
+      feeRate = feerate.feerate;
     }
     let nonce;
-    if (!['BTC', 'BCH', 'DOGE', 'LTC'].includes(chain)) {
+    let feeUnit;
+    if (wallet.chain.toUpperCase() === 'XRP') {
+      feeRate = feeRateParam == null ? feeRate : xrpl.xrpToDrops(feeRateParam); // convert feeRateParam from XRP to drops
```
escottalexander, line 51: Add this here so that the sequence (XRP name for nonce) is updated correctly:
`nonce = await wallet.getNonce();`

packages/bitcore-client/bin/wallet-send
```diff
@@ -40,14 +43,21 @@ const main = async () => {
...
+      console.log('Fee rate:', feeRate, 'drops');
+    } else if (!wallet.isUtxoChain()) {
       nonce = await wallet.getNonce();
-      feeRate = Math.ceil(feeRateParam == null ? feeRate : (feeRateParam * 1e9)); // convert feeRateParam from Gwei to Wei
+      feeRate = feeRateParam == null ? feeRate : Web3.utils.toWei(feeRateParam.toString(), 'gwei').toNumber(); // convert feeRateParam from Gwei to Wei
```
escottalexander, line 55: Getting an error here about `toNumber` not existing. You can solve by wrapping the logic in `Number(...)`.

escottalexander (changes requested):
Nice fixes. Great idea to add XRP support. It just needs a few tweaks to work.

nitsujlangston (approved):

PR #3674, opened 2024-01-04
[BCC] Handle single output tx fee bumping

kajoseph:
This handles the case when the tx is a sweep/max send and all given utxos should be used.

Also moved the fee bump logic into the Wallet class for integrations.


packages/bitcore-client/bin/wallet-bump-tx
escottalexander, line 1: You can remove the reference to Web3 on line 8.

escottalexander (changes requested):
Everything is good here and works. Just remove the Web3 ref at the top of wallet-bump-tx.

escottalexander (approved):
LGTM!

nitsujlangston (approved):

PR #3679, opened 2024-01-11
[Insight] Fix address page confs; Fix tx refresh; Add tx conf listener

kajoseph:
This includes some general improvements to insight:

Address page:
* Fixed a bug where unconfirmed txs would show the tip height as confirmations instead of `unconfirmed` (i.e. unconfirmed tx's would show as having 800k+ confirmations)
* Fixed the number of transactions to be correct
* Fixed "Nro. Transactions" to use more common abbrev. for number (`No.`)
* Made the header say "Address" instead of "Summary" and made the actual address more prevalent

Transaction page:
* Fixed the issue where refreshing the page would not refresh confirmations, even when shift+refreshing
* Added polling for confirmations if unconfirmed. Plays a sound upon status change.


lanchana (approved):
LGTM!

nitsujlangston (approved):

PR #3683, opened 2024-01-16
[BCN] Move descendant invalidation to pruning service

kajoseph:
This PR:

* Moves descendant invalidation to the pruning service by recycling the processAllInvalidTxs() function to invalidate descendants. The sync process will mark immediate replacements as replaced/invalid and then this process cleans up descendant trees asynchronously.
* Adds a script to manually fix specific replacement tree.

### DIFF
--- a/packages/bitcore-node/scripts/fixReplacedTree.js
+++ b/packages/bitcore-node/scripts/fixReplacedTree.js
@@ -42,6 +42,8 @@ if (!['BTC', 'BCH', 'DOGE', 'LTC'].includes(chain) || !['mainnet', 'testnet', 'r
 const real = !!args.find(a => a === '--real');
 const force = !!args.find(a => a === '--force');
 
+real && console.log('~~~~ DRY RUN ~~~~');
+console.log('Connecting to storage...');
 Storage.start()
   .then(async () => {
     const tx = await TransactionStorage.collection.findOne({ chain, network, txid });
--- a/packages/bitcore-node/scripts/fixReplacedTree.js
+++ b/packages/bitcore-node/scripts/fixReplacedTree.js
@@ -1,0 +1,71 @@
+#!/usr/bin/env node
+
+const CWC = require('crypto-wallet-core');
+const { Storage } = require('../build/src/services/storage');
+const { TransactionStorage } = require('../build/src/models/transaction');
+
+function usage(errMsg) {
+  console.log('USAGE: ./fixreplacedTree.js <txid> <replacementTxid> [options]');
+  console.log('OPTIONS:');
+  console.log('  --chain <value>      BTC, BCH, DOGE, or LTC');
+  console.log('  --network <value>    mainnet, testnet, or regtest');
+  console.log('  --real               Write the change to the db. If not given, will only do a dry run');
+  console.log('  --force              Force the overwrite of an existing replacedByTxid field');
+  if (errMsg) {
+    console.log('\nERROR: ' + errMsg);
+  }
+  process.exit();
+}
+
+const args = process.argv.slice(2);
+
+if (args.includes('--help') || args.includes('-h')) {
+  usage();
+}
+
+if (!args[0] || !/^[a-f0-9]{64}$/i.test(args[0])) {
+  usage('Missing or invalid txid param.');
+}
+
+if (!args[1] || !/^[a-f0-9]{64}$/i.test(args[1])) {
+  usage('Missing replacementTxid param.');
+}
+
+const [txid, replacementTxid] = args;
+const chain = args[args.indexOf('--chain') + 1];
+const network = args[args.indexOf('--network') + 1];
+
+if (!['BTC', 'BCH', 'DOGE', 'LTC'].includes(chain) || !['mainnet', 'testnet', 'regtest'].includes(network)) {
+  usage('Invalid chain and/or network param(s).');
+}
+
+const real = !!args.find(a => a === '--real');
+const force = !!args.find(a => a === '--force');
+
+Storage.start()
+  .then(async () => {
+    const tx = await TransactionStorage.collection.findOne({ chain, network, txid });
+
+    if (!tx) {
+      console.log('No tx found for txid.');
+      return;
+    }
+    
+    if (tx.replacedByTxid) {
+      console.log('Tx already has replacement txid: ' + tx.replacedByTxid);
+      if (!force) {
+        return;
+      }
+    }
+    
+    if (real) {
+      await TransactionStorage._invalidateTx({ chain, network, invalidTxid: txid, replacedByTxid: replacementTxid });
+
+      const tx = await TransactionStorage.collection.findOne({ chain, network, txid });
+      console.log('Tx replaced', tx.replacedByTxid);
+    } else {
+      console.log('Dry run complete.');
+    }
+  })
+  .catch(console.error)
+  .finally(Storage.stop.bind(Storage))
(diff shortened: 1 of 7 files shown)


kajoseph:
Something like this would fix the sync issue, but child txs would not be marked appropriately.

I'm considering having this fn merely set the replaced tx as such and then have a "cleanup" worker/process mark the children appropriately.

packages/bitcore-node/scripts/fixReplacedTree.js
```diff
@@ -0,0 +1,71 @@
...
+    const tx = await TransactionStorage.collection.findOne({ chain, network, txid });
+
+    if (!tx) {
+      console.log('No tx found for txid.');
+      return;
+    }
+    
+    if (tx.replacedByTxid) {
+      console.log('Tx already has replacement txid: ' + tx.replacedByTxid);
+      if (!force) {
```
MichaelAJay, line 56: how about a console log in the tx.replaceByTxid block, after the !force block. Something along the lines of "console.log('tx.replaceByTxid will be re-replaced')

I know that's a dumb message - but just to give you the idea. It will also serve as nice code documentation.

packages/bitcore-node/scripts/fixReplacedTree.js
```diff
@@ -0,0 +1,71 @@
...
+
+const [txid, replacementTxid] = args;
+const chain = args[args.indexOf('--chain') + 1];
+const network = args[args.indexOf('--network') + 1];
+
+if (!['BTC', 'BCH', 'DOGE', 'LTC'].includes(chain) || !['mainnet', 'testnet', 'regtest'].includes(network)) {
+  usage('Invalid chain and/or network param(s).');
+}
+
+const real = !!args.find(a => a === '--real');
```
MichaelAJay, line 42: What about if (!real) console.log('Script is being dry run').

Again, not only a nice message, but useful documentation

packages/bitcore-node/scripts/fixReplacedTree.js
```diff
@@ -0,0 +1,71 @@
...
+        return;
+      }
+    }
+    
+    if (real) {
+      await TransactionStorage._invalidateTx({ chain, network, invalidTxid: txid, replacedByTxid: replacementTxid });
+
+      const tx = await TransactionStorage.collection.findOne({ chain, network, txid });
+      console.log('Tx replaced', tx.replacedByTxid);
+    } else {
```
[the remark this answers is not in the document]
kajoseph: No, I'll keep it where it is so it's closer to the `if (real)` eval.

packages/bitcore-node/scripts/fixReplacedTree.js
kajoseph: --force was already given by the user, so no need to over log

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -673,29 +673,33 @@ export class TransactionModel extends BaseTransaction<IBtcTransaction> {
   async _invalidateTx(params: {
     chain: string;
     network: string;
-    invalidTxid: string,
-    replacedByTxid?: string // only provided at the beginning of the ancestral tree. Txs that spend unconfirmed outputs aren't "replaced"
-    invalidParentTxids?: string[] // empty at the beginning of the ancestral tree. 
+    invalidTxid: string;
+    replacedByTxid?: string; // only provided at the beginning of the ancestral tree. Txs that spend unconfirmed outputs aren't "replaced"
+    invalidParentTxids?: string[]; // empty at the beginning of the ancestral tree.
+    simple?: boolean; // if true, don't invalidate descendants
```
MichaelAJay, line 679: This param name sticks out to me as one that could use some work. What about 'skipInvalidateDescendants'?

packages/bitcore-node/src/models/transaction.ts
```diff
@@ -673,29 +673,33 @@ export class TransactionModel extends BaseTransaction<IBtcTransaction> {
...
+        spentHeight: SpentHeightIndicators.pending,
+        mintTxid: invalidTxid
+      };
 
-    // spent outputs of invalid tx
-    const spentOutputsStream = await CoinStorage.collection.find(spentOutputsQuery);
-    const seenTxids = new Set();
-    let output: ICoin | null;
+      // spent outputs of invalid tx
+      const spentOutputsStream = await CoinStorage.collection.find(spentOutputsQuery);
```
MichaelAJay, line 692: Getting rid of the await here could stop someone from making a mistake when working with the Cursor.

packages/bitcore-node/src/services/pruning.ts
```diff
@@ -112,144 +121,173 @@ export class PruningService {
...
-    logger.info(`Removed all pending txs older than ${days} days: ${rmCount}`);
+            })
+          )
+          .on('finish', resolve)
+          .on('error', reject);
+      });
+      logger.info(`Removed all pending ${chain}:${network} txs older than ${days} days: ${rmCount}`);
+      this.lastRunTimeOld = Date.now();
+    } catch (err: any) {
+      logger.error(`Error processing old mempool txs: ${err.stack || err.message || err}`);
```
MichaelAJay, line 202: This method is always called from w/in a try block. Is the behavior that you want to not re-throw? Seems like you'll miss information this way.

packages/bitcore-node/src/services/pruning.ts
```diff
@@ -112,144 +121,173 @@ export class PruningService {
...
+          continue;
+        }
+        seen.add(vout.mintTxid);
+        const tx = await this.transactionModel.collection.findOne({ chain, network, txid: vout.mintTxid });
+        if (!tx) {
+          logger.error(`Coin ${vout.mintTxid} has no corresponding tx`);
+          continue;
+        }
+        if (tx.replacedByTxid) {
+          if (await this.invalidateTx(chain, network, tx)) {
```
MichaelAJay, line 243: Using something that can update a DB directly in a conditional may lead to developer headaches. Consider assigning to 'didInvalidateTx' directly above this line & then checking that assignment in the conditional. Same on line 251

packages/bitcore-node/src/services/pruning.ts
```diff
@@ -112,144 +121,173 @@ export class PruningService {
...
-        .on('finish', resolve)
-        .on('error', reject);
-    });
+          }
+        }
+      }
+      logger.info(`Invalidated ${invalidCount} (processed ${realCount}) pending TXOs for ${chain}:${network}`);
+      this.lastRunTimeInvalid = Date.now();
+    } catch (err: any) {
+      logger.error(`Error processing invalid txs: ${err.stack || err.message || err}`);
```
MichaelAJay, line 261: This is called from a try block. If you don't throw an error, then you'll never receive the catch behavior specified where it's called.

packages/bitcore-node/src/services/pruning.ts
```diff
@@ -112,144 +121,173 @@ export class PruningService {
...
+        }        
+        await this.transactionModel._invalidateTx({
+          chain,
+          network,
+          invalidTxid: tx.txid,
+          replacedByTxid: tx.replacedByTxid
+        });
+        return true;
+      } catch (err: any) {
+        logger.error(`Error invalidating tx ${tx.txid}: ${err.stack || err.message || err}`);
```
MichaelAJay, line 285: Purposeful no throw?

packages/bitcore-node/src/services/pruning.ts
kajoseph: I just want to log and continue

MichaelAJay (approved):

nitsujlangston (approved):

PR #3686, opened 2024-01-23
[Libs] Add tx size & vsize, use it for accurate fee calculation

kajoseph:
This adds `size` and `vsize` properties to the Transaction prototype (except for the BCH lib, it only gets the `size` property). The size/vsize is then used in fee calculations to produce a more accurate fee (which is a lot less for segwit txs). The previously used `_estimateSize()` produced a very rough size estimation that was "fine" for a low fee market, but in this high fee market the size and fee calculations matter.

In addition to being more accurate, using the size/vsize was ~2x faster in my performance testing compared to `_estimateSize()`.


escottalexander (changes requested):
To sum up our conversation in Slack: vsize is calcuated without the signature which leads to an underestimation.

escottalexander (changes requested):
A few questions/recommendations to look at. Otherwise seems to work as expected.

escottalexander (approved):
LGTM!

nitsujlangston (approved):

PR #3689, opened 2024-01-25
[BWS] Transactions - Create Type 2 tx's

leolambo:


kajoseph (changes requested):
very minor nit picks. Otherwise, it still works w/ the wallet and nothing was broken

kajoseph (approved):
tACK

PR #3692, opened 2024-01-28
Adding ability to derive solona address

micahriggan:


kajoseph (commented):
This seems to function fine. I left a few comments below.

It would be pretty easy to modify `bitcore-mnemonic/lib` to get a ed25519 seed instead of a secp256k1 (Just a one-line change). Otherwise, it would require importing another lib to convert a mnemonic to a seed. :-1:

I started going down the path of making `bitcore-lib` fully SLIP-10 compliant (deriving ed25519 keys) and it's fairly substantial - doable but well beyond the scope of this PR, especially because it would require a move away from the `elliptic` lib (which is probably a good idea anyway).

PR #3706, opened 2024-03-08
[BCC] Add XRP flags methods to bitcore-client; improvements

kajoseph:
This PR

* Primarily adds XRP flag methods to bitcore-client
* Stores the path along w/ the local key/address
* Better wallet check output
* Fixes CWC xrpl export
* Adds a `close` method to storage and calls after all applicable utils. This really only applies to Mongo wallets.


leolambo (approved):
tACK

nitsujlangston (approved):

PR #3714, opened 2024-03-29
[BCN, BWS, CWC, Bitcore-libs] Add named testnets to as networks

escottalexander:
This PR adds support for named testnets. The bulk of the complexity is in alias resolution so that we maintain backwards compatibility.

We should create another PR with front end updates for BWC and Insight.


packages/crypto-wallet-core/src/derivation/paths.ts
```diff
@@ -1,42 +1,27 @@
...
-    mainnet: "m/44'/2'/",
-    livenet: "m/44'/2'/",
-    testnet: "m/44'/2'/",
-    regtest: "m/44'/2'/"
+    default: "m/44'/2'/"
   },
   MATIC: {
-    mainnet: "m/44'/60'/", // the official matic derivation path is 966 but users will expect address to be same as ETH
-    livenet: "m/44'/60'/",
-    testnet: "m/44'/60'/"
```
escottalexander, line 37: Quick thing to note is that MATIC never had a derivation path set for regtest so it always used he default "m/44'/1'/" path. As a result a test had to be updated with the correct address.

packages/bitcore-node/test/integration/matic/p2p.spec.ts
```diff
@@ -85,7 +85,7 @@ describe('Polygon', function() {
     const addresses = await wallet.getAddresses();
     expect(addresses).to.exist;
     expect(addresses.length).to.eq(1);
-    expect(addresses[0].toLowerCase()).to.equal('0xb875c670b079cce8f6ab4f013bd471359c877ac0');
+    expect(addresses[0].toLowerCase()).to.equal('0xa4e131D8c33FC059E9d245489DB03a4A61A2F32B');
```
escottalexander, line 88: Checkout packages/crypto-wallet-core/src/derivation/paths.ts for a comment that explains this otherwise questionable change.

kajoseph (changes requested):
Wanted to go ahead and get this to you. Consider this a WIP review. I haven't gone back through it after discussing the networks with you

kajoseph (changes requested):

packages/bitcore-node/scripts/migrateToNamedTestnets.js
```diff
@@ -0,0 +1,195 @@
...
+const networkMapping = {
+  ETH: {
+    testnet: 'goerli'
+  },
+  BTC: {
+    testnet: 'testnet3'
+  },
+  BCH: {
+    testnet: 'testnet3'
+  }
```
kajoseph, line 27: Need to add other testnets
```
DOGE: {
    testnet: 'testnet3'
  },
  LTC: {
    testnet: 'testnet4',
  },
  MATIC: {
    testnet: 'mumbai'
  }
```

packages/bitcore-wallet-service/src/lib/common/utils.ts
```diff
@@ -275,12 +276,29 @@ export class Utils {
...
 
-  static compareNetworks(network1, network2) {
-    network1 = network1?.toLowerCase();
-    network2 = network2?.toLowerCase();
+  static compareNetworks(network1, network2, chain) {
+    network1 = network1 ? this.getNetworkName(chain, network1.toLowerCase()) : null;
+    network2 = network2 ? this.getNetworkName(chain, network2.toLowerCase()) : null;
 
     if (network1 == network2) return true;
     if (Config.allowRegtest && ['testnet', 'regtest'].includes(network1) && ['testnet', 'regtest'].includes(network2)) return true;
```
kajoseph, line 284: Hmm. We still need to match on network type down here... I wonder if we need a util fn like 
```
function isTestnet(network) {
  return !['regtest', 'mainnet', 'livenet'].includes(network);
}
```

or
```
function getNetworkType(network) {
  if (['mainnet', 'livenet'].includes(network)) {
    return 'mainnet';
  ]
  if (network === 'regtest') {
     return 'regtest';
  }
  return 'testnet';
}
```

packages/bitcore-wallet-service/src/scripts/migrateToNamedTestnets.js
```diff
@@ -0,0 +1,155 @@
...
+const networkMapping = { 
+  eth: {
+    testnet: 'goerli'
+  },
+  btc: {
+    testnet: 'testnet3'
+  },
+  bch: {
+    testnet: 'testnet3'
+  },
```
kajoseph, line 17: need to add other testnets

kajoseph (changes requested):

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1164,7 +1164,7 @@ export class WalletService implements IWalletService {
           return cb(new ClientError('The wallet you are trying to join was created for a different chain'));
         }
 
-        if (!Utils.compareNetworks(wallet.network, xPubKey.network.name)) {
+        if (!xPubKey.network.is(Utils.getGenericName(wallet.network))) {
```
kajoseph, line 1167: I actually don' think this line needs to change. The Utils.compareNetworks() should handle the translation. As this is, it breaks when wallet.network == regtest

kajoseph (changes requested):

packages/bitcore-lib-cash/lib/networks.js
```diff
@@ -77,69 +83,73 @@ function prefixToArray(prefix) {
...
-}
 
-function indexNetworkBy(network, keys) {
-  for(var i = 0; i <  keys.length; i++) {
-    var key = keys[i];
-    var networkValue = network[key];
-    if(!_.isUndefined(networkValue) && !_.isObject(networkValue)) {
-      if(!networkMaps[networkValue]) {
-        networkMaps[networkValue] = [];
+  for (const value of Object.values(network)) {
```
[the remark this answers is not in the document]
escottalexander: What is the point of the indexBy properties on the regtest network?

packages/bitcore-wallet-service/src/lib/common/utils.ts
escottalexander: Good catch. I removed that in err. I feel like the existing getGenericName does the same thing that your second function does.

packages/bitcore-lib-cash/lib/networks.js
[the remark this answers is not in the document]
kajoseph: It's to be able to specify which values you want to add to the networkMap. Yeah, it doesn't seem useful to me either. It could be useful in theory. I'm not sure why regtest is is the only one that has it. Maybe a previous regtest had key values that conflicted w/ testnet? :shrug:

packages/bitcore-wallet-service/src/lib/common/utils.ts
kajoseph: `getGenericName` only returns "testnet" based on Constants.NETWORK_ALIASES. `getGenericName('holesky')` returns "holesky", for example.

Speaking of which, shouldn't `Constants.NETWORK_ALIASES.eth.testnet` be "goerli"?

packages/bitcore-lib-cash/lib/networks.js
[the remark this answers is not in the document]
kajoseph: Or it might have been to make sure that testnet was returned when calling `get` with a key value. The ordering of adding the network should make that unnecessary, though.

I'm fine if you want to remove the indexBy to make the libs more uniform.

PR #3720, opened 2024-04-08
[BCN]: Add Optimism, Arbitrum, and Base support

leolambo:
- Supports the labeling and use of realtime and historical only rpc providers.
- Supports serving chain state from external providers.
- Supports Arbitrum, Base, and Optimism chains.
- Adds scaffolding for External Sync. (Full functionality in subsequent PR)
    - External Sync is designed to sync a chain based on a parameters set in the config
    - This is helpful for sparse syncing. i.e sync 1 transaction every 5 minutes or syncing ONLY blockNumber / 1000
    - Maintaining sparse data for certain chains can help minimize external rpc calls while using minimal storage
    - External Sync is a child of P2P
    
    

**TESTING**

Setup your bitcore.config.json for desired chains.
- chainSource must be 'external' to use external providers, see bitcore-test.config.json
- for valid config values, see packages/bitcore-node/src/types/Config.ts
- Moralis only works for Eth, Matic, Base testnet's but works for all mainnets.

Send GET and POST requests for new chains to your BCN


kajoseph (changes requested):
WIP review

kajoseph (changes requested):

kajoseph (approved):
tACK

PR #3722, opened 2024-04-12
[CWC] Re-add tfFullyCanonicalSig flag to XRP transactions

kajoseph:
`tfFullyCanonicalSig` is deprectated, but removing from CWC causes signature validation failures between client and server. Adding it back in for backward compatibility.

REF: https://xrpl.org/docs/references/protocol/transactions/common-fields/#global-flags


leolambo (approved):
LGTM!

nitsujlangston (approved):

PR #3736, opened 2024-05-16
[BCN] API Stream Error Handling

leolambo:
This PR handles mid-stream errors.

If an error occurs mid stream the following is written to the response: {"error": "An error occurred during data stream"}. Then, the error is logged on the server.


kajoseph (approved):
tACK. Nice work

PR #3739, opened 2024-05-22
[Bitcore] Chain Expansion Improvements

leolambo:
This PR

- Estimates fees for all EVM block types
- Fixes merged streams
- Uses chain for fee estimate
- Improves Axios error logging
- Gets Chain Id at runtime
- Updates default fee values


kajoseph (approved):
LGTM - nice work!

PR #3747, opened 2024-05-30
[BCN] Transform External API Transactions

leolambo:


packages/bitcore-node/src/providers/chain-state/external/streams/apiStream.ts
```diff
@@ -118,7 +123,7 @@ export class ExternalApiStream extends Readable {
           if (opts.jsonl) {
             if (isFirst) {
               isFirst = false;
-            } else {
+            } else if (!data.endsWith('\n')) {
```
kajoseph, line 126: What's the case for this? The newline write below happens _before_ data is written.

currently:
```
data1\n <-- this gets written on the next line just before data2 on line 137
data2
```
but it seems that this change would result in:
```
data1data2\n
```

EDIT: I see why you did this, though this solution looks wrong when reading the code. It assumes that the previous line also had a newline char.  After looking at it some more, we could actually just get rid of the `if (opts.jsonl)` block. The transforms you added to the stream pipeline before this will always have a newline, right? I think at worst we can just have 
```
if (opts.jsonl) {
  data.endsWith('\n') || data += '\n';
}
```
oooor we can just assume it'll always have a newline and get rid of it and change the `else` to `if (!opts.jsonl)`

kajoseph (commented):
Suggestion for a cleanup. The `isFirst` doesn't actually do anything for jsonl...

kajoseph (approved):

PR #3756, opened 2024-06-18
[FIX] import accounts with index > 0 when chain does not match coin

gabrielbazan7:


packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -3096,7 +3096,7 @@ export class API extends EventEmitter {
         const clonedSettings = JSON.parse(JSON.stringify(settings));
         let c = key.createCredentials(null, {
           coin: clonedSettings.coin,
-          chain: clonedSettings.coin, // chain === coin for stored clients
+          chain: clonedSettings.chain, // chain === coin for stored clients. NOT TRUE ANYMORE
```
kajoseph, line 3099: what do you mean by this? It should be true...unless there's a small set of people who created ABR/OP/BASE wallets where this is not true (but it should be true going forward)

kajoseph (approved):

PR #3757, opened 2024-06-21
[BWS, CWC] Xrp payouts

leolambo:


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -4737,15 +4739,18 @@ export class WalletService implements IWalletService {
...
 
-    const sub = TxConfirmationSub.create({
-      copayerId: this.copayerId,
-      walletId: this.walletId,
-      txid: opts.txid,
-      amount: opts.amount,
-      isCreator: true
-    });
+    const txids = _.isArray(opts.txid) ? opts.txid : [opts.txid];
+    for (let i = 0; i < txids.length; i++) {
```
kajoseph, line 4743: Do we need the (i)ndex below or could this be simplified to `for (const txid of txids)`? - total nitpick

packages/crypto-wallet-core/src/transactions/xrp/index.ts
```diff
@@ -10,7 +10,7 @@ enum HashPrefix {
 }
 export class XRPTxProvider {
   create(params: {
-    recipients: Array<{ address: string; amount: string }>;
+    recipients: Array<{ address: string; amount: string; recipientsTag?: number }>;
```
kajoseph, line 13: is there a reason this is `recipientsTag` instead of `recipientTag`? ...or just `tag` so we don't need the `||` below?

packages/bitcore-wallet-service/src/lib/chain/xrp/index.ts
```diff
@@ -244,6 +246,7 @@ export class XrpChain implements IChain {
     const network = tx.network;
     const unsignedTxs = tx.uncheckedSerialize();
     const signedTxs = [];
+    const txids = [];
```
kajoseph, line 249: should this maybe be ` = tx.txids() || [];` so it doesn't overwrite previously added sigs?

packages/bitcore-wallet-service/src/lib/chain/xrp/index.ts
```diff
@@ -136,7 +136,8 @@ export class XrpChain implements IChain {
     const recipients = outputs.map(output => {
       return {
         amount: output.amount,
-        address: output.toAddress
+        address: output.toAddress,
+        recipientTag: output.tag
```
kajoseph, line 140: `tag` is already a param in some places for the destTag (e.g. CWC). Maybe we should keep that var name to simplify passing around instead of having to map between recipientTag <=> tag

kajoseph (changes requested):

packages/bitcore-wallet-service/src/lib/chain/xrp/index.ts
```diff
@@ -137,14 +137,15 @@ export class XrpChain implements IChain {
       return {
         amount: output.amount,
         address: output.toAddress,
-        recipientTag: output.tag
+        tag: output.tag
       };
     });
     const unsignedTxs = [];
     for (let index = 0; index < recipients.length; index++) {
+      const _tag = recipients[0]?.tag || destinationTag
```
kajoseph, line 145: lint: add semicolon

kajoseph (approved):
tACk

PR #3773, opened 2024-07-22
[BWC] Add multisig secret regtest support

kajoseph:


packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -611,10 +611,21 @@ export class API extends EventEmitter {
...
+    switch (network) {
+      case 'testnet':
+        networkChar = 'T';
+        break;
+      case 'regtest':
+        networkChar = 'R';
+        break;
+      default:
+        networkChar = 'L';
+    }
```
MichaelAJay, line 624: Looks fine. Although it would perhaps be more apparent that you're simply performing a mapping action if you did something like:
const networkToNetworkChar = {
  livenet: 'L',
  regtest: 'R',
  testnet: 'T'
};
const networkChar = networkToNetworkChar[network] || networkToNetworkChar.livenet;

Certainly potato/potato. Same goes for the mapping in the other direction below.

MichaelAJay (approved):
Changes are logically consistent. Both present the same default, but with additional checks ('regtest' & 'R' cases, respectively).

MichaelAJay (approved):

nitsujlangston (approved):

PR #3775, opened 2024-07-24
[BWS/BWC] Add taproot wallet support

kajoseph:
Non-multisig taproot wallet support

### DIFF
--- a/packages/bitcore-lib/lib/transaction/input/taproot.js
+++ b/packages/bitcore-lib/lib/transaction/input/taproot.js
@@ -48,10 +48,7 @@ TaprootInput.prototype.getSignatures = function(transaction, privateKey, index, 
     signature: Signature.fromSchnorr(signature),
     sigtype: sigtype
   });
-  if (!this.isValidSignature(transaction, txSig)) {
-    return [];
-  }
-  return [txSig];
+  return this.isValidSignature(transaction, txSig) ? [txSig] : [];
 };
 
 
--- a/packages/bitcore-lib/lib/transaction/input/taproot.js
+++ b/packages/bitcore-lib/lib/transaction/input/taproot.js
@@ -40,14 +40,18 @@ TaprootInput.prototype.getSignatures = function(transaction, privateKey, index, 
   if (!signature) {
     return [];
   }
-  return [new TransactionSignature({
+  const txSig = new TransactionSignature({
     publicKey: privateKey.publicKey,
     prevTxId: this.prevTxId,
     outputIndex: this.outputIndex,
     inputIndex,
     signature: Signature.fromSchnorr(signature),
     sigtype: sigtype
-  })];
+  });
+  if (!this.isValidSignature(transaction, txSig)) {
+    return [];
+  }
+  return [txSig];
 };
 
 
@@ -74,7 +78,7 @@ TaprootInput.prototype.isValidSignature = function(transaction, signature) {
  * @return {boolean}
  */
 TaprootInput.prototype.isFullySigned = function() {
-  return this.output.script.isTaproot() || this.hasWitnesses();
+  return this.output.script.isTaproot() && this.hasWitnesses();
 };
 
 /**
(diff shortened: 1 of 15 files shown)


packages/bitcore-wallet-client/src/lib/common/utils.ts
```diff
@@ -269,6 +269,13 @@ export class Utils {
           );
         }
         break;
+      case Constants.SCRIPT_TYPES.P2TR:
```
MichaelAJay, line 272: In the P2PKH section above, you check state on publicKeys. Is that a behavior check which may be useful for all switch cases (and thus, moved above the switch statement), or are there cases where the first arg (in this case, anyway) can be undefined?

packages/bitcore-lib/lib/transaction/input/taproot.js
```diff
@@ -40,14 +40,18 @@ TaprootInput.prototype.getSignatures = function(transaction, privateKey, index,
...
+  const txSig = new TransactionSignature({
     publicKey: privateKey.publicKey,
     prevTxId: this.prevTxId,
     outputIndex: this.outputIndex,
     inputIndex,
     signature: Signature.fromSchnorr(signature),
     sigtype: sigtype
-  })];
+  });
+  if (!this.isValidSignature(transaction, txSig)) {
```
MichaelAJay, line 51: Just a stylistic consideration:

return this.isValidSignature(transaction, txSig) ? [txSig] : []

Makes it clear at a glance that this is the end of the method and the return is based on a conditional.

packages/bitcore-lib/lib/transaction/input/taproot.js
```diff
@@ -103,4 +107,17 @@ TaprootInput.prototype.addSignature = function(transaction, signature) {
 };
 
 
+// TODO verify that this is the correct MAX size.
```
MichaelAJay, line 110: Is this TODO TODONE?

packages/bitcore-lib/test/transaction/input/taproot.js
```diff
@@ -0,0 +1,127 @@
...
+    const input = transaction.inputs[0];
+    input._estimateSize().should.equal(148);
+  });
+  it('it\'s signature can be removed', function() {
+    const transaction = new Transaction()
+      .from(output)
+      .to(address, 1000000);
+    const input = transaction.inputs[0];
+
+    transaction.sign(privateKey);
```
MichaelAJay, line 60: In the spirit of this test (as "can count missing signatures" above), consider asserting that it is fullySigned before clearing signatures?

packages/bitcore-lib/test/transaction/input/taproot.js
```diff
@@ -0,0 +1,127 @@
...
+    input.clearSignatures();
+    input.isFullySigned().should.equal(false);
+  });
+  it('returns an empty array if private key mismatches', function() {
+    const transaction = new Transaction()
+      .from(output)
+      .to(address, 1000000);
+    const input = transaction.inputs[0];
+    const signatures = input.getSignatures(transaction, new PrivateKey(), 0);
+    signatures.length.should.equal(0);
```
MichaelAJay, line 70: Unlikely - but if signatures were a string, it would also pass. Consider asserting array too

packages/bitcore-lib/test/transaction/input/taproot.js
```diff
@@ -0,0 +1,127 @@
...
+      input.isFullySigned().should.equal(true);
+    });
+    it('it\'s size can be estimated', function() {
+      const transaction = new Transaction()
+        .from(taprootOutput)
+        .to(address, 1000000);
+      const input = transaction.inputs[0];
+      input._estimateSize().should.equal(57.5);
+    });
+    it('it\'s signature can be removed', function() {
```
MichaelAJay, line 92: see pertinent notes above about asserting "isFullySigned true" & array (this test and the next)

packages/bitcore-lib/lib/transaction/input/taproot.js
kajoseph: No. This works for the base case of just a normal wallet address, but if you stuff a tapscript into the p2tr, I doubt this is the max size. I will address this when I implement multisig taproot wallets.

MichaelAJay (approved):

MichaelAJay (approved):

MichaelAJay (approved):
All files approved, functional tests complete

nitsujlangston (approved):

PR #3777, opened 2024-07-28
[BWS] Get Pending Token Tx's

leolambo:
Get pending token txp's to correctly calculate locked balances [here](https://github.com/bitpay/bitcore/blob/master/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts#L99)

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
+++ b/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
@@ -102,15 +102,15 @@ export class EthChain implements IChain {
           // Factor gas used for tokens
           fees += txp.fee;
           // Filter tokens. getPendingTxs returns all txps when given a native currency
-          if (!opts.tokenAddress) {
+          if (txp.tokenAddress && !opts.tokenAddress) {
             return false;
           }
           amounts += txp.amount;
           return true;
         });
         
-        // Do not lock eth multisig amount
-        const lockedSum = opts.multisigContractAddress ? 0 : amounts + fees || 0; // TODO support big int
+        // TODO support big int
+        const lockedSum = (amounts + fees) || 0;  // previously set to 0 if opts.multisigContractAddress
         const convertedBalance = this.convertBitcoreBalance(balance, lockedSum); 
         server.storage.fetchAddresses(server.walletId, (err, addresses: IAddress[]) => {
           if (err) return cb(err);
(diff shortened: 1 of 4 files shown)


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -3307,11 +3305,14 @@ export class WalletService implements IWalletService {
     } else {
       this.storage.fetchPendingTxs(this.walletId, (err, txps) => {
         if (err) return cb(err);
-
+        if (opts.tokenAddress) {
+          txps = txps.filter(txp => opts.tokenAddress === txp.tokenAddress);
+        } else {
+          txps = txps.filter(txp => !txp.tokenAddress);
```
kajoseph, line 3311: This breaks the nonce check in server.ts line 3059. I think we can just remove this line and only filter if opts.tokenAddress was given.

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -3329,9 +3330,7 @@ export class WalletService implements IWalletService {
             });
           },
           err => {
-            txps = _.reject(txps, txp => {
-              return txp.status == 'broadcasted';
-            });
+            txps = txps.filter(txp => txp.status !== 'broadcasted');
```
kajoseph, line 3333: :ok_hand:

kajoseph (changes requested):

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -93,10 +93,27 @@ export class EthChain implements IChain {
...
         if (err) return cb(err);
-        // Do not lock eth multisig amount
-        const lockedSum = opts.multisigContractAddress ? 0 : _.sumBy(txps, 'amount') || 0;
+        let fees = 0;
+        let amounts = 0;
+
+        txps = txps.filter(txp => {
+          // Add gas used for tokens when getting native balance
+          if (!opts.tokenAddress) {
+            fees += txp.fee;
```
kajoseph, line 105: Might need to `|| 0` here. Or are we 100% certain that all txps are going to have fee populated?

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -3307,11 +3305,12 @@ export class WalletService implements IWalletService {
     } else {
       this.storage.fetchPendingTxs(this.walletId, (err, txps) => {
         if (err) return cb(err);
-
+        if (opts.tokenAddress) {
+          txps = txps.filter(txp => opts.tokenAddress === txp.tokenAddress);
```
kajoseph, line 3309: Would be a good idea to ensure there aren't case discrepancies
```
opts.tokenAddress?.toLowerCase() === txp.tokenAddress?.toLowerCase()
```

kajoseph (commented):
Testing now, but so far it looks good. Just a couple of smaller things

kajoseph (approved):
tACK

PR #3778, opened 2024-07-29
[BWS, CWC] Add XRP txp's in output order

leolambo:
In order to support multiple transactions we need to insure the outputs are deterministically added to the txp

### DIFF
--- a/packages/bitcore-wallet-client/src/lib/common/utils.ts
+++ b/packages/bitcore-wallet-client/src/lib/common/utils.ts
@@ -477,7 +477,9 @@ export class Utils {
         multisigContractAddress,
         multiSendContractAddress,
         isTokenSwap,
-        gasLimit
+        gasLimit,
+        multiTx,
+        outputOrder
       } = txp;
       const recipients = outputs.map(output => {
         return {
@@ -511,6 +513,29 @@ export class Utils {
           gasLimit
         };
         unsignedTxs.push(Transactions.create({ ...txp, ...multiSendParams }));
+      } else if (multiTx) {
+        // Add unsigned transactions in outputOrder
+        for (let index = 0; index < outputOrder.length; index++) {
+          const outputIdx = outputOrder[index];
+          if (!outputs?.[outputIdx]) {
+            throw new Error('Output index out of range');
+          }
+          const recepient = {
+            amount: outputs[outputIdx].amount,
+            address: outputs[outputIdx].toAddress,
+            tag: outputs[outputIdx].tag
+          }
+          const _tag = recepient?.tag || destinationTag;
+          const rawTx = Transactions.create({
+            ...txp,
+            ...recepient,
+            tag: _tag ? Number(_tag) : undefined,
+            chain: _chain,
+            nonce: Number(txp.nonce) + Number(index),
+            recipients: [recepient]
+          });
+          unsignedTxs.push(rawTx);
+        }
       } else {
         for (let index = 0; index < recipients.length; index++) {
           const rawTx = Transactions.create({
--- a/packages/bitcore-wallet-service/src/lib/chain/xrp/index.ts
+++ b/packages/bitcore-wallet-service/src/lib/chain/xrp/index.ts
@@ -139,7 +139,7 @@ export class XrpChain implements IChain {
     const length = multiTx ? outputOrder.length : outputs.length;
     for (let index = 0; index < length; index++) {
       let outputIdx = index;
-      if (outputOrder.length) {
+      if (multiTx) {
         outputIdx = outputOrder[index];
       }
       if (!outputs?.[outputIdx]) {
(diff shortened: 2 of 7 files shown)


packages/bitcore-wallet-service/src/lib/chain/xrp/index.ts
```diff
@@ -133,24 +133,30 @@ export class XrpChain implements IChain {
...
-      return {
-        amount: output.amount,
-        address: output.toAddress,
-        tag: output.tag
-      };
-    });
     const unsignedTxs = [];
-    for (let index = 0; index < recipients.length; index++) {
-      const _tag = recipients[0]?.tag || destinationTag;
+    const length = multiTx ? outputOrder.length : outputs.length;
```
kajoseph, line 139: Why is there a difference between outputs.length and outputOrder.length?

For XRP, can't we infer that `multiTx` is true if output[s/Order].length > 1 instead of explicitly passing in `multiTx`? Is there ever a case where an XRP txp has outpus.length > 1 and we _wouldn't_ send out multiple txs?

packages/bitcore-wallet-client/src/lib/common/utils.ts
```diff
@@ -511,6 +513,29 @@ export class Utils {
...
+            address: outputs[outputIdx].toAddress,
+            tag: outputs[outputIdx].tag
+          }
+          const _tag = recepient?.tag || destinationTag;
+          const rawTx = Transactions.create({
+            ...txp,
+            ...recepient,
+            tag: _tag ? Number(_tag) : undefined,
+            chain: _chain,
+            nonce: Number(txp.nonce) + Number(index),
```
kajoseph, line 534: :thinking: This is some hidden logic. So the txp.nonce isn't actually the nonce that gets sent (except the first tx). I guess this is ok in this XRP case... An alternative would be to put the nonces/sequences explicitly on the individual outputs objects, but either way there's going to need to be some special logic around finding the next nonce when "stacking" txps. Something to be aware.

packages/bitcore-wallet-service/src/lib/chain/xrp/index.ts
leolambo: outputOrder previously was used strictly for utxo currencies, for a case that is not clear to me. it did some increments during txp creation that made  outputs.length != outputOrder.length. I used the same variable because the name makes sense. from original txproposal.ts:

```
let numOutputs = x.outputs.length + 1;
if (x.instantAcceptanceEscrow) {
  numOutputs = numOutputs + 1;
}
x.outputOrder = _.range(numOutputs);
```

You are correct in stating XRP txp outpus.length > 1  would always be a multiTx. to me it was cleaner to set and check that variable  instead of checking txp.chain == 'xrp' && txp.outpus.length > 1 in multiple places (see last xrp payout bitcore pr). also would port well if we support any other chain that needs this functionality (god forbid)

packages/bitcore-wallet-client/src/lib/common/utils.ts
kajoseph: Right. All the info is there to be able to determine what the sequence is for each output. I'm just not sure if it's worth being explicit in order to avoid confusion if something goes wrong, or if it hinders extensibility. I'm inclined to say what you have is fine...again the only real point of confusion that I foresee is when trying to determine the next nonce when stacking txps. I would not expect fresh eyes to the system to guess that the next nonce would need to be `txp.nonce + Math.max(txp.outputOrder)`. The worst case, though, is that the nonces overlap and some txs fail to send...which explicitly setting the nonce on the outputs wouldn't prevent, but just potentially reduce the likelihood.

Anyway, I'm not suggesting that one way or the other is better. Just trying to verbally process between the two for your consideration. I'll let you decide which to go with.

packages/bitcore-wallet-service/src/lib/chain/xrp/index.ts
kajoseph: Ah! So, I can give some background on outputOrder. It's meant to randomize the outputs so the change address is not in a predictable position, thus enhancing privacy. And speaking of the change address, outputs.length != outputOrder.length for UTXO txps because the change output is reflected as the changeAddress on the txp instead of being in the outputs array. Since there's no concept as change in XRP land, outputs.length should always equal outputOrder.length.

I think keeping a multiTx flag on the txp is fine... :thinking:  The only thing to consider may be from the perspective of other currencies. A) it might be tempting for someone to set that as true for a UTXO or EVM txp, for example, when what they should really be doing is creating multiple txps. B) it might be confusing to someone else who might wonder "why would I ever want that? Is this a use case I need to consider for X currency?". In those cases, keeping multiTx implicit on the schema and having a helper fn that explains when a txp should be multiTx would be better.

kajoseph (approved):

PR #3781, opened 2024-07-31
feat: Sardine

mahume:
- add methods and API endpoints for fetching supported tokens from Sardine
- add tests for new function

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/expressapp.ts
+++ b/packages/bitcore-wallet-service/src/lib/expressapp.ts
@@ -1745,6 +1745,17 @@ export class ExpressApp {
       });
     });
 
+    router.post('/v1/service/sardine/getSupportedTokens', async (req, res) => {
+      let server, response;
+      try {
+        server = getServer(req, res);
+        response = await server.sardineGetSupportedTokens(req);
+        return res.json(response);
+      } catch (ex) {
+        return returnError(ex, res, req);
+      }
+    });
+
     router.post('/v1/service/sardine/currencyLimits', (req, res) => {
       let server;
       try {
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -5787,6 +5787,40 @@ export class WalletService implements IWalletService {
     });
   }
 
+  sardineGetSupportedTokens(req): Promise<any> {
+    return new Promise((resolve, reject) => {
+      const keys = this.sardineGetKeys(req);
+      const API = keys.API;
+      const CLIENT_ID = keys.CLIENT_ID;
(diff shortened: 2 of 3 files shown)


packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -1745,6 +1745,19 @@ export class ExpressApp {
       });
     });
 
+    router.post('/v1/service/sardine/getSupportedTokens', (req, res) => {
+      getServerWithAuth(req, res, server => {
```
Gamboster, line 1749: I would not use getServerWithAuth to fetch supported tokens/currencies. You can do it directly like this example:

router.post('/v1/service/ramp/assets', async (req, res) => {
      let server, response;
      try {
        server = getServer(req, res);
        response = await server.rampGetAssets(req);
        return res.json(response);
      } catch (err) {
        return returnError(err ?? 'unknown', res, req);
      }
    });

packages/bitcore-wallet-client/src/lib/api.ts
```diff
@@ -3415,6 +3415,15 @@ export class API extends EventEmitter {
     });
   }
 
+  sardineGetSupportedTokens(data): Promise<any> {
```
Gamboster, line 3418: A BWC bump will not be necessary. By not using getServerWithAuth we can access through our BWS URL directly (.../v1/service/sardine/getSupportedTokens). So you can remove this function.

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -5787,6 +5787,41 @@ export class WalletService implements IWalletService {
...
+      const secretBase64 = Buffer.from(secret).toString('base64');
+
+      const headers = {
+        'Content-Type': 'application/json',
+        'Authorization': `Basic ${secretBase64}`,
+      };
+
+      const URL: string = API + '/v1/supported-tokens';
+
+      this.request.post(
```
Gamboster, line 5807: we should use this.request.get(...

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -5787,6 +5787,41 @@ export class WalletService implements IWalletService {
...
+        URL,
+        {
+          headers,
+          body: req.body,
```
Gamboster, line 5811: This method does not requiere body parameters
https://docs.sardine.ai/docs/integrate-payments/on-ramp-api/operations/list-supported-tokens

Gamboster (changes requested):

mahume:
Changes pushed. Please test locally as I am coding this blind without Bitcore running locally.

Gamboster (approved):
LGTM

kajoseph (approved):

PR #3784, opened 2024-08-08
[BCN] EVM external sync

kajoseph:
This PR:

* Refactors module configuration to more granular by being chain-network specific.
* Adds external "syncing" for EVM chains. Syncing is in quote because all it's really doing is listening for blocks, tx, and relevant txs by relevant EVM account and signaling them (this allows the socket to then notify BWS).
* Refactors moralis to be an independent module
* Adds better stop/cleanup logic & logging


packages/bitcore-node/src/providers/chain-state/evm/api/routes.ts
```diff
@@ -178,4 +187,50 @@ export class EVMRouter {
...
+          source: 'moralis',
+          sourceId: req.body.streamId,
+          tag: req.body.tag,
+          body: req.body,
+          timestamp: new Date(),
+          processed: false
+        });
+        return res.end();
+      } catch (err: any) {
+        logger.error('Error processing moralis webhook: %o', err.stack || err.messgae || err);
```
leolambo, line 231: typo, `err.message`

packages/bitcore-node/src/models/webhook.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export class WebhookModel extends BaseModel<IWebhook> {
+  constructor(storage?: StorageService) {
+    super('webhook', storage);
+  }
+  allowedPaging = [];
+
+  onConnect() {
+    // capped at 0.5 million docs
+    this.db?.createCollection(this.collectionName, { capped: true, size: 500000 })
```
leolambo, line 25: Note: MongoDB interprets this as a size in bytes, not a document count. So this is 50 mb

packages/bitcore-node/src/models/webhook.ts
```diff
@@ -0,0 +1,54 @@
...
+      .catch((err) => { if (err.codeName !== 'NamespaceExists') throw err; });
+    this.collection.createIndex({ chain: 1, network: 1, source: 1 });
```
leolambo, line 27: suggestion: run in background to prevent collection locking `this.collection.createIndex({ chain: 1, network: 1, source: 1 }, { background: true });`

packages/bitcore-node/src/models/webhook.ts
```diff
@@ -0,0 +1,54 @@
+import { ObjectID } from 'mongodb';
```
leolambo, line 1: ObjectID is deprecated in the latest MongoDB versions and seldom used in 4.0>, consider using ObjectId

leolambo (changes requested):
Works well just a few smaller changes

packages/bitcore-node/src/models/webhook.ts
kajoseph: Nice catch! Actually that should be 500 kB. I increased to 100 MiB

leolambo (approved):
Awesome work, tACK

nitsujlangston (approved):

PR #3786, opened 2024-08-13
[BCC] Fix PayPro for EVM L2s and make error handling better

kajoseph:


leolambo (approved):
tACK

packages/bitcore-client/bin/wallet-paypro
```diff
@@ -67,9 +67,15 @@ const main = async () => {
   if (!['BTC', 'BCH', 'DOGE', 'LTC'].includes(chain)) {
     nonce = inputNonce || await wallet.getNonce();
   }
-  const currency = token || chain;
+
+  const { data: currencies } = await doRequest({
+    method: 'GET',
+    uri: 'https://bitpay.com/currencies',
```
nitsujlangston, line 73: Shouldn't it get everything it needs from the paypro response?

packages/bitcore-client/bin/wallet-paypro
```diff
@@ -127,23 +133,23 @@ const main = async () => {
...
+    json: { chain, currency, transactions: unsignedTransactions },
+    headers: {
+      'Content-Type' : 'application/payment-verification',
+      'x-paypro-version' : '2'
+    },
+    agentOptions: {
+      rejectUnauthorized: false
+    },
+  });
+  console.log(paymentVerificationResponse);
```
nitsujlangston, line 149: Console log

nitsujlangston (approved):

packages/bitcore-client/bin/wallet-paypro
kajoseph: I'll check, but I don't think the PP response returns currency metadata. This call is to get the native currency for L2 chains (e.g. the OP chain's native currency is ETH_op...which either needs to be hard coded somewhere or I can pull it programmatically from this API call)

kajoseph: Turns out, it's trivial to add the needed metadata to the PP options response. Also, the below logic to find the native currency can be circumvented with --token until the PP options includes the native metadata.

I've removed this API call.

nitsujlangston (approved):

PR #3790, opened 2024-08-29
[All] Clean up dependency tree for Node20+ support

kajoseph:


packages/bitcore-wallet-client/test/paypro.test.js
```diff
@@ -1,14 +1,14 @@
...
+const chai = require('chai');
+const sinon = require('sinon');
+const crypto = require('crypto');
+const should = chai.should();
+const { PayPro: payPro } = require('../ts_build/lib/paypro');
+const TestData = require('./testdata');
 
 function mockRequest(bodyBuf, headers) {
-  bodyBuf = _.isArray(bodyBuf) ? bodyBuf : [bodyBuf];
+  bodyBuf = Array.isArray(bodyBuf) ? bodyBuf : [bodyBuf];
```
MichaelAJay, line 11: This is fine. The spirit of this line could be stated as "If bodyBuf isn't an array, make it one". So I would propose:
if (!Array.isArray(bodyBuf)) {
  bodyBuf = [bodyBuf];
}

Admittedly completely subjective.

packages/bitcore-wallet-service/src/lib/expressapp.ts
```diff
@@ -67,13 +66,17 @@ export class ExpressApp {
...
-    const POST_LIMIT = 1024 * 100 /* Max POST 100 kb */;
+    const POST_LIMIT = 1024 * 100; // Max POST 100 KB
+    const POST_LIMIT_LARGE = 2 * 1024 * 1024; // Max POST 2 MB
 
-    this.app.use(
-      bodyParser.json({
-        limit: POST_LIMIT
-      })
-    );
+    this.app.use((req, res, next) => {
```
MichaelAJay, line 72: What about this?
`    this.app.use((req, res, next) => {
      const postLimit = req.path.includes('/txproposals') ? POST_LIMIT_LARGE : POST_LIMIT;
      return express.json({ limit: postLimit })(req, res, next);
    });`

MichaelAJay: Can't figure out how to format it well. The suggestion is assigning a const with the limit of the ternary, & then just referencing it once w/in the express.json call.

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -338,9 +338,9 @@ export class InternalStateProvider implements IChainStateService {
...
     let totalMissingValue = 0;
     const missingStream = cursor.pipe(
-      through2(
-        { objectMode: true },
-        async (spentCoin: MongoBound<ICoin>, _, done) => {
+      new Transform({
+        objectMode: true,
+        async transform(spentCoin: MongoBound<ICoin>, _, next) {
           if (!seen[spentCoin.spentTxid]) {
             seen[spentCoin.spentTxid] = true;
```
MichaelAJay, line 345: For what it's worth - it seems strange to negative the conditional that got us to this block first thing. I think this belongs at the bottom of the conditional block, since presumably the work done is what marks 'seen' as true.

packages/bitcore-wallet-service/src/lib/expressapp.ts
kajoseph: Use triple backticks for formatting

MichaelAJay (approved):

nitsujlangston (approved):

PR #3791, opened 2024-08-29
Fix schnorr sig to always be 64 bytes.

msinkec:
`k + e*d mod n` is sometimes less than 2^32. In that case we need to pad the buffer with a zero byte.


kajoseph:
@msinkec You'll need to sign your commit before I can merge this in

kajoseph:
Hi @msinkec , would you please sign your commits? This repo requires that all commits be [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) by the authors. Alternatively, I'd be happy to copy your changes into my own commits/PRs if you don't care to take credit for your contributions (both this and your OP_CSV PR).

msinkec:
@kajoseph Done. The commit is signed now. I'll do the same for the other PR shortly.

PR #3800, opened 2024-09-26
[Libs] Add testnet4 to BTC networks

kajoseph:
REF: https://github.com/bitcoin/bips/blob/master/bip-0094.mediawiki


leolambo (approved):
LGTM

PR #3802, opened 2024-10-02
[BWS] Fix walletCheck query param

kajoseph:
'walletCheck' is supposed to be in the query params, not the URI params


bbybobby1 (approved):
LGTM

nitsujlangston (approved):

PR #3803, opened 2024-10-03
[BCN] Add deep dive into string of tx replacements

kajoseph:


packages/bitcore-node/src/services/pruning.ts
```diff
@@ -265,9 +268,23 @@ export class PruningService {
   }
 
   async invalidateTx(chain: string, network: string, tx: ITransaction) {
```
MichaelAJay, line 270: It's not immediately clear that tx is definitely supposed to have `replacedByTxid`, although without it, the behavior is not breaking.

An inline fix would be to change the type on tx:
tx: Omit<ITransaction, 'replacedByTxid'> & { replacedByTxid: string }

Or maybe specified in jsdocs that the expectation is that this method is only called with transactions which definitely have `replacedByTxid`

MichaelAJay (approved):

nitsujlangston (approved):

PR #3817, opened 2024-11-01
fix: ecdsa sign would return r as a BN.js instance

r4mmer:
### Description

The `Q.x` would return a BN without bitcore-lib's methods, this is avoided by calling `getX` instead which converts to the appropriate instance.

### DIFF
--- a/packages/bitcore-lib/lib/crypto/ecdsa.js
+++ b/packages/bitcore-lib/lib/crypto/ecdsa.js
@@ -211,7 +211,7 @@ ECDSA.prototype._findSignature = function(d, e) {
     badrs++;
     k = this.k;
     Q = G.mul(k);
-    r = Q.x.umod(N);
+    r = Q.getX().umod(N);
     s = k.invm(N).mul(e.add(d.mul(r))).umod(N);
   } while (r.cmp(BN.Zero) <= 0 || s.cmp(BN.Zero) <= 0);
 


kajoseph:
Q.x is an instance of `./bn.js` since `./bn.js` is a monkey patch of the dependency `bn.js`. At the point of your change `Q.x instanceof BN` is true and has the monkey patched methods on it. What's your usage?

r4mmer:
Hi @kajoseph ! Sorry for the delayed response.
We use bitcore-lib on a react-native environment and we also need some node apis so we use rn-nodeify for the patches.

The actual issue is a bit complicated, but we use the message [sign](https://github.com/bitpay/bitcore/blob/7735e6eca2af3a348d922004592da571f54e584e/packages/bitcore-lib/lib/message.js#L53)  to validate that the owner of the private key is sending the message.

The validation of the signed message would fail, we tracked the issue to [this line](https://github.com/bitpay/bitcore/blob/master/packages/bitcore-lib-cash/lib/crypto/signature.js#L207) `this.r.toBuffer({ size: 32 })` since the buffer returned here was not 32 bytes long, after some investigation we saw that it is because `r = Q.x.umod(N);` is actually a `bn.js` instance and not a `./bn.js` instance.

The actual issue is that the BN.toBuffer ignores the size argument as you can see below.
image

This makes the `verify` method fail since it tries to read exactly 32 bytes for the `r` and `s`.

r4mmer:
Hi again @kajoseph !
Just pinging to see if I can do anything to help with this issue.

kajoseph:
@r4mmer After looking at it some more, I'm fine with this change. To make it consistent, would you mind also updating the other 3 libs as well? (`bitcore-lib-cash/lib/crypto/ecdsa.js`, `bitcore-lib-ltc...`, `bitcore-lib-doge...`)

kajoseph:
Alternatively, you could change the return statement to return new BN instances like so.
```
return new Signature({
    s: new BN(s),
    r: new BN(r),
    compressed: privkey.publicKey.compressed
  });
```

This may fix https://github.com/bitpay/bitcore/issues/3860 and won't add the overhead of getX()'s cloning and multiple transforming.

PR #3832, opened 2024-12-17
[Libs] Fix PrivateKey instantiation with trimmed hex

kajoseph:
This PR:

* Fixes PrivateKey instantiation with hex strings that are missing leading zeros. See newly added test for an example.
* Adds tests
* Cleans up `new Buffer` (deprecated) and `new BN(Buffer.from` (misleading)


SethMcGuire:
LGTM

nitsujlangston (approved):

PR #3833, opened 2024-12-17
[CWC] Keep private key compressed property in signing

kajoseph:
The `sign` function converts the signing keys to strings which strips the keys of their `compressed` property. Uncompressed keys then are compressed and any inputs in the tx to the uncompressed key aren't signed. This PR fixes that and adds a test. Note that this will now support a mixture of BOTH compressed and uncompressed keys in a single tx


SethMcGuire:
LGTM

nitsujlangston (approved):

PR #3837, opened 2024-12-30
[BCN] Add network to wallet queries for testnet3 -> testnet4 migration

kajoseph:
This is meant to be a temp change so that wallets won't grab testnet3 data during the data purge

### DIFF
--- a/packages/bitcore-node/scripts/migrateTestnetWallets.js
+++ b/packages/bitcore-node/scripts/migrateTestnetWallets.js
@@ -5,11 +5,53 @@ const { WalletAddressStorage } = require('../build/src/models/walletAddress');
 const { Storage } = require('../build/src/services/storage');
 const { wait } = require('../build/src/utils');
 
-const chain = 'BTC';
-const oldNetwork = 'testnet3';
-const newNetwork = 'testnet4';
-const dryRun = true; // make sure you stop sync services before running this script
-const batchSize = 10000;
+function usage(errMsg) {
+  console.log('USAGE: ./migrateTestnetWallets.js [options]');
+  console.log('OPTIONS:');
+  console.log('  --chain <value>         REQUIRED - e.g. BTC, BCH, DOGE, LTC...');
+  console.log('  --oldNetwork <value>    REQUIRED - e.g. testnet3');
+  console.log('  --newNetwork <value>    REQUIRED - e.g. testnet4');
+  console.log('  --batchSize <value>     Number of documents to update at a time. Default: 10000');
+  console.log('  --deDup                 Delete duplicate walletAddresses (can happen if users');
+  console.log('  --doit                  Save the migration to the db. Make sure you stop sync services before running this script');
+  if (errMsg) {
+    console.log('\nERROR: ' + errMsg);
+  }
+  process.exit();
+}
+
+const args = process.argv.slice(2);
+
+if (args.includes('--help') || args.includes('-h')) {
+  usage();
+}
+
+const chainIdx = args.indexOf('--chain');
+const oldNetworkIdx = args.indexOf('--oldNetwork');
+const newNetworkIdx = args.indexOf('--newNetwork');
+const chain = args[chainIdx + 1]?.toUpperCase();
+const oldNetwork = args[oldNetworkIdx + 1]?.toLowerCase();
+const newNetwork = args[newNetworkIdx + 1]?.toLowerCase();
+const deDup = !args.includes('--deDup');
+const dryRun = !args.includes('--doit');
+
+if (
+  chainIdx === -1 ||
+  oldNetworkIdx === -1 ||
+  newNetworkIdx === -1 ||
+  !chain ||
+  !oldNetwork ||
+  !newNetwork
+) {
+  usage('Missing required options.');
+}
+
+const batchSizeIdx = args.indexOf('--batchSize');
+const batchSize = (batchSizeIdx > -1 && parseInt(args[batchSizeIdx + 1])) || 10000;
+if (batchSize > -1 && isNaN(parseInt(args[batchSizeIdx + 1]))) {
+  usage('batchSize must be an integer');
+}
+
 
 let quit = false;
 process.on('SIGINT', () => {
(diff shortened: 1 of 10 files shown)


packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -383,6 +383,9 @@ export class InternalStateProvider implements IChainStateService {
       wallets: wallet._id,
       'wallets.0': { $exists: true }
     };
+    if (wallet.chain === 'BTC' && ['testnet3', 'testnet4'].includes(wallet.network)) {
```
MichaelAJay, line 386: You may avoid confusion if you check this conditional on the initial query assignment (line 380):

```javascript
const query = {
   ...,
   network: wallet.chain === 'BTC' && ['testnet3', 'testnet4'].includes(wallet.network) ? wallet.network : network
}
```

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -591,7 +603,7 @@ export class InternalStateProvider implements IChainStateService {
    *
    * @returns Array<string>
```
MichaelAJay, line 604: Recommend removing jsdoc or changing to parity with TS return type:
`@returns {Promise<Array<string>>}`

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
kajoseph: I don't want to set the network to anything if the condition is not met, including undefined.

packages/bitcore-node/scripts/migrateTestnetWallets.js
```diff
@@ -0,0 +1,68 @@
...
+    if (dryRun) {
+      return;
+    }
+    console.log('----------------------------------------');
+
+    console.log(`Updating ${chain} ${oldNetwork} => ${newNetwork} wallets in 10 seconds...`);
+    !quit && await wait(10000);
+    console.log(`Updating ${chain} ${oldNetwork} wallets...`);
+    let wallets = await WalletStorage.collection.find({ chain, network: oldNetwork }).project({ _id: 1 }).limit(batchSize).toArray();
+    let walletsUpdated = 0;
```
MichaelAJay, line 42: This (and walletAddressesUpdated) is assigned & modified - but never used.

kajoseph: It is used in the log inside the while loop to output the % completed

MichaelAJay (approved):

packages/bitcore-node/scripts/migrateTestnetWallets.js
```diff
@@ -5,11 +5,53 @@ const { WalletAddressStorage } = require('../build/src/models/walletAddress');
...
+  !chain ||
+  !oldNetwork ||
+  !newNetwork
+) {
+  usage('Missing required options.');
+}
+
+const batchSizeIdx = args.indexOf('--batchSize');
+const batchSize = (batchSizeIdx > -1 && parseInt(args[batchSizeIdx + 1])) || 10000;
+if (batchSize > -1 && isNaN(parseInt(args[batchSizeIdx + 1]))) {
```
MichaelAJay, line 51: Seems like you were aiming for a comparison with batchSizeIdx here (that's what the -1 indicates to me).
So I'm clear, the behavior you're going for is:
Default value 10,000, if batchSize arg not included.
If batchSize arg is included & NOT parseInt-able, then exit process with usage message.

Here's a suggestion (if the above is correct):
```javascript
// Default batch size
let batchSize = 10000;
if (batchSizeIdx > -1) {
  const parsedBatchSize = parseInt(args[batchSizeIdx + 1]);
  if (isNan(parsedBatchSize)) { usage('<msg>' }
  else { batchSize = parsedBatchSize }
}
```

packages/bitcore-node/scripts/migrateTestnetWallets.js
```diff
@@ -5,11 +5,53 @@ const { WalletAddressStorage } = require('../build/src/models/walletAddress');
...
-const dryRun = true; // make sure you stop sync services before running this script
-const batchSize = 10000;
+function usage(errMsg) {
+  console.log('USAGE: ./migrateTestnetWallets.js [options]');
+  console.log('OPTIONS:');
+  console.log('  --chain <value>         REQUIRED - e.g. BTC, BCH, DOGE, LTC...');
+  console.log('  --oldNetwork <value>    REQUIRED - e.g. testnet3');
+  console.log('  --newNetwork <value>    REQUIRED - e.g. testnet4');
+  console.log('  --batchSize <value>     Number of documents to update at a time. Default: 10000');
+  console.log('  --deDup                 Delete duplicate walletAddresses (can happen if users');
```
MichaelAJay, line 15: Console log statement seems incomplete? Also (see negation below) - does setting the deDup flag mean that the script SHOULD dedup or that it SHOULDN'T?

packages/bitcore-node/scripts/migrateTestnetWallets.js
```diff
@@ -5,11 +5,53 @@ const { WalletAddressStorage } = require('../build/src/models/walletAddress');
...
+  usage();
+}
+
+const chainIdx = args.indexOf('--chain');
+const oldNetworkIdx = args.indexOf('--oldNetwork');
+const newNetworkIdx = args.indexOf('--newNetwork');
+const chain = args[chainIdx + 1]?.toUpperCase();
+const oldNetwork = args[oldNetworkIdx + 1]?.toLowerCase();
+const newNetwork = args[newNetworkIdx + 1]?.toLowerCase();
+const deDup = !args.includes('--deDup');
```
MichaelAJay, line 35: Why negation?

packages/bitcore-node/scripts/migrateTestnetWallets.js
kajoseph: fixed

packages/bitcore-node/scripts/migrateTestnetWallets.js
```diff
@@ -54,8 +113,14 @@ Storage.start()
     let walletAddresses = await WalletAddressStorage.collection.find({ chain, network: oldNetwork }).project({ _id: 1 }).limit(batchSize).toArray();
     let walletAddressesUpdated = 0;
     while (walletAddresses.length > 0 && !quit) {
-      const res = await WalletAddressStorage.collection.updateMany({ _id: { $in: walletAddresses.map(c => c._id) } }, { $set: { network: newNetwork } });
-      walletAddressesUpdated += res.modifiedCount;
+      try {
+        const res = await WalletAddressStorage.collection.updateMany({ _id: { $in: walletAddresses.map(c => c._id) } }, { $set: { network: newNetwork } });
+        walletAddressesUpdated += res.modifiedCount;
+      } catch (e) {
+        if (e.message.includes('E11000') && deDup) {
```
MichaelAJay, line 120: - If a unique constraint is violated, then it should be, with 100% certainty, because setting network to newNetwork violated the unique constraint.
- This would indicate that network should enforce a unique constraint (which of course seems off)
- But even if it did, we should never try to updateMany with a unique constraint - that's going to fail every time more than one document is updated.

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
MichaelAJay: But you already are setting network above in your initial `query` assignment. Is the difference that it may be null there? If you follow the typings from StreamWalletTransactionsParams, `network` is expected as a string.

kajoseph: Yeah, the query is set based on the API url query param, but that may be wrong during the migration if BWS still has the wallet as testnet3, for example. Also, keeping it this way keeps this temp logic consistent with other parts of this file

MichaelAJay (approved):

nitsujlangston (approved):

PR #3843, opened 2025-01-24
[REF] add missing parameters to paypro response

gabrielbazan7:


packages/bitcore-wallet-client/src/lib/payproV2.ts
```diff
@@ -479,6 +481,14 @@ export class PayProV2 {
       }
     }
 
+    if(responseData.time) {
```
kajoseph, line 484: lint: space after `if`

kajoseph (changes requested):

kajoseph (approved):

PR #3852, opened 2025-02-25
[BWS/BCN] Thorough scanning, cleanup some BWS lodash

kajoseph:
This PR:

* Adds a thorough scanning feature for support wallets to be able to remediate an txs that _should_ be attached to a wallet but for some reason aren't
* Cleans up lodash calls that are easily replaced by native calls (e.g. .filter, .map, .reduce, .slice, loops...)
* Adds types to vars for easier navigation through code


leolambo (approved):
Nice work, tACK

PR #3863, opened 2025-03-14
Ref: move external services to their own file

Gamboster:
Currently all external service functions are within server.ts, as are unit tests in server.js. I refactored this so that each external service used in buy/sell/swap features has its own file for both its functions and unit tests, keeping server.ts a bit more organized.
This shouldn't impact calls from our app or website.


kajoseph (approved):
tACK

Nice work! Thanks for cleaning this up - it's much better!

PR #3865, opened 2025-03-17
[BWS] Update bws email templates

nicholasjs:
closes B1-2358


packages/bitcore-wallet-service/templates/en/new_incoming_tx.html
```diff
@@ -0,0 +1,13 @@
+<div class="status-icon mb-icon">
+    <img src="icons/down_arrow_green.svg" alt="Incoming Transaction Icon" width="70">
```
kajoseph, line 2: this image does not resolve in an email. I'd recommend having this be parameterized and then stuffing the \<svg\> HTML in here along with the title, amount, etc.

kajoseph (changes requested):
I have a PR [here](https://github.com/bitpay/bitcore/pull/3868) that you may find useful for local testing.

packages/bitcore-wallet-service/templates/en/new_incoming_tx.html
nicholasjs: yeah i realized that after testing in email, thanks for this. i ended up adding icon pngs to the cdn in the bitpay repo. do you have a preference for one way or another?

kajoseph: I would rather not involve the bitpay repo if we can help it. Putting the svg into the html removes an unnecessary dependence on bitpay.com

kajoseph (approved):
tACK

Nice work!

PR #3867, opened 2025-03-17
[BWC] Allow getNonce for XRP

leolambo:
We want to be able to create xrp proposals client side with predefined nonce's. As a result, getNonce needs to allow xrp.


kajoseph (approved):
tACK

PR #3868, opened 2025-03-18
[BWS] Cleanup email service

kajoseph:
This cleans up the email service by
* making sendgrid more configurable using either bws.config.js or an env var (existing)
* adding support for another mailer option (for testing)
* removing unnecessary lodash


packages/bitcore-wallet-service/src/lib/emailservice.ts
```diff
@@ -22,13 +24,11 @@ export interface Recipient {
   unit: string;
 }
 
-const Mustache = require('mustache');
-const fs = require('fs');
-const path = require('path');
+const defaultRequest = require('request');
```
MichaelAJay, line 27: Is this necessary, and if so, does it have to use require syntax? Could the other request import be used instead? If not, can we change to import syntax?

MichaelAJay (approved):

nitsujlangston (approved):

PR #3869, opened 2025-03-19
[BWS] Add script to fix EVM address suffix

kajoseph:


kajoseph:
Decided to go with a different fix for this. The address suffix is because there's a unique index on the coin+address in the db, so account-based chains use a :testnet suffix to differentiate them and then strip the suffix in `addressFromStorageTransform`

PR #3878, opened 2025-03-28
[BWC]  BWC cleanup

kajoseph:
This PR

* Cleans up JSDocs
* Removes lodash
* Improves code styling


packages/bitcore-wallet-client/src/lib/request.ts
```diff
@@ -161,7 +185,7 @@ export class Request {
         ret = new Error(
           body.code +
             ': ' +
-            (_.isObject(body.message)
+            (typeof body.message === 'object'
```
gabrielbazan7, line 188: typeof null === 'object' returns true
Maybe we can replace it with: typeof body.message === 'object' && body.message !== null

gabrielbazan7: And then: body.message ?? ''

kajoseph: Good call - missed that one

kajoseph: > And then: body.message ?? ''

Do you mean add a coalesce after the tertiary statement like 
```
(body.message && typeof body.message === 'object'
              ? JSON.stringify(body.message)
              : body.message ?? 'Unknown error')
```
?

gabrielbazan7: yes, that's fine

gabrielbazan7 (approved):
ACK !

nitsujlangston (approved):

PR #3886, opened 2025-04-08
[BWS] Fix noisy logs in bws

kajoseph:
This PR:

* Fixes an issue where it's trying to iterate over a nullish `txp.outputs`
* Short circuits getChain() when coin === undefined instead of throwing and logging, which becomes very noisy


leolambo (approved):
tACK

nitsujlangston (approved):

PR #3888, opened 2025-04-10
[Mnemonic] Support ed25519 seeds

leolambo:


packages/bitcore-lib/lib/hdprivatekey.js
```diff
@@ -417,7 +417,8 @@ HDPrivateKey.fromSeed = function(hexa, network) {
   if (hexa.length > MAXIMUM_ENTROPY_BITS * BITS_TO_BYTES) {
     throw new hdErrors.InvalidEntropyArgument.TooMuchEntropy(hexa);
   }
-  var hash = Hash.sha512hmac(hexa, Buffer.from('Bitcoin seed'));
+  keyType = keyType ? `${keyType} seed` : 'Bitcoin seed';
```
kajoseph, line 420: I'd rather keep bitcore-lib to be bitcoin-only, rather than modifying it to accommodate non-BTC spec.

This fn isn't very large or complex. I suggest we copy the guts of this fn to the mnemonic.js `toHDPrivateKey` fn and mod/generalize it there. That way, we keep bitcore-lib in spec with BTC and de-couple bitcore-mnemonic from being BTC-specific.

kajoseph (changes requested):

kajoseph (approved):
Nice work!

PR #3889, opened 2025-04-14
emoved unused lodash imports in bitcore-node

MicahMaphet:


packages/bitcore-node/src/utils/mergeWith.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @param object destination object
+ * @param source source object
+ * @param customizer function to customize assigned values
+ * @returns object
+ */
+export function mergeWith<TObject, TSource>(
+object: TObject,
+source: TSource,
+customizer: MergeCustomizer
+): TObject & TSource {
```
kajoseph: We actually don't need this fn to be this sophisticated. We're only using it in one place, so we can do a somewhat naive/simple deep merge function. See below

packages/bitcore-node/src/config.ts
```diff
@@ -95,7 +95,7 @@ const Config = function(): ConfigType {
 
   let foundConfig = findConfig();
   const mergeCopyArray = (objVal, srcVal) => (objVal instanceof Array ? srcVal : undefined);
-  config = _.mergeWith(config, foundConfig, mergeCopyArray);
+  config = mergeWith(config, foundConfig, mergeCopyArray);
```
kajoseph: I think we can just do something like this (i have not tested):

```
  function merge(dest: any, src: any) {
    for (const key in src) {
      if (typeof src[key] === 'object' && src[key] !== null && !Array.isArray(src[key])) {
        dest[key] = merge(dest[key], src[key]);
      } else {
        dest[key] = src[key];
      }
    }
    return dest;
  }
  config = merge(config, foundConfig);
```

packages/bitcore-node/src/modules/bitcoin/VerificationPeer.ts
```diff
@@ -222,11 +222,11 @@ export class VerificationPeer extends BitcoinP2PWorker implements IVerificationP
...
       const txs = p2pBlock.transactions ? p2pBlock.transactions.slice(1) : [];
-      const spends = _.chain(txs)
+
+      const spends = txs
         .map(tx => tx.inputs)
-        .flatten()
-        .map(input => input.toObject())
-        .value();
+        .flat()
+        .map(input => input.toObject());
```
kajoseph: can be condensed to:

(one-liner)
```
txs.flatMap(tx => tx.inputs.map(input => input.toObject());
```

(a bit more readable)
```
txs
  .flatMap(tx => tx.inputs)
  .map(input => input.toObject())
```

packages/bitcore-node/src/utils/index.ts
```diff
@@ -83,4 +79,74 @@ export function castToBool(input: any): boolean {
...
+ * For each property in the source object:
+ * It sets the destination object property to the source property unless
+ * both properties are object and the destination object property is not an array.
+ *
+ * @param object destination object
+ * @param source source object
+ */
+export function merge<TDest, TSrc>(dest: TDest, src: TSrc): TDest & TSrc {
+  for (const key in src) {
+    const destProp = dest !== undefined ? (dest as any)[key] : undefined;
```
kajoseph: Can be simplified to `const destProp = dest?.[key];`. But also, when would dest be null or undefined? If it is, it'll break below anyway when you assign `(dest as any)[key] = result`.

packages/bitcore-node/src/utils/index.ts
```diff
@@ -83,4 +79,74 @@ export function castToBool(input: any): boolean {
...
+    const srcProp = src[key];
+    let result;
+    if (srcProp instanceof Object && destProp instanceof Object && !Array.isArray(destProp)) {
```
kajoseph: `instanceof Object` is not very reliable. It's true for classes, Date, String, and other cases. I'd recommend a helper function like so:
```
function isPlainObject(x) {
    return typeof x === 'object' &&
           x !== null &&
           !Array.isArray(x) &&
           Object.getPrototypeOf(x) === Object.prototype;
}
```

packages/bitcore-node/src/utils/index.ts
MicahMaphet: This merge functions starts with the destination object and merge properties from the source object. The destination properties do not necessarily have all the source properties.

kajoseph: Oh ok, I think you meant `dest[key]` is undefined/null, but `dest` should be defined. If `dest` is undefined/null, then `dest[key] = value` will throw an error below.

kajoseph (approved):

PR #3896, opened 2025-04-30
WIP - DNM - chore: remove lerna and use npm workspaces instead

DanielRHayes:
# What?

I propose we move from using `lerna` with `npm` to using `[turbo](https://turborepo.com/docs)` with `[pnpm](https://pnpm.io/)` for the monorepo.  This is a first step towards that goal:

* Step 1: remove `lerna` and use npm workspaces instead <---- this PR
* Step 2: switch from npm workspaces to pnpm workspaces
* Step 3: add `turbo`


DanielRHayes:
Not going to do this, closing

PR #3899, opened 2025-05-05
[BWC, BWS]: Solana Support

leolambo:
Adds support for Solana using SolanaKit

Notes:
Solana wallets are singleAddress. Address derivation happens on BWC.key generation. The public key is stored in clientDerivedPublicKey to be parsed. A completely new address with the same key would require a new wallet with a different account number.

e.g

```
// Wallet 1
clients1.fromString(
  k.createCredentials(null, {
    coin: 'sol',
    network: 'livenet',
    account: 0,
    n: 1
  })
);
clients1.createWallet(
  'mywallet1',
  'creator',
  1,
  1,)

// Wallet 2

clients2.fromString(
  k.createCredentials(null, {
    coin: 'sol',
    network: 'livenet',
    account: 1,
    n: 1
  })
);
clients2.createWallet(
  'mywallet2',
  'creator',
  1,
  1,)
```


packages/bitcore-wallet-client/src/lib/common/utils.ts
```diff
@@ -467,13 +469,17 @@ export class Utils {
       }
       const unsignedTxs = [];
       // If it is a token swap its an already created ERC20 transaction so we skip it and go directly to ETH transaction create
-      const isERC20 = tokenAddress && !payProUrl && !isTokenSwap;
+      const isToken = tokenAddress && !payProUrl && !isTokenSwap;
       const isMULTISIG = multisigContractAddress;
       const chainName = chain.toUpperCase();
+      let tokenType = 'ERC20'
```
MichaelAJay, line 475: tokenType only reassigned (potentially) directly below on a condition - consider using const w/ ternary operator instead.
`const tokenType = chainName === 'SOL' ? 'SPL' : 'ERC20'`

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -43,9 +72,9 @@ export class Key {
   public use44forMultisig: boolean;
   public compliantDerivation: boolean;
   public BIP45: boolean;
-
+  public signatureScheme: string
```
MichaelAJay, line 75: Is this ever used anywhere? If so, is it either ECDSA or EDDSA? If so, or if it has a set of known values, consider a string literal.

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -121,17 +134,21 @@ export class Key {
...
+          this.#setPrivKeyEncrypted({
+            value: sjcl.encrypt(
+              opts.password,
+              xpriv.toString(),
+              opts
+            ),
+            ...params
+          });
+          const xPrivKeyEncrypted = this.#getPrivKeyEncrypted(params);
+          if (xPrivKeyEncrypted) throw new Error('Could not encrypt');
```
MichaelAJay: Is this conditional correct? Should it be !xPrivKeyEncrypted?

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -276,22 +308,24 @@ export class Key {
...
 
-    if (this.isPrivKeyEncrypted()) {
+    if (this.isPrivKeyEncrypted(algo)) {
       $.checkArgument(
         password,
         'Private keys are encrypted, a password is needed'
       );
       try {
-        keys.xPrivKey = sjcl.decrypt(password, this.#xPrivKeyEncrypted);
+        let xPrivKeyEncrypted = this.#getPrivKeyEncrypted({ algo });
```
MichaelAJay, line 324: Never reassigned. Use const

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -302,7 +336,7 @@ export class Key {
         throw new Error('Could not decrypt');
       }
     } else {
-      keys.xPrivKey = this.#xPrivKey;
+      keys.xPrivKey = this.#getPrivKey({ algo });
       keys.mnemonic = this.#mnemonic;
       if (fingerPrintUpdated) {
```
MichaelAJay, line 344: Not your code change - but this is a bug. fingerPrintUpdated is only possibly set to true in the if clause above this else clause. It could never be true here.

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -345,16 +379,29 @@ export class Key {
...
 
-  derive(password, path): Bitcore.HDPrivateKey {
+  derive(password, path, algo?): Bitcore.HDPrivateKey {
     $.checkArgument(path, 'no path at derive()');
-    const xPrivKey = new Bitcore.HDPrivateKey(
-      this.get(password).xPrivKey,
-      NETWORK
-    );
-    const deriveFn = this.compliantDerivation
+    let deriveFn;
```
MichaelAJay, line 387: This is mis-scoped. Put it in the else block where it is used and declare as const.

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -438,10 +492,14 @@ export class Key {
     $.shouldBeUndefined(opts.useLegacyPurpose);
 
     let path = this.getBaseAddressDerivationPath(opts);
-    let xPrivKey = this.derive(password, path);
+    let xPrivKey = this.derive(password, path, algo);
+    let clientDerivedPublicKey;
```
MichaelAJay, line 499: Avoid reassignable declarations when possible. Set as const w/ ternary:
`const clientDerivedPublicKey = algo === Constants.ALGOS.EDDSA ? this.#getChildKeyEDDSA(password, path)?.pubKey : undefined`

While you're at it, `path` is never reassigned above. Update to const.

MichaelAJay: Alternatively, put this closer to its use - when setting args, just set the value there.

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -466,7 +524,8 @@ export class Key {
       keyId: this.id,
       requestPrivKey,
       addressType: opts.addressType,
-      walletPrivKey: opts.walletPrivKey
+      walletPrivKey: opts.walletPrivKey,
+      clientDerivedPublicKey,
```
MichaelAJay, line 531: See comment above for declaration. Consider instead:
`clientDerivedPublicKey: algo === Constants.ALGOS.EDDSA ? this.#getChildKeyEDDSA(password, path)?.pubKey : undefined` to keep it nearest to its single use.

packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -534,6 +593,37 @@ export class Key {
...
+      const addressIndex = 0;
+      const xPrivKey = this.get(password, Constants.ALGOS.EDDSA).xPrivKey
+      const { privKey, pubKey } = Deriver.derivePrivateKey(
+        chain.toUpperCase(),
+        txp.network,
+        xPrivKey, // derived
+        addressIndex,
+        isChange
+      );
+      async.waterfall([
```
MichaelAJay: Already using Promise-chaining, and there's only one promise chain in the array. Is there a need for the waterfall? Should  be able to just return the cb from the then & catch blocks.

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+      if (err) {
+        return cb(err);
+      }
+      // getPendingTxs returns all txps when given a native currency
+      server.getPendingTxs(opts, (err, txps) => {
+        if (err) return cb(err);
+        let fees = 0;
+        let amounts = 0;
+
+        txps = txps.filter(txp => {
```
MichaelAJay, line 48: Filter is being used to mutate external variables, and its output is unused. Consider `for-of` or `reduce`.
`
const { fees, amounts } = txps.reduce((acc, txp) => {
  // Add gas used for tokens when getting native balance
  if (!opts.tokenAddress) {
    acc.fees += txp.fee || 0;
  }
  
  // Filter tokens when getting native balance
  if (!(txp.tokenAddress && !opts.tokenAddress)) {
    acc.amounts += txp.amount;
  }
  
  return acc;
}, { fees: 0, amounts: 0 });
`

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+          }
+          return cb(null, convertedBalance);
+        });
+      });
+    });
+  }
+
+  getFee(server, wallet, opts) {
+    return new Promise(resolve => {
+      const numSignatures = opts.signatures || 1;
```
MichaelAJay, line 84: I can't find where signatures is coming from - but just want to ensure it's a number and not an array.

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+    const unsignedTxs = [];
+    for (let index = 0; index < recipients.length; index++) {
+      let params = {
+        ...recipients[index],
+        recipients: [recipients[index]]
+      };
+      unsignedTxs.push(Transactions.create({ ...txp, chain, ...params }));
+    }
+
+    let tx = {
```
MichaelAJay, line 122: Never reassigned. Use const.

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+    const unsignedTxs = tx.uncheckedSerialize();
+    const signedTxs = [];
+    for (let index = 0; index < signatures.length; index++) {
+      const signed = Transactions.applySignature({
+        chain: this.chain, // TODO use lowercase always to avoid confusion
+        tx: unsignedTxs[index],
+        signature: signatures[index]
+      }); // expected to be in raw string format
+      signedTxs.push(signed);
+      // TDOD id = the base64 version of the signature. no crucial
```
MichaelAJay: You've got a TODO in here. Did you want to clean it up? Or at least fix the typo so it's findable later?

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+  addSignaturesToBitcoreTx(tx, inputs, inputPaths, signatures, xpub) {
+    if (signatures.length === 0) {
+      throw new Error('Signatures Required');
+    }
+
+    const unsignedTxs = tx.uncheckedSerialize();
+    const signedTxs = [];
+    for (let index = 0; index < signatures.length; index++) {
+      const signed = Transactions.applySignature({
+        chain: this.chain, // TODO use lowercase always to avoid confusion
```
MichaelAJay, line 155: Did you want to toLowerCase this?

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+        tx: unsignedTxs[index],
+        signature: signatures[index]
+      }); // expected to be in raw string format
+      signedTxs.push(signed);
+      // TDOD id = the base64 version of the signature. no crucial
+      tx.id = Transactions.getHash({ tx: signed, chain: this.chain });
```
MichaelAJay, line 160: tx.id isn't using w/in the for loop, right? It should be moved outside of the for loop, after it.

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+  getTransactionCount(server, wallet, from) {
+    return new Promise((resolve, reject) => {
+      server._getTransactionCount(wallet, from, (err, count) => {
+        if (err) return reject(err);
+        return resolve(count);
+      });
+    });
+  }
+
+  getWalletSendMaxInfo(server, wallet, opts, cb) {
```
MichaelAJay, line 174: In BtcChain, server is typed as `WalletServer`, which makes 'getBalance' easy to navigate to. If that's what server is here as well, consider updating the typing for better dev ergonomics

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
leolambo: yea its defined in bitcore-node(pr coming soon). fee calculated number of signatures x base fee

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+        utxosBelowFee: 0,
+        amountBelowFee: 0,
+        amount: availableAmount - fee,
+        feePerKb: opts.feePerKb,
+        fee
+      });
+    });
+  }
+
+  checkValidTxAmount(output): boolean {
```
MichaelAJay, line 188: The try/catch block here is inefficient. Also, is 0 a valid transaction amount? In BtcChain, this method checks on <=.

Think about:
`
if (<false conditions>) {
  // log statement (similar to catch block)
  return false
}
return true
`

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+        throw new Error('output.amount is not a valid value: ' + output.amount);
+      }
+      return true;
+    } catch (err) {
+      logger.warn(`Invalid output amount (${output.amount}) in checkValidTxAmount: $o`, err);
+      return false;
+    }
+  }
+
+  checkTx(txp) {
```
MichaelAJay, line 204: You call this.getBitcoreTx, and if it doesn't throw, you return null.
If it does throw, you log and return an error (instead of throwing it).

Is this right?

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    return null;
+  }
+
+  selectTxInputs(server, txp, wallet, _opts, cb) {
+    server.getBalance({ wallet }, (err, balance) => {
+      if (err) return cb(err);
+      const { totalAmount, availableAmount } = balance;
+      // calculate how much spave is needed to find rent amount
+      const minRentException = Defaults.MIN_XRP_BALANCE;
```
MichaelAJay: MIN SOL BALANCE?

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+        return cb(Errors.INSUFFICIENT_FUNDS);
+      } else if (availableAmount < txp.getTotalAmount()) {
+        return cb(Errors.LOCKED_FUNDS);
+      } else {
+        return cb(this.checkTx(txp));
+      }
+    });
+  }
+
+  validateAddress(wallet, inaddr, opts) {
```
MichaelAJay, line 231: Is there any way to differentiate which address wasn't valid? The caller won't ever know if it was to or from.

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+    const bc = server._getBlockchainExplorer(wallet.chain || wallet.coin, wallet.network);
+    bc.getReserve((err, reserve) => {
+      if (err) {
+        return cb(err);
+      }
+      return cb(null, reserve);
+    });
+  }
+
+  getReserve(server: WalletService, wallet: IWallet, cb: (err?, reserve?: number) => void) {
```
MichaelAJay, line 268: Should this not return the valid rent value?

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2621,7 +2639,7 @@ export class WalletService implements IWalletService {
                   next();
                 },
                 async next => {
-                  if (!opts.nonce) {
+                  if (!opts.nonce && !Constants.SVM_CHAINS[wallet.chain.toUpperCase()]) {
```
MichaelAJay, line 2642: Could you put a comment in here on why we skip SOL here?

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
leolambo: thats what getMinimumRent is used for this is xrp only

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
leolambo: this is expected behavior for all chains

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
leolambo: yup, consistent with other chains

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
leolambo: the big number conversion in the if statement can throw an error. see eth implementation

0 is a valid amount as fee is calculated separately. amount is strictly sol or spl being moved fromAddress to toAddress

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
MichaelAJay: Right - I'm just wanting to make sure `opts.signatures` is indeed a number, and not an array of signatures. Just making sure this wasn't supposed to be opts.signatures.length

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
```diff
@@ -0,0 +1,324 @@
...
+      payProUrl,
+      tokenAddress,
+      isTokenSwap,
+      multiTx
+    } = txp;
+    if (multiTx) {
+      throw Errors.MULTI_TX_UNSUPPORTED;
+    }
+    const isSPL = tokenAddress && !payProUrl && !isTokenSwap;
+    const chain = isSPL ? `${this.chain}SPL` : this.chain;
```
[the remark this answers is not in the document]
leolambo: updating CWC to use SOLSPL @gabrielbazan7

packages/bitcore-wallet-service/src/lib/chain/sol/index.ts
leolambo: added `signatures` type to the functions declaration in the interface

gabrielbazan7 (approved):
ACK

packages/bitcore-wallet-service/src/lib/chain/index.ts
```diff
@@ -46,7 +47,7 @@ export interface IChain {
   getChangeAddress(server: WalletService, wallet: IWallet, opts: { changeAddress: string } & any);
   checkDust(output: { amount: number; toAddress: string; valid: boolean }, opts: { outputs: any[] } & any);
   checkScriptOutput(output: { script: string; amount: number; });
-  getFee(server: WalletService, wallet: IWallet, opts: { fee: number; feePerKb: number } & any);
+  getFee(server: WalletService, wallet: IWallet, opts: { fee: number; feePerKb: number; signtures?: number } & any);
```
gabrielbazan7, line 50: typo `signtures`

kajoseph (approved):

PR #3901, opened 2025-05-06
[bitcore-client] add a bin for bitcore-wallet to make it a cli

DanielRHayes:
# What?

* Add a `bin` for bitcore-wallet so you can `npm i -g bitcore-client` and then call `bitcore-wallet`
* Open to other names other then `bitcore-wallet`


kajoseph (commented):
There is already a `bitcore-wallet` bin in the `bitcore-wallet` package (albeit old and outdated). Should we ever dust off that package, it would be confusing to have both be `bitcore-wallet`. It might be better to keep this bin consistent and call it `bitcore-client`

kajoseph (approved):
LGTM

PR #3911, opened 2025-05-20
[Insight] Fix difficulty calculation

MicahMaphet:
Recent blocks in the blockchain where showing a difficulty of ~ 1.29. Fixed the difficulty calculation so it now aligns with other block chain explorers like blockchain.com. I used calculations from https://en.bitcoin.it/wiki/Difficulty.


MicahMaphet:
Note it matches mempool.space
image
image

PR #3919, opened 2025-05-28
[CWC/BCN] Upgrade Web3 to v4

kajoseph:


packages/crypto-wallet-core/src/utils/index.ts
```diff
@@ -52,4 +52,43 @@ export function encodeBuffer(buffer: Buffer, encoding: BufferEncoding | 'base58'
...
\ No newline at end of file
+}
+
+/**
+ * Evaluates if the given string is a valid hex string (with or without 0x prefix)
+ * @param {string} str The string to evaluate
+ * @returns {boolean} True if the string is a valid hex string (with or without 0x prefix)
+ */
+export function isHexString(str: string): boolean {
+  if (!str || typeof str !== 'string') {
```
MichaelAJay: This conditional would be more declarative if it was expressed as:
`typeof str !== 'string' || str === ""`. Technically, the criteria "there is no character in str that is not a valid hex character" is true for the empty string - so changing this conditional more specifically says "exclude the empty string" - the typeof check handles all the other falsy values.

packages/crypto-wallet-core/src/utils/index.ts
```diff
@@ -52,4 +52,43 @@ export function encodeBuffer(buffer: Buffer, encoding: BufferEncoding | 'base58'
...
+ * Converts a number, string, or bigint to a hex string with 0x prefix
+ * This function makes the following assumption(s):
+ * 1. If the input is a non-prefixed numeric string (e.g. '123'), it is assumed to be a decimal number
+ * 2. If the input is a non-prefixed hex string (e.g. '123a'), it is assumed to be a hex number (as opposed to a utf8 string to be converted to hex)
+ * @param {number|string|bigint} input The input number, string, or bigint
+ * @returns {string} The hex string with 0x prefix
+ */
+export function toHex(input: number | string | bigint): string {
+  try {
+    if ((typeof input !== 'number' && typeof input !== 'string' && typeof input !== 'bigint') || input === '') {
```
MichaelAJay: Double error creation - duplicating the error message would improve performance

edit: DEduping the error message, that is

packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -273,19 +268,19 @@ export class RippleStateProvider extends InternalStateProvider implements IChain
...
     const minLedgerIndex = Number(ledgers[0]);
-    const allTxs: AccountTxResponse['result']['transactions'] = [];
+    const allTxs: xrpl.AccountTxResponse['result']['transactions'] = [];
     const limit = Number(limitArg) || 100;
     const options = {
       ledger_index_min: minLedgerIndex,
       limit,
       binary: false
-    } as AccountTxRequest;
+    } as xrpl.AccountTxRequest;
```
MichaelAJay: Consider using `Partial` utility type instead of type assertion

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
```diff
@@ -602,59 +603,63 @@ export class BaseEVMStateProvider extends InternalStateProvider implements IChai
...
 
-    args.endBlock = Math.min(args.endBlock ?? tip, tip);
-    args.startBlock = Math.max(args.startBlock != null ? Number(args.startBlock) : args.endBlock - 10000, 0);
+    endBlock = Utils.BI.min<bigint>([endBlock ?? tip, tip]) as bigint;
+    startBlock = Utils.BI.max<bigint>([startBlock != null ? startBlock : endBlock - 10000n, 0n]) as bigint;
     
-    if (isNaN(args.startBlock!) || isNaN(args.endBlock!)) {
-      throw new Error('startBlock and endBlock must be numbers');
-    } else if (args.endBlock < args.startBlock) {
+    if (endBlock < startBlock!) {
```
MichaelAJay: Make the conditional match the error's omitted message: `startBlock > endBlock`

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
```diff
@@ -690,14 +695,15 @@ export class BaseEVMStateProvider extends InternalStateProvider implements IChai
...
   ) {
     const addresses = await this.getWalletAddresses(walletId);
-    const allTokenQueries = Array<Promise<Array<Partial<Transaction>>>>();
+    const allTokenQueries = Array<Promise<Array<Partial<Web3Types.TransactionInfo>>>>();
     for (const walletAddress of addresses) {
       const transfers = this.getErc20Transfers(network, walletAddress.address, tokenAddress, args);
       allTokenQueries.push(transfers);
     }
     const batches = await Promise.all(allTokenQueries);
     const txs = batches.reduce((agg, batch) => agg.concat(batch));
```
MichaelAJay: A few questions here:
1) this.getErc20Transfers returns `Promise<Array<Partial<Web3Types.Transaction>>>`, but it's pushing to a ...Web3Types.TransactionInfo array.
2) Why the intermediate allTokenQueries array? Why not:
```
const batches = await Promise.all(addresses.map(walletAddress => this.getErc20Transfers(network, walletAddress.address, tokenAddress, args)
```
3) Why txs is batches.reduce instead of batches.flat?

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
```diff
@@ -690,14 +695,15 @@ export class BaseEVMStateProvider extends InternalStateProvider implements IChai
...
-    return txs.sort((tx1, tx2) => tx1.blockNumber! - tx2.blockNumber!);
+    // TODO fix this.
```
MichaelAJay: What needs fixing here?

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
```diff
@@ -932,7 +938,11 @@ export class BaseEVMStateProvider extends InternalStateProvider implements IChai
     return r;
   }
 
-  async _getBlockNumberByDate(params) {
+  async _getBlockNumberByDate(params: {
+    date: Date;
+    network?: string;
+    chainId?: string | bigint;
```
MichaelAJay, line 961: chainId is unused in the method implementation, although it is included in AT LEAST this file. Should you mark as deprecated or unused?

packages/bitcore-node/src/providers/chain-state/evm/api/gnosis.ts
```diff
@@ -61,11 +60,11 @@ export class GnosisApi {
     if (!blockHeight || blockHeight < 0) return Promise.resolve([]);
     const contract = await this.multisigFor(chain, network, gnosisFactory);
     const contractInfo = await contract.getPastEvents('ContractInstantiation', {
-      fromBlock: web3.utils.toHex(blockHeight),
-      toBlock: web3.utils.toHex(blockHeight)
-    });
+      fromBlock: Utils.toHex(blockHeight),
+      toBlock: Utils.toHex((blockHeight))
```
MichaelAJay: Double parens?

packages/bitcore-node/src/providers/chain-state/evm/api/gnosis.ts
```diff
@@ -61,11 +60,11 @@ export class GnosisApi {
...
+    }) as Web3Types.EventLog[];
     return this.convertMultisigContractInstantiationInfo(
-      contractInfo.filter(info => info.returnValues.sender.toLowerCase() === sender.toLowerCase())
+      contractInfo.filter(info => (info.returnValues.sender as string).toLowerCase() === sender.toLowerCase())
```
MichaelAJay: May be a little more verbose, but could make this runtime safe and remove the type assertion with:
`typeof info.returnValues?.sender === 'string' && ...`

packages/bitcore-node/src/providers/chain-state/evm/models/transaction.ts
```diff
@@ -527,33 +529,35 @@ export class EVMTransactionModel extends BaseTransaction<IEVMTransaction> {
     return tx;
   }
 
-  convertRawTx(chain: string, network: string, tx: Partial<ErigonTransaction | GethTransaction>, block?: IEVMBlock): IEVMTransactionInProcess {
+  convertRawTx(chain: string, network: string, tx: Partial<Web3Types.TransactionInfo>, block?: IEVMBlock): IEVMTransactionInProcess {
     if (!block) {
-      const txid = tx.hash || '';
+      const txid = tx.hash as string || '';
```
MichaelAJay, line 534: The contract specifies tx.hash as `string | Uint8Array`

packages/bitcore-node/src/providers/chain-state/evm/models/transaction.ts
```diff
@@ -527,33 +529,35 @@ export class EVMTransactionModel extends BaseTransaction<IEVMTransaction> {
...
       const from = tx.from || '';
-      const value = Number(tx.value);
-      const fee = Number(tx.gas) * Number(tx.gasPrice);
-      const abiType = this.abiDecode(tx.input!);
-      const nonce = tx.nonce || 0;
+      const value = BigInt(tx.value!);
+      const gas = BigInt(tx.gas || -1); // -1 indicates unknown
+      const gasPrice = BigInt(tx.gasPrice || -1);
+      const fee = gas < 0n || gasPrice < 0n ? -1n : gas * gasPrice;
+      const abiType = this.abiDecode(tx.input as string);
```
MichaelAJay, line 541: Same as above - Web3Types.Byte is string or Uint8Array

packages/bitcore-node/src/providers/chain-state/evm/api/gnosis.ts
kajoseph: What do you have against parens?

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
kajoseph: Good catch - it was leftover and I forgot to remove it.

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
kajoseph: IIRC, there was some conflicting behavior with the moralis class' overriding method when this method's definition didn't have the chainId. Something about the overriding method's signature misaligning with this fn's signature. I'll do some testing to see if that's still the case (or maybe i'm misremembering).

packages/bitcore-node/src/providers/chain-state/evm/models/transaction.ts
kajoseph: yes, but it's a string unless specified otherwise in the method call.

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
kajoseph: Ah! Yes, i found the issue and am adding a comment to explain.

MichaelAJay (approved):

PR #3921, opened 2025-06-03
[Insight] Use Arrow to Indicate TX Spent

MicahMaphet:
Arrow linking to the transaction of the output is used to indicate if the output has been spent.
From:
image
To:
image
Spent and unspent transaction outputs align.
Screenshot from 2025-06-03 07-43-24
Nothing particularly different when expanded. Old:
image
New:
image

Added title to output arrow.
image


MicahMaphet:
This is what it looks like with like now.
image

PR #3930, opened 2025-06-10
Draft: [BWS] Refresh transaction proposal on publish

leolambo:
This PR allows a transaction proposal to refresh relevant data on publish.* This is helpful for Solana transactions. Recent blockhash's are necessary to send transactions but blockhash's become invalid after 1 minute. Transaction proposals that sit become stale quickly hence the need for a refresh mechanism.

This PR also allows proposals to be published more than once. This will allow refreshes as needed. For a proposal to be republished it will need `refreshOnPublish: true` set on txp creation. This is set automatically for Solana txps that do not use nonces. Additionally, we save the original raw tx pre refresh for verification consistency. 

With these changes it is recommended to run publish before signing and broadcasting for all BWS assisted Solana transactions.

Additional Changes
- decreasing Solana blockheight cache to maintain fresh data
- adding Solana to 1inch mapping

***This feature will also allow better nonce management for EVM chains as we can update the proposal with a desired nonce right before signing.

### DIFF
--- a/packages/crypto-wallet-core/src/derivation/sol/index.ts
+++ b/packages/crypto-wallet-core/src/derivation/sol/index.ts
@@ -1,6 +1,6 @@
 import { encoding, HDPrivateKey } from 'bitcore-lib';
 import * as ed25519 from 'ed25519-hd-key';
-import { default as Deriver, IDeriver, Key } from '..';
+import { DeriverProxy, IDeriver, Key } from '..';
 
 export class SolDeriver implements IDeriver {
   deriveAddress(_network, _xpubkey, _addressIndex, _isChange) {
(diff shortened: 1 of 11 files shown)


packages/crypto-wallet-core/src/derivation/sol/index.ts
```diff
@@ -30,8 +31,7 @@ export class SolDeriver implements IDeriver {
 
   derivePrivateKey(network, xPriv, addressIndex, isChange, addressType) {
     const changeNum = isChange ? 1 : 0;
-    const pathPrefix = new DeriverProxy().pathFor('SOL', network, addressIndex);
-    const path = `${pathPrefix}/${changeNum}'`;
+    const path = `${Paths['SOL'].default}${addressIndex}/${changeNum}'`;
```
gabrielbazan7: `${Paths['SOL'].default}${addressIndex}'/${changeNum}'` — addressIndex need the ' for hardened derivation

gabrielbazan7 (approved):
ACK. Confirmed that the `Blockhash not found` error is resolved.

PR #3936, opened 2025-06-16
[Insight] Context pill smaller on mobile + sizing and color changes

MicahMaphet:
* Made context pill smaller [1]
* and much smaller on mobile [2]
* Made pill dark for dark mode [3]
* Condensed search placeholder on mobile when searching a currency [4]
* Moved magnifying glass to the left when on mobile when pill is shown and squeezed the search together in general when pill is showing on mobile [5]
* Added scope circle around close button to make it easier to click. Does not show unless hovered over. Easier to click on mobile in particular. [6]

Old            |  New [2][3][4][5]
:-------------------------:|:-------------------------:
image | image

Old: image
[1][3] New: image

[6] 
image

### DIFF
--- a/packages/insight/src/components/pill.tsx
+++ b/packages/insight/src/components/pill.tsx
@@ -1,4 +1,4 @@
-import {FC, useEffect, useState} from 'react';
+import {FC} from 'react';
 import styled from 'styled-components';
 import CloseLightSvg from 'src/assets/images/close-light.svg'
 import {Black, White, Slate30} from '../assets/styles/colors';
@@ -66,6 +66,13 @@ const NetworkLabel = styled.span`
   }
 `;
 
+const CurrencyImg = styled.img`
+  height: 75%;
+  @media screen and (max-width: ${size.mobileL}) {
+    height: 60%;
+  }
+`;
+
 interface PillProps {
   currency?: string,
   network?: string,
@@ -73,18 +80,10 @@ interface PillProps {
 }
 
 export const Pill: FC<PillProps> = ({ currency, network, onCloseClick }) => {
-  const [isMobile, setIsMobile] = useState(window.innerWidth < Number(size.mobileL.slice(0, -2)));
-  
-  useEffect(() => {
-    window.addEventListener('resize', () => {
-      setIsMobile(window.innerWidth < Number(size.mobileL.slice(0, -2)));
-    });
-  }, []);
-
   return (
     currency ?
       <PillBubble>
-        <img src={`https://bitpay.com/img/icon/currencies/${currency}.svg`} alt={currency} style={{height: isMobile ? '60%' : '75%'}} />
+        <CurrencyImg src={`https://bitpay.com/img/icon/currencies/${currency}.svg`} alt={currency} />
         <NetworkLabel>{network}</NetworkLabel>
         <PillCloseButtonScope onClick={onCloseClick}>
           <PillCloseButtonCircle>


MicahMaphet:
I could add titles explaining what the pill does: [pill-titles](https://github.com/MicahMaphet/bitcore/tree/pill-titles)

packages/insight/src/components/pill.tsx
```diff
@@ -1,23 +1,30 @@
...
+import {Black, Slate30, White} from '../assets/styles/colors';
+import { isMobile } from 'src/utilities/helper-methods';
 
 const PillBubble = styled.div`
-  padding: 7px;
-  margin-right: 10px;
   display: flex;
   align-items: center;
-  height: 40px;
+  ${isMobile() ?
```
MichalStachnik: Please do something like this instead:
```
@media screen and (max-width: ${size.size.mobileL}) {}
```
It preserves consistency with how the rest of the project handles mobile styles and utilizes native css media queries.

Check this file for reference: `src/assets/styles/transaction.ts`

packages/insight/src/components/pill.tsx
```diff
@@ -1,23 +1,30 @@
...
+     margin-right: 5px;
+     margin-left: -3px;`
+    :
+    `height: 38px;
+     padding: 5.5px 7px;
+     margin-right: 10px;`
+  }
   border-radius: 25px;
-  background: ${Slate30};
+  background: ${() => useTheme().dark ? '#333' : Slate30};
```
MichalStachnik: Please use this style to remain consistent:
```
background: ${({theme: {dark}}) => (dark ? '#333' : Slate30};
```

packages/insight/src/components/pill.tsx
```diff
@@ -32,13 +39,17 @@ interface PillProps {
...
+  const theme = useTheme();
+
   return (
     currency ?
       <PillBubble>
-        <img src={`https://bitpay.com/img/icon/currencies/${currency}.svg`} alt={currency} style={{height: '120%'}} />
-        <p style={{textTransform: 'capitalize', color: Black, padding: '5px'}}>{network}</p>
+        <img src={`https://bitpay.com/img/icon/currencies/${currency}.svg`} alt={currency} style={{height: '100%'}} />
+        <p style={{textTransform: 'capitalize', color: theme.dark ? White : Black, padding: '5px', fontSize: isMobile() ? '12px' : '16px' }}>
+          {isMobile() ? (network === 'mainnet' ? 'main' : 'test') : network}
```
MichalStachnik: I don't think we should not change the label based on screen size, it should remain either `main` or `mainnet` at all times, but I would defer to @kajoseph here.

packages/insight/src/utilities/helper-methods.ts
```diff
@@ -187,4 +188,8 @@ export const getDifficultyFromBits = (bits: number) => {
   const maxBody = Math.log(0x00ffff);
   const scaland = Math.log(256);
   return Math.exp(maxBody - Math.log(bits & 0x00ffffff) + scaland * (0x1d - ((bits & 0xff000000) >> 24)));
+}
+
+export const isMobile = () => {
```
MichalStachnik: Right now this function is only fired when a component is rendered and does not watch for `window.innerWidth` change. If you wanted a more robust solution you could do something like:
```
useEffect(() => {
    const isMobile = () => window.innerWidth < Number(size.mobileL.slice(0, -2));
    window.addEventListener('resize', isMobile);
    isMobile();
    return () => window.removeEventListener('resize', isMobile);
  }, [])
  ```
  
  Though, I think it's best to just utilize the existing pattern in the project with native media queries like:
  ```
  @media screen and (max-width: ${size.size.mobileL}) {}
  ```

File: packages/insight/src/components/search.tsx [hunk 5]
```diff
@@ -163,7 +164,7 @@ const Search: FC<SearchProps> = ({borderBottom, id, setErrorMessage}) => {
   return (
     <SearchForm onSubmit={search} borderBottom={borderBottom}>
       <span style={{display: 'flex', alignItems: 'center' }}>
-        <img src={searchIcon} alt='Search' style={{padding: 7}}></img>
+        <img src={searchIcon} alt='Search' style={{padding: 7, marginLeft: isMobile() && currency ? '-8px': 0}}></img>
```
MichalStachnik: I would avoid mixing passing function calls and ternaries like this, it's a bit cleaner to just evaluate this before the component return and use that. 

You could also extract this out to its own styled component:
```
const searchImg = styled.img``;
```

packages/insight/src/components/pill.tsx
MicahMaphet: I'll wait on @kajoseph for now.

packages/insight/src/components/search.tsx
```diff
@@ -34,6 +34,14 @@ const SearchForm = styled.form<{ borderBottom?: boolean }>`
     borderBottom ? `1px solid ${colors.borderColor}` : 'none'};
 `;
 
+const SearchImg = styled.img`
+  alt: search;
```
MichalStachnik: I don't think you can add an `alt` tag in css, I would add this back into the html:
```
<SearchImg src={searchIcon} alt="search" />
```

packages/insight/src/components/pill.tsx
kajoseph: I don't have a strong opinion. Functionally, saving 3-4 characters doesn't matter since ppl will pretty much always be pasting into the search bar. It's nice that it's cute and condensed, though.

I asked Brittany what she thought and she voted we keep the full name and "you could also make it not as wide by just using the x icon and not the circle around it (make the x white in mobile so its contrasty enough)"

kajoseph: You could also get rid of the "for" in the search bar placeholder. "Search block, transaction, or address" ...or just replace it with "Search[ on \<chain\> \<network\>]" on mobile.

packages/insight/src/components/search.tsx
```diff
@@ -153,8 +161,12 @@ const Search: FC<SearchProps> = ({borderBottom, id, setErrorMessage}) => {
   }
 
   const searchInputPlaceholder = useMemo(() => {
```
MichalStachnik, line 169: Right now this `useMemo` only gets evaluated on change of `currency` and `network`. So if a user comes to the page on desktop screen and then they make it smaller, the placeholder would not change.

You should add `window.innerWidth` to the useMemo dependency array on Line 175.

PR #3937, opened 2025-06-17
[BWS, CWC] Solana swap support

leolambo:
Updates signing to support dex swaps and lookup table transacitons.
Also adds some improvements for spl tokens

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
+++ b/packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
@@ -625,6 +625,25 @@ export class V8 {
       .catch(cb);
   }
 
+  getRentMinimum(space, cb) {
+    if (!space) {
+      return cb(null, Defaults.MIN_SOL_BALANCE);
+    }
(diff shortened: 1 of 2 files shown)


packages/bitcore-wallet-service/src/externalServices/oneInch.ts
```diff
@@ -119,11 +119,12 @@ export class OneInchService {
           arb: 42161,
           base: 8453,
           op: 10,
+          sol: 501
         };
 
         const chainId = chainIdMap[chain];
 
-        const URL: string = `${credentials.API}/v5.2/${chainId}/tokens`;
+        const URL: string = `https://api.1inch.dev/token/v1.2/${chainId}?provider=1inch&country=US/tokens`;
```
gabrielbazan7: Before, the token data was inside data.body.tokens, but now it’s directly in data.body

gabrielbazan7 (approved):
ACK. I’ve tested the new token info endpoint and confirmed that SLP token transfers are working correctly. The new signing functions also work well.

PR #3942, opened 2025-06-25
[BWC] EDDSA key encryption function support

leolambo:


packages/bitcore-wallet-client/src/lib/key.ts
```diff
@@ -267,6 +253,109 @@ export class Key {
     }
   }
 
+  private setFromExtendedPrivateKey (extendedPrivateKey, opts: { password?: string; algo?: string }) {
```
kajoseph, line 256: lint space between fn name and args

PR #3947, opened 2025-06-27
[BCN] Token instruction improvements

leolambo:
Adds support for decoding Solana lookup table txs and various improvements in transforming SPL token transfer txs

relies on https://github.com/bitpay/crypto-rpc/pull/225


packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -215,15 +215,26 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
...
         let before;
         let count = 0;
+        let _address = address
+        if (tokenAddress) {
+          try {
+            const { rpc } = await this.getRpc(network);
+            _address =  await rpc.getConfirmedAta({ solAddress: address, mintAddress: tokenAddress });
+            if (!_address) throw new Error('Missing ATA');
+          } catch (e: any) {
+            throw new Error('Error getting ata address: %o', e);
```
kajoseph: Looks like a copy n paste error. You logger.error'd before, but now that you're throwing the %o won't be populated by e.

kajoseph (approved):

Issue #3949, opened 2025-06-30
drop lodash


dynst:
https://github.com/bitpay/bitcore/blob/c464b8806a8801a43bea5102ce281fc70045945c/packages/bitcore-lib/package.json#L44

It's an enormous dependency that hasn't had a release since 2021. How much of it is even needed? `_.includes` is just a standard array method now. [`is-what`](https://www.npmjs.com/package/is-what) can be imported for helpers like `_.isObject`. `_.shuffle` is replaceable by [`just-shuffle`](https://www.npmjs.com/package/just-shuffle)

https://github.com/lodash/lodash/issues/5989#issuecomment-2897446421

PR #3950, opened 2025-06-30
remove some lodash calls

dynst:
#3949


kajoseph:
Thanks for the PR @dynst . Before I can merge anything, you'll need to resolve the merge conflict and sign all your commits per the [contributing](https://github.com/bitpay/bitcore/blob/master/CONTRIBUTING.md#contributing-code) guidelines. Additionally, here are a few notes regarding the work:

* We'd like to avoid `.forEach` and instead use for-of loops. `.forEach` loops are relatively slow and antiquated.
* lodash typically handles non-iterable values cleanly (e.g. `_.includes(null, 'someval')` => `false`, `_.each(null, ...)` => `null`) whereas similar declarative calls (e.g. `null.includes('someval')`) will throw an error. This is worth considering in the codebase when null, undefined, or other non-iterable values may be acceptable.

Issue #3820, opened 2024-11-01
Point does not lie on the curve


EnVires:
Error: Point does not lie on the curve
    at Point.validate (G:\Projects JS\cryptapi-main\node_modules\bitcore-lib-cash\lib\crypto\point.js:121:11)
    at new PublicKey (G:\Projects JS\cryptapi-main\node_modules\bitcore-lib-ltc\lib\publickey.js:53:14)
    at PublicKey.fromPrivateKey (G:\Projects JS\cryptapi-main\node_modules\bitcore-lib-ltc\lib\publickey.js:221:10)
    at PrivateKey.toPublicKey (G:\Projects JS\cryptapi-main\node_modules\bitcore-lib-ltc\lib\privatekey.js:349:30)
    at Transaction.getSignatures (G:\Projects JS\cryptapi-main\node_modules\bitcore-lib-ltc\lib\transaction\transaction.js:1200:47)
    at Transaction.sign (G:\Projects JS\cryptapi-main\node_modules\bitcore-lib-ltc\lib\transaction\transaction.js:1189:15)
    at Transactions.buildTransaction (G:\Projects JS\cryptapi-main\dist\lib\ltc\transaction.js:129:21)
    at Transactions.estimateFee (G:\Projects JS\cryptapi-main\dist\lib\ltc\transaction.js:99:34)
    at Transactions.transfer (G:\Projects JS\cryptapi-main\dist\lib\ltc\transaction.js:44:32)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    
Problem with transaction signing
If use this code
`private async _checkBitcore(payload: any, currency: string) {
        const {address} = payload
        const inp = BitcoreInputs.create(this._db, currency)
        const input = await this._getInput(inp, address);
        const utxos = await getUtxos(address, currency.toUpperCase())
        const outs = await this._processUtxos(utxos, BitcoreOutput.create);
        console.log(`Толкаем ${currency}`, input, outs)
        switch (currency) {
            case 'btc':
                return await this._btcTransfer.transfer(input, outs);
            case 'ltc':
                return await this._ltcTransfer.transfer(input, outs);
            case 'bch':
                return await this._bchTransfer.transfer(input, outs);
        }
    }`
    all ok
    But
    if use `import {Transactions} from "../../lib/ltc/transaction"
import {Actor} from "../../lib/actor"
import {
    p2p, Input,
    sendMessage
} from "."
[...]

PR #3952, opened 2025-07-01
fix: Point does not lie on the curve

Sayrix:
Closes https://github.com/bitpay/bitcore/issues/3820

From https://github.com/bitpay/bitcore/pull/3205


Sayrix:
This PR fixed the issue on my end.

PR #3959, opened 2025-07-08
[BCN] Added /:blockId/fee to Block API Route

MicahMaphet:
Added route to get the mean, median, and mode of all the transactions in a block.

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/internal/internal.ts
+++ b/packages/bitcore-node/src/providers/chain-state/internal/internal.ts
@@ -686,8 +686,8 @@ export class InternalStateProvider implements IChainStateService {
     const transactions = blockId.length >= 64 
       ? await TransactionStorage.collection.find({ chain, network, blockHash: blockId }).toArray()
       : await TransactionStorage.collection.find({ chain, network, blockHeight: parseInt(blockId, 10) }).toArray();
-    if (transactions.length === 0)
-      return;
+    if (transactions.length <= 1)
+      return { feeTotal: 0, mean: 0, median: 0, mode: 0 };
 
     let feeRateSum = 0;
     let feeTotal = 0;
@@ -695,17 +695,16 @@ export class InternalStateProvider implements IChainStateService {
     const freq = {};
     let mode = 0, maxCount = 0;
     for (const tx of transactions) {
-      if (tx.fee && tx.size) { // does not add fee rate 0 or divide by zero
-        const rate = tx.fee / tx.size;
-        feeRates.push(rate);
-        feeRateSum += rate;
-        feeTotal += tx.fee;
-        
-        freq[rate] = (freq[rate] || 0) + 1;
-        if (freq[rate] > maxCount) {
-          mode = rate;
-          maxCount = freq[rate];
-        }
+      if (tx.coinbase) continue; // skip coinbase transaction
+      const rate = tx.fee && tx.size ? tx.fee / tx.size : 0; // does not add fee rate 0 or divide by zero
+      feeRates.push(rate);
+      feeRateSum += rate;
+      feeTotal += tx.fee || 0;
+      
+      freq[rate] = (freq[rate] || 0) + 1;
+      if (freq[rate] > maxCount) {
+        mode = rate;
+        maxCount = freq[rate];
       }
     }
     const mean = feeRateSum / feeRates.length;
--- a/packages/bitcore-node/test/integration/routes.spec.ts
+++ b/packages/bitcore-node/test/integration/routes.spec.ts
@@ -43,10 +43,11 @@ async function addTransactions(transactions: {
   chain: 'BTC' | 'BCH';
   fee: number,
   size: number,
-  blockHeight: number 
+  blockHeight: number,
+  coinbase?: boolean
 }[]) {
   for (const tx of transactions) {
-    const { chain, fee, size, blockHeight } = tx;
+    const { chain, fee, size, blockHeight, coinbase } = tx;
     await TransactionStorage.collection.insertOne(
       {
         chain: chain,
@@ -56,7 +57,7 @@ async function addTransactions(transactions: {
         blockHeight: blockHeight,
         blockTime: new Date('2025-07-07T17:38:02.000Z'),
         blockTimeNormalized: new Date('2025-07-07T17:38:02.000Z'),
-        coinbase: true,
+        coinbase: coinbase!!,
         fee: fee,
         inputCount: 1,
         locktime: 0,
@@ -85,19 +86,21 @@ describe('Routes', function() {
       { chain: 'BCH', height: 102 },
     ]);
     await addTransactions([
-      { chain: 'BTC', fee: 0, size: 133, blockHeight: 100 },
+      { chain: 'BTC', fee: 0, size: 133, blockHeight: 100, coinbase: true },
       { chain: 'BTC', fee: 20000, size: 1056, blockHeight: 100 }, 
       { chain: 'BTC', fee: 20000, size: 1056, blockHeight: 100 }, 
       { chain: 'BTC', fee: 25000, size: 1056, blockHeight: 100 }, 
       { chain: 'BTC', fee: 30000, size: 1056, blockHeight: 100 }, 
-      { chain: 'BTC', fee: 35000, size: 1056, blockHeight: 100 }, 
+      { chain: 'BTC', fee: 35000, size: 1056, blockHeight: 100 },
 
-      { chain: 'BTC', fee: 0, size: 133, blockHeight: 103 },
+      { chain: 'BTC', fee: 0, size: 133, blockHeight: 101, coinbase: true },
+
+      { chain: 'BTC', fee: 0, size: 133, blockHeight: 103, coinbase: true },
       { chain: 'BTC', fee: 9000, size: 1056, blockHeight: 103 }, 
       { chain: 'BTC', fee: 10000, size: 1056, blockHeight: 103 }, 
       { chain: 'BTC', fee: 11000, size: 1056, blockHeight: 103 }, 
       
-      { chain: 'BCH', fee: 0, size: 133, blockHeight: 100 },
+      { chain: 'BCH', fee: 0, size: 133, blockHeight: 100, coinbase: true },
       { chain: 'BCH', fee: 2000, size: 1056, blockHeight: 100 }, 
       { chain: 'BCH', fee: 2000, size: 1056, blockHeight: 100 }, 
       { chain: 'BCH', fee: 2500, size: 1056, blockHeight: 100 }, 
@@ -212,6 +215,19 @@ describe('Routes', function() {
         expect(res.body.mean).to.equal((9000 / 1056 + 10000 / 1056 + 11000 / 1056) / 3);
         expect(res.body.median).to.equal(10000 / 1056);
         expect(res.body.mode).to.equal(9000 / 1056);
+        done();
+      });
+    });
+
+    it('should calculate fee data of block with only coinbase transaction as 0', done => {
+      request
+      .get('/api/BTC/regtest/block/101/fee')
+      .expect(200, (err, res) => {
+        if (err) console.error(err);
+        expect(res.body.feeTotal).to.equal(0);
+        expect(res.body.mean).to.equal(0);
+        expect(res.body.median).to.equal(0);
+        expect(res.body.mode).to.equal(0);
         done();
       });
     });


packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +147,41 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
...
+    return res.status(404).send(`block not found with id ${blockId}`);
+
+  let feeTotal = 0;
+  let feeRates: number[] = [];
+  transactions.map((tx) => {
+    const rate = tx.fee / tx.size;
+    feeTotal += rate;
+    if (rate) // does not add fee rate 0
+      feeRates.push(rate);
+  });
```
kajoseph: .map is unnecessary since you aren't mapping to anything. A regular ole `for (const tx of transactions)` will work here.

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +147,41 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
...
+  feeRates.sort((a, b) => a - b);
+  let mean = feeTotal / feeRates.length;
+
+  let median = feeRates.length % 2 === 1
+    ? feeRates[Math.floor(feeRates.length / 2)]
+    : (feeRates[Math.floor(feeRates.length / 2)] + feeRates[Math.ceil(feeRates.length / 2)]) / 2;
+  let mode = feeRates[0], maxCount = 1;
+  let freq = {};
+  feeRates.forEach(rate => {
```
kajoseph: Don't use .forEach. It's slow and antiquated

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +147,41 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
...
+    ? await TransactionStorage.collection.find({ chain, network, blockHash: blockId }).toArray()
+    : await TransactionStorage.collection.find({ chain, network, blockHeight: parseInt(blockId, 10) }).toArray();
+  if (transactions.length == 0)
+    return res.status(404).send(`block not found with id ${blockId}`);
+
+  let feeTotal = 0;
+  let feeRates: number[] = [];
+  transactions.map((tx) => {
+    const rate = tx.fee / tx.size;
+    feeTotal += rate;
```
kajoseph: Might want to put this after the `if (rate)` in case rate is NaN for some reason.

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +147,41 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
...
+      feeRates.push(rate);
+  });
+  feeRates.sort((a, b) => a - b);
+  let mean = feeTotal / feeRates.length;
+
+  let median = feeRates.length % 2 === 1
+    ? feeRates[Math.floor(feeRates.length / 2)]
+    : (feeRates[Math.floor(feeRates.length / 2)] + feeRates[Math.ceil(feeRates.length / 2)]) / 2;
+  let mode = feeRates[0], maxCount = 1;
+  let freq = {};
```
kajoseph: I think these can all be const (except mode)

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +147,41 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
...
+  feeRates.forEach(rate => {
+    freq[rate] = (freq[rate] || 0) + 1;
+    if (freq[rate] > maxCount) {
+      mode = rate;
+      maxCount = freq[rate];
+    }
+  });
+
+  return res.json({mean, median, mode})
```
kajoseph: lint: add spaces between {}'s (e.g `{ mean, median, mode }`. Might as well return feeTotal as well.

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +147,41 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
   }
 });
 
+router.get('/:blockId/feeRate', async function(req: Request, res: Response) {
```
kajoseph: Nit pick: what if we called this endpoint `feeData` instead of `feeRate`? or just `fee`

kajoseph (commented):
Need to test this with other chains

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +147,42 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
...
+  const freq = {};
+  for (const rate of feeRates) {
+    freq[rate] = (freq[rate] || 0) + 1;
+    if (freq[rate] > maxCount) {
+      mode = rate;
+      maxCount = freq[rate];
+    }
+  }
+
+  return res.json({ feeRateTotal, mean, median, mode })
```
kajoseph: Ah, well I don't think feeRateTotal is of any value but feeTotal would be useful (a sum of all tx.fee)

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +147,42 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
...
+  let mode = feeRates[0], maxCount = 1;
+  
+  const freq = {};
+  for (const rate of feeRates) {
+    freq[rate] = (freq[rate] || 0) + 1;
+    if (freq[rate] > maxCount) {
+      mode = rate;
+      maxCount = freq[rate];
+    }
+  }
```
kajoseph, line 166: We can get rid of this loop by combining it with the one above

packages/bitcore-node/src/routes/api/block.ts
kajoseph: Also, lint: add a ;

MicahMaphet:
@kajoseph Should I add caching?

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+import app from '../../src/routes';
+import { intAfterHelper, intBeforeHelper } from '../helpers/integration';
+import { resetDatabase } from '../helpers';
+import { BitcoinBlockStorage } from '../../src/models/block';
+import { expect } from 'chai';
+
+const request = supertest(app);
+
+describe('Routes', function() {
+  this.timeout(500000);
```
kajoseph: Probably don't need this

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      .get('/api/BTC/regtest/block/tip')
+      .set('Accept', 'application/json')
+      .expect(200, (err, res) => {
+        if (err) console.error(err);
+        expect(res.body.height).to.equal(313);
+        done();
+      });
+  });
+
+  it('should respond with a 404 status code for an unknown path', done => {
```
kajoseph: should maybe move this above the first test and then put the block tests inside a `block` describe...just for organization's sake

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +148,49 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
...
+  const { chain, network, blockId } = req.params;
+  const transactions = blockId.length >= 64 
+    ? await TransactionStorage.collection.find({ chain, network, blockHash: blockId }).toArray()
+    : await TransactionStorage.collection.find({ chain, network, blockHeight: parseInt(blockId, 10) }).toArray();
+  if (transactions.length == 0) {
+    logger.error(`block not found with id ${blockId}`);
+    return res.status(404).send(`block not found with id ${blockId}`);
+  }
+    
+  if (feeCache[blockId]) {
```
kajoseph: Let's move the cache check before the transaction storage db call

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+import { expect } from 'chai';
+
+const request = supertest(app);
+
+describe('Routes', function() {
+  this.timeout(500000);
+  before(intBeforeHelper);
+  after(async () => intAfterHelper());
+
+  before(async () => {
```
kajoseph: Let's combine this before hook with the one on line 12

kajoseph (commented):
Looking good so far! Just a few things. Plus, I assume you're wanting to add tests for the new block fee endpoint, right?

packages/bitcore-node/test/integration/routes.spec.ts
MicahMaphet: I get this error when I don't have that line:
```
  1) Routes
       "before all" hook:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (node:internal/timers:614:17)
      at processTimers (node:internal/timers:549:7)
```

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+      request
+      .get('/api/BTC/regtest/block/100/fee')
+      .expect((res) => {
+        // transaction data is defined in before function
+        expect(res.body.feeTotal).to.equal(20000 + 20000 + 25000 + 30000 + 35000)
+        expect(res.body.mean).to.equal((20000 / 1056 + 20000 / 1056 + 25000 / 1056 + 30000 / 1056 + 35000 / 1056) / 5)
+        expect(res.body.median).to.equal(25000 / 1056)
+        expect(res.body.mode).to.equal(20000 / 1056)
+      })
+      .expect(200, done);
```
kajoseph: can you have multiple .expects and are you sure the first one will be completed before the second one is called? If not, done() might be called too soon. 

For clarity and consistency anyway, I'd just call done() at the end of the first expect like in the first block test.

packages/bitcore-node/test/integration/routes.spec.ts
kajoseph: Hmm, ok. Well let's put the timeout inside the before hook then (you'll need to change it from an arrow fn) and change it to something a little more reasonable like 10 seconds instead of over 8 minutes

MicahMaphet: On my computer it consistently breaks with a timeout of 7 seconds and consistently works at 7.5 seconds. I'm not sure how much this could vary or how much of a buffer is necessary. I'll just double what works on my machine.

packages/bitcore-node/test/integration/routes.spec.ts
MicahMaphet: It seems to be waiting for it. When I make it check an absurdly long computation before the average fee checks, it pauses as I would expect. What problems do you think could be caused by the multiple .expects?

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+async function addBlocks(blocks: {
+   height: number,
+}[]) {
+  for (const block of blocks) {
+    const { height } = block;
+    await BitcoinBlockStorage.collection.insertOne(
+      {
+        network: 'regtest',
+        chain: 'BTC',
+        hash: '4c58c6cab141839d66cb99e10757522d379509c9e90a89d39ee990fe6e08ab3a',
```
MicahMaphet, line 23: @kajoseph should I go ahead and make this a parameter or randomly generate the block hash? For my purposes it is not necessary. I could do that for all the data fields or revert back to what I had before without the addBlocks and addTransactions functions.

packages/bitcore-node/test/integration/routes.spec.ts
kajoseph: I'd be concerned about an early done() call which would report the test as passing before the conditions are tested. I've seen this _alot_ in my time. I wouldn't be surprised if there are a couple such tests in this codebase.

packages/bitcore-node/test/integration/routes.spec.ts
kajoseph: Nah, it's fine like this. No need to make things more complicated unless there's a need for non-deterministic values. The addBlock and addTransactions functions are good...they help describe the code a bit.

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+        });  
+    });    
+
+    it('should get block by height', done => {
+      request
+        .get('/api/BTC/regtest/block/101')
+        .expect((res) => {
+          expect(res.body.height).to.equal(101);
+        })
+        .expect(200, done);
```
kajoseph: Here as well

packages/bitcore-node/test/integration/routes.spec.ts
MicahMaphet: I was easily able to make all the tests fail by messing up the expected values. I'm inclined to think that means none of the tests are being skipped (I just added 99 at the end of each expected value):
```
  0 passing (7s)
  4 failing

  1) Routes
       should respond with a 404 status code for an unknown path:
     Error: expected 40499 "undefined", got 404 "Not Found"
      at Test._assertStatus (node_modules/supertest/lib/test.js:266:12)
      at Test._assertFunction (node_modules/supertest/lib/test.js:281:11)
      at Test.assert (node_modules/supertest/lib/test.js:171:18)
      at Server.assert (node_modules/supertest/lib/test.js:131:12)
      at Object.onceWrapper (node:events:627:28)
      at Server.emit (node:events:513:28)
      at emitCloseNT (node:net:2416:8)
      at processTicksAndRejections (node:internal/process/task_queues:89:21)

  2) Routes
       Block
         should respond with a 200 code for block tip and return expected data:

      Uncaught AssertionError: expected 103 to equal 10399
      + expected - actual

      -103
      +10399

      at Test.<anonymous> (test/integration/routes.spec.ts:101:38)
      at Test.assert (node_modules/supertest/lib/test.js:179:6)
      at Server.assert (node_modules/supertest/lib/test.js:131:12)
      at Object.onceWrapper (node:events:627:28)
      at Server.emit (node:events:513:28)
      at emitCloseNT (node:net:2416:8)
      at processTicksAndRejections (node:internal/process/task_queues:89:21)

  3) Routes
       Block
         should get block by height:

      Uncaught AssertionError: expected 101 to equal 10199
      + expected - actual

      -101
      +10199

      at Test.<anonymous> (test/integration/routes.spec.ts:111:38)
      at Test.assert (node_modules/supertest/lib/test.js:179:6)
      at Server.assert (node_modules/supertest/lib/test.js:131:12)
      at Object.onceWrapper (node:events:627:28)
      at Server.emit (node:events:513:28)
      at emitCloseNT (node:net:2416:8)
      at processTicksAndRejections (node:internal/process/task_queues:89:21)

  4) Routes
       Block
         should calculate fee data (total, mean, median, and mode) for block correctly:

      Uncaught AssertionError: expected 130000 to equal 3595099
      + expected - actual

      -130000
      +3595099

      at Test.<anonymous> (test/integration/routes.spec.ts:122:38)
      at Test.assert (node_modules/supertest/lib/test.js:179:6)
      at Server.assert (node_modules/supertest/lib/test.js:131:12)
      at Object.onceWrapper (node:events:627:28)
      at Server.emit (node:events:513:28)
      at emitCloseNT (node:net:2416:8)
      at processTicksAndRejections (node:internal/process/task_queues:89:21)
```

MicahMaphet:
@kajoseph should I add /tip/fee?

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -146,6 +148,41 @@ router.get('/before-time/:time', async function(req: Request, res: Response) {
   }
 });
 
+router.get('/tip/fee', async function(req: Request, res: Response) {
```
kajoseph: This actually can be easily combined with the other endpoint. You'll just need to do something like
```
if (blockId === 'tip') {
  const tip = await ChainStateProvider.getLocalTip({ chain, network });
  if (!tip) {
    return res.status(404).send(`tip not found for ${chain}:${network}`);
  }
  blockId = tip.height.toString();
}
...
```

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -676,4 +676,44 @@ export class InternalStateProvider implements IChainStateService {
...
+    chain: string,
+    network: string,
+    blockId: string
+  }) {
+    const { chain, network, blockId } = params;
+    const transactions = blockId.length >= 64 
+      ? await TransactionStorage.collection.find({ chain, network, blockHash: blockId }).toArray()
+      : await TransactionStorage.collection.find({ chain, network, blockHeight: parseInt(blockId, 10) }).toArray();
+    if (transactions.length === 0)
+      return;
```
kajoseph: We should probably return _something_ here. All zeros would make sense.

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -676,4 +676,44 @@ export class InternalStateProvider implements IChainStateService {
...
+
+    let feeRateSum = 0;
+    let feeTotal = 0;
+    const feeRates: number[] = [];
+    const freq = {};
+    let mode = 0, maxCount = 0;
+    for (const tx of transactions) {
+      if (tx.fee && tx.size) { // does not add fee rate 0 or divide by zero
```
kajoseph: There's an issue here when the only tx is a coinbase tx which has a tx.fee of 0. It results in feeRates.length being 0 which produces a NaN mean and median. We should exclude the coinbase tx, but include 0 fee transactions. Something like this:

```
    if (transactions.length === 0 || transactions.every(tx => tx.coinbase))
      return;

    ...

    for (const tx of transactions) {
      if (tx.coinbase) continue; // skip coinbase transaction
      const rate = tx.fee && tx.size ? tx.fee / tx.size : 0; // does not add fee rate 0 or divide by zero
      feeRates.push(rate);
      feeRateSum += rate;
      feeTotal += tx.fee || 0;
      
      freq[rate] = (freq[rate] || 0) + 1;
      if (freq[rate] > maxCount) {
        mode = rate;
        maxCount = freq[rate];
      }
    }
```

kajoseph (approved):
Nice work! 

tACK

PR #3965, opened 2025-07-25
[BWS] Add feePerKb to Solana getFee

leolambo:
This PR sets Solana's base fee as the feePerKb for the chain. This value does not change. Solana's fee structure charges a base fee per signature as opposed to transaction size used in UTXO and computation needs (e.g gas used) EVM chains. 

There is only one `Fee Level` implemented for Solana. There are no benefits for increasing the fee (e.g speeding up a transaction) therefore no additional fee levels are needed. Faster transactions are achieved through the priority fee  which is an optional params on for Solana BWS transactions proposals. You can get a priority fee based on percentile relative to all other priority fees using the Bitcore Node api endpoint `/api/SOL/:network/priorityFee/:percentile`

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -2575,7 +2575,7 @@ export class WalletService implements IWalletService {
    * @param {string} opts.memo - Optional. Solana transaction memo
    * @param {number} opts.decimals - Optional. Numbet of decimal of a desited token
    * @param {string} opts.fromAta - Optional. ATA addres of the sender (Solana)
-   * * @param {number} opts.numSignatures - Optional. Number of signatures required for the transation. For Solana fee calculation.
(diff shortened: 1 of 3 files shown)


gabrielbazan7 (approved):
LGTM

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -2574,6 +2575,7 @@ export class WalletService implements IWalletService {
    * @param {string} opts.memo - Optional. Solana transaction memo
    * @param {number} opts.decimals - Optional. Numbet of decimal of a desited token
    * @param {string} opts.fromAta - Optional. ATA addres of the sender (Solana)
+   * * @param {number} opts.numSignatures - Optional. Number of signatures required for the transation. For Solana fee calculation.
```
kajoseph: lint

PR #3969, opened 2025-08-04
[BCN] Route tests for /block

MicahMaphet:
Added tests to all /block routes

### DIFF
--- a/packages/bitcore-node/src/routes/api/block.ts
+++ b/packages/bitcore-node/src/routes/api/block.ts
@@ -77,18 +77,13 @@ router.get('/:blockHash/coins/:limit/:pgnum', async function(req: Request, res: 
     console.log(err);
   }
 
+  const numOfTxs = await TransactionStorage.collection.countDocuments({ chain, network, blockHash });
   const skips = maxLimit * (pageNumber - 1);
-  const transactions = await TransactionStorage.collection.find({ chain, network, blockHash });
-  const transactionsArray = await transactions.toArray();
-  const numOfTxs = transactionsArray.length;
   try {
-    const txs =
-      numOfTxs < maxLimit
-        ? transactionsArray
-        : await TransactionStorage.collection.find({ chain, network, blockHash })
-            .skip(skips)
-            .limit(maxLimit)
-            .toArray();
+    const txs = await TransactionStorage.collection.find({ chain, network, blockHash })
+        .skip(maxLimit < numOfTxs ? skips : 0)
+        .limit(maxLimit)
+        .toArray();
 
     if (!txs) {
       return res.status(422).send('No txs for page');
--- a/packages/bitcore-node/src/routes/api/block.ts
+++ b/packages/bitcore-node/src/routes/api/block.ts
@@ -77,10 +77,11 @@ router.get('/:blockHash/coins/:limit/:pgnum', async function(req: Request, res: 
     console.log(err);
   }
 
-  const numOfTxs = await TransactionStorage.collection.countDocuments({ chain, network, blockHash });
   const skips = maxLimit * (pageNumber - 1);
+  const numOfTxs = await TransactionStorage.collection.countDocuments({ chain, network, blockHash });
   try {
-    const txs = await TransactionStorage.collection.find({ chain, network, blockHash })
+    const txs = await TransactionStorage.collection
+        .find({ chain, network, blockHash })
         .skip(maxLimit < numOfTxs ? skips : 0)
         .limit(maxLimit)
         .toArray();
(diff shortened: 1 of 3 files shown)


packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -77,14 +77,15 @@ router.get('/:blockHash/coins/:limit/:pgnum', async function(req: Request, res:
...
+  const numOfTxs = transactionsArray.length;
   try {
-    let txs =
+    const txs =
       numOfTxs < maxLimit
-        ? await TransactionStorage.collection.find({ chain, network, blockHash }).toArray()
-        : await TransactionStorage.collection
-            .find({ chain, network, blockHash })
+        ? transactionsArray
+        : await TransactionStorage.collection.find({ chain, network, blockHash })
```
kajoseph: Ideally, evaluating the numOfTxs first would prevent us from using gobs of memory by pulling _all_ transactions to see if we even want them all.

However, I don't see a reason why we couldn't simply do 1 call: this one on line 88. If `numTxs < maxLimit` is true, then `.limit(maxLimit)` will have no effect anyway.

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -77,34 +77,31 @@ router.get('/:blockHash/coins/:limit/:pgnum', async function(req: Request, res:
     console.log(err);
   }
 
-  let skips = maxLimit * (pageNumber - 1);
-  let numOfTxs = await TransactionStorage.collection.find({ chain, network, blockHash }).count();
+  const skips = maxLimit * (pageNumber - 1);
+  const numOfTxs = await TransactionStorage.collection.countDocuments({ chain, network, blockHash });
```
MicahMaphet, line 81: I decided to use .countDocuments because .count is deprecated. Both worked for the tests so I don't have a problem changing it back.

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -7,227 +7,539 @@ import { intAfterHelper, intBeforeHelper } from '../helpers/integration';
...
 
+const expectObjectToHaveProps = (obj: any, props: Record<string, string>) => {
+  expect(obj).to.include.all.keys(Object.keys(props));
+  for (const key in props) {
+    expect(obj[key]).to.be.a(props[key]);
+  }
+};
+
+const minutesAgo = (minutes: number): Date => new Date(Date.now() - 1000 * 60 * minutes);
+
```
MicahMaphet: I could move these functions to test/helpers

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -7,227 +7,539 @@ import { intAfterHelper, intBeforeHelper } from '../helpers/integration';
...
     sandbox.restore();
   });
-  
+
   it('should respond with a 404 status code for an unknown path', done => {
     request.get('/unknown').expect(404, done);
   });
 
   describe('Block', function() {
+    const testBlock = (block: any) => {
```
MicahMaphet, line 186: I personally found it to be readable and intuitive to have the functions specific to a group of tests to be defined right above those tests. However this might be bad practice. I find it intuitive because its similar to a after all hook but only for the following group of tests in the since that it is the base test that is run on all the tests in the group.

packages/bitcore-node/test/integration/routes.spec.ts
```diff
@@ -7,227 +7,539 @@ import { intAfterHelper, intBeforeHelper } from '../helpers/integration';
...
+        if (err) console.error(err);
+        const { txids, inputs, outputs } = res.body;
+        expect(txids).to.be.empty;
+        expect(inputs).to.be.empty;
+        expect(outputs).to.be.empty;
+        done();
+      });
+    });
+
+    for (let i = 1; i <= 3; i++) {
```
MicahMaphet, line 445: I hope this isn't annoying/convoluted. I was trying to reduce the file size. It kind of makes it easier to read because it indicates that these three it-should statements are really one test.

packages/bitcore-node/src/routes/api/block.ts
```diff
@@ -77,34 +77,31 @@ router.get('/:blockHash/coins/:limit/:pgnum', async function(req: Request, res:
...
-      numOfTxs < maxLimit
-        ? await TransactionStorage.collection.find({ chain, network, blockHash }).toArray()
-        : await TransactionStorage.collection
-            .find({ chain, network, blockHash })
-            .skip(skips)
-            .limit(maxLimit)
-            .toArray();
+    const txs = await TransactionStorage.collection
+        .find({ chain, network, blockHash })
+        .skip(maxLimit < numOfTxs ? skips : 0)
```
kajoseph: This is odd behavior. I would expect the code to be obedient if I told it to skip x records.

packages/bitcore-node/test/integration/routes.spec.ts
kajoseph: This is fine. You might consider adding a comment that explains the purpose to save some time for future readers.

PR #3970, opened 2025-08-05
[BCN] Decode raw transaction Solana endpoint

leolambo:


packages/bitcore-node/src/providers/chain-state/svm/api/routes.ts
```diff
@@ -89,7 +90,7 @@ export class SVMRouter {
         if (err?.code != null) {
           res.status(400).send(err.message);
         } else {
-          logger.error('Rent Error::%o', err.stack || err.message || err);
+          logger.error('Rent Error: %o', err.stack || err.message || err);
```
kajoseph: Is this an accurate error message prefix?

PR #3974, opened 2025-08-08
Chain details page

MicahMaphet:
Added a page with general chain information and a list of blocks with expandable data.
The chain information includes:
- 2 graphs, price over the last 24 hours (same data as the front page) and fee data over the last 128 blocks
- Both graphs have adjustable time scales using the dropdown
- Above both graphs is the amount the fee or price has changed in the time scale

When the rows are expanded, block data excluding transactions is shown
image
image

### DIFF
--- a/packages/insight/src/components/graph-metadata-spans.tsx
+++ b/packages/insight/src/components/graph-metadata-spans.tsx
@@ -8,7 +8,7 @@ export const FeeMetadataSpan: FC<{ range: string }> = ({ range }) => {
 
   const numBlocks = Number(range.slice(0, range.indexOf(' ')));
   const fees = blocks
-    .slice(blocks.length - numBlocks)
+    .slice(0, numBlocks)
     .map(block => block.feeData.median);
 
   let total = 0;


packages/insight/src/utilities/helper-methods.ts
```diff
@@ -183,8 +183,64 @@ export const getLib = (currency: string) => {
   }
 };
 
+export const getName = (currency: string) => {
+  switch (currency.toUpperCase()) {
+    case 'BTC':
+    default:
```
MichalStachnik: It's best practice to have the default at the end of the switch, so perhaps change this to
```
export const getName = (currency: string) => {
  switch (currency.toUpperCase()) {
    case 'BTC':
      return 'Bitcoin';
    case 'BCH':
      return 'Bitcoin Cash';
    case 'DOGE':
      return 'Doge';
    case 'LTC':
      return 'Litecoin';
    case 'ETH':
      return 'Ethereum';
    default:
      return 'Bitcoin'; // or throw an error for unknown currencies
  }
}
```

packages/insight/src/pages/blocks.tsx
```diff
@@ -1,188 +1,57 @@
...
       })
-      .catch((e: any) => {
-        setError(e.message || 'Something went wrong. Please try again later.');
-      })
       .finally(() => {
-        setIsLoading(false);
         nProgress.done();
+      })
+      .catch((e: any) => {
+        setError(e.message || 'Something went wrong. Please try again later.');
```
MichalStachnik, line 35: I think it's better to follow convention here and have your promise handling be .then(), .catch(), .finally().

This way we handle success first, then handle errors, then cleanup.

packages/insight/src/components/currency-tile.tsx
```diff
@@ -197,12 +197,12 @@ const CurrencyTile: FC<CurrencyTileProps> = ({currency}) => {
   const {height, time, transactionCount, size} = data[0];
   const imgSrc = `https://bitpay.com/img/icon/currencies/${currency}.svg`;
 
-  const gotoAllBlocks = async () => {
```
MichalStachnik, line 200: I don't think this method needs to be `async/await`

packages/insight/src/components/chain-header.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const priceList = (priceDisplay?.data?.[0]?.priceDisplay || []);
+
+  const feeRanges = ['128 Blocks', '32 Blocks', '16 Blocks', '8 Blocks'];
+  const priceRanges = ['24 Hours', '12 Hours', '6 Hours', '3 Hours'];
+
+  const [feeSelectedRange, setFeesSelectedRange] = useState('32 Blocks');
+  const [priceSelectedRange, setPriceSelectedRange] = useState('24 Hours');
+  
+  const [feeChangeSpan, setFeeChangeSpan] = useState(() => { return <span>null</span>; });
+  const [priceChangeSpan, setPriceChangeSpan] = useState(() => { return <span>null</span>; });
```
MichalStachnik: I think this is a bit of a react anti-pattern. I would just make these 2 spans their own components with their own files. You could get rid of the useEffects and just move that logic into the smaller components and pass everything you need into them as props.

This will also reduce unnecessary re-rendering of this parent component when the useEffect's are re-ran each time.

packages/insight/src/components/block-list.tsx
```diff
@@ -0,0 +1,204 @@
...
+                  columnRightExpandedDataKeys = ['Next block', 'Nonce', 'Confirmations'];
+                } else {
+                  columnLeftExpandedDataKeys = ['Previous block', 'Bits', 'Version', 'Block reward', 'Miner fees'];
+                  columnRightExpandedDataKeys = ['Next block', 'Nonce', 'Confirmations', 'Difficulty', 'Fee data'];
+                }
+                const columnLeftExpandedData : Array<{label: string, value: any}> = columnLeftExpandedDataKeys.map(key => dataRowsDB[key]);
+                const columnRightExpandedData : Array<{label: string, value: any}> = columnRightExpandedDataKeys.map(key => dataRowsDB[key]);
+
+                return (
+                  <React.Fragment key={index}>
```
MichalStachnik, line 155: Let's use `<> </>` instead of `React.Fragment` to remain consistent.

packages/insight/src/components/block-list.tsx
```diff
@@ -0,0 +1,204 @@
...
+        setBlocksList(_blocksList.concat(newData));
+      } else {
+        setHasMore(false);
+      }
+    } catch (e: any) {
+      setError(e.message || 'Something went wrong. Please try again later.');
+    }
+  };
+
+  const gotoSingleBlockDetailsView = async (hash: string) => {
```
MichalStachnik: I don't _think_ this needs to be `async/await`

packages/insight/src/components/block-list.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+const getBlocksUrl = (currency: string, network: string) => {
+  return `${getApiRoot(currency)}/${currency}/${network}/block?limit=200`;
+};
+
+const BlockList: FC<{currency: string, network: string, blocks: Array<BitcoinBlockType & Partial<FeeData>>}> = ({currency, network, blocks}) => {
+  const theme = useTheme();
+  const navigate = useNavigate();
+  const [expandedBlocks, setExpandedBlocks] = useState<number[]>([]);
+  const [blocksList, setBlocksList] = useState(blocks);
```
MichalStachnik: I'm not sure you want or even need to be doing this. You already have `blocks` available as a prop coming in from a higher state, I think just use that instead of setting a local copy. This way you have one source of truth.

If you need to derive a new `blocksList` from `blocks` I would reach for `useMemo`.

packages/insight/src/Routing.tsx
```diff
@@ -1,7 +1,7 @@
-import React, {lazy, Suspense} from 'react';
+import {lazy, Suspense} from 'react';
 import {Navigate, Route, Routes} from 'react-router-dom';
 import Home from './pages';
-const Blocks = lazy(() => import('./pages/blocks'));
```
MichalStachnik, line 4: Is there a reason for removing the lazy loading of this component?

packages/insight/src/components/block-list.tsx
MicahMaphet: I believe you need React.Fragments for the key property since it is in a list of react fragments
https://react.dev/reference/react/Fragment#:~:text=Here%E2%80%99s%20a%20situation,the%20key%20attribute%3A

MichalStachnik: Ah yes, let's keep it.

packages/insight/src/Routing.tsx
MicahMaphet: Not really. Its some kind of refactoring artifact.

MichalStachnik: Ok in that case I would keep it, it helps reduce the size of our initial javascript bundle.

packages/insight/src/components/block-list.tsx
MicahMaphet: I changed it to utilize useContext. I liked doing it this way but it is the first instance of it in insight.

packages/insight/src/pages/blocks.tsx
```diff
@@ -1,188 +1,79 @@
...
-`;
+type BlocksContextType = {
+  blocks: BitcoinBlockType[] | undefined;
+  setBlocks: React.Dispatch<React.SetStateAction<BitcoinBlockType[] | undefined>>;
+};
 
-const TdLink = styled.td`
-  color: ${({theme: {colors}}) => colors.link};
-`;
+const BlocksContext = createContext<BlocksContextType | undefined>(undefined);
```
MichalStachnik: Please add this to a new directory `src/contexts`

packages/insight/src/components/block-list.tsx
```diff
@@ -0,0 +1,207 @@
...
+                } else {
+                  columnLeftExpandedDataKeys = ['Previous block', 'Bits', 'Version', 'Block reward', 'Miner fees'];
+                  columnRightExpandedDataKeys = ['Next block', 'Nonce', 'Confirmations', 'Difficulty', 'Fee data'];
+                }
+                const columnLeftExpandedData : Array<{label: string, value: any}> = columnLeftExpandedDataKeys.map(key => dataRowsDB[key]);
+                const columnRightExpandedData : Array<{label: string, value: any}> = columnRightExpandedDataKeys.map(key => dataRowsDB[key]);
+
+                return (
+                  <React.Fragment key={index}>
+                    <BlockListTableRow key={index}>
```
MichalStachnik: I think we only need the outer one to have the `key`

packages/insight/src/components/block-list.tsx
```diff
@@ -0,0 +1,207 @@
...
+    } catch (e: any) {
+      setError(e.message || 'Something went wrong. Please try again later.');
+    }
+  };
+
+  const gotoSingleBlockDetailsView = (hash: string) => {
+    navigate(`/${currency}/${network}/block/${hash}`);
+  };
+
+  if (!blocks?.length) return null;
```
MichalStachnik: You have a similar check on line 41, I would try to consolidate them into one check. I think this one before the return is a preferred location.

kajoseph (commented):
Overall, this looks great! Nice work! I do have a couple of comments:

* [x] The tip has a next block link that throws a nasty uncaught error when clicked. We should at least remove the link...maybe gray out the link arrow and put a tooltip on it.
* [x] The block number blue looks rather dark in dark mode. It's also inconsistent with the blue on the block details/transactions page. We should make these blues consistent between light/dark modes.
* [x] The fee rate data at the top seems odd to me. I don't know that it's particularly helpful to have a diff for the past X blocks. It appears to be essentially a copy n paste from the Exchange Rate graph, but for fee rate data I think it makes more sense to just display as an average across the past X blocks instead of a diff.
* [x] The graphs seem to be off by 1. For example: the 3 hour graph shows 2 hours of data.

PR #3977, opened 2025-08-11
[CLI] Add command mode, erc20 support, message signing

kajoseph:
This PR:
- Adds "command mode" to be able run bitcore-cli non-interactively with the `--command` option
- Adds support for ERC20 tokens
- Adds arbitrary message signing
- Fixes key derivation for TSS keys to be chain-specific
- Fixes wallet tx history for ERC20 transactions
- Does a lot of refactoring and improving on typing and flow


packages/bitcore-cli/src/cli-commands.ts
```diff
@@ -0,0 +1,39 @@
+import { type ICliOptions } from '../types/cli';
+import { type IWallet } from '../types/wallet';
+import { Utils } from './utils';
+
+export function getCommands(args: { wallet: IWallet, opts?: ICliOptions }) {
+  const { wallet, opts } = args;
```
MichaelAJay: opts unused

packages/bitcore-cli/src/cli.ts
```diff
@@ -1,231 +1,303 @@
...
+          await commands.create.createWallet(cmdParams);
+          break;
+        case 'join':
+          await commands.join.joinWallet(cmdParams);
+          break;
+        case 'import-seed':
+          const mnemonic = await prompt.password({
+            message: 'Enter your 12-24 word mnemonic phrase:',
+            mask: '*',
+            validate: (input) => input.split(' ').length >= 12 && input.split(' ').length <= 24 ? undefined : 'Mnemonic must be between 12 and 24 words.',
```
MichaelAJay: All values in [12, 24] are valid?

packages/bitcore-cli/src/commands/txhistory.ts
```diff
@@ -1,99 +1,158 @@
...
   await Utils.paginate(async (page, viewAction) => {
-    if (viewAction === 'f') {
+    if (viewAction === ViewAction.TOGGLE_FORMAT) {
       compact = !compact; // toggle compact view
       printRaw = false; // reset printRaw when toggling format
     }
 
-    printRaw = viewAction === 'r' ? !printRaw : printRaw;
-    const exportToFile = viewAction === 'e';
+    printRaw = viewAction === ViewAction.TOGGLE_RAW ? !printRaw : printRaw;
```
MichaelAJay, line 70: You've got `conditional ? false : true`.

`!conditional ? true : false` and then you can see that you can just drop the ternary altogether

packages/bitcore-cli/src/commands/txproposals.ts
```diff
@@ -1,32 +1,59 @@
...
 
-  const txps = await wallet.client.getTxProposals({
-    forAirGapped: false, // TODO
-    // limit: pageSize,
-    // skip: (page - 1) * pageSize
-  });
+  const txps = opts.command && opts.proposalId
+    ? [await wallet.client.getTx(opts.proposalId)]
+    : await wallet.client.getTxProposals({
+      forAirGapped: false, // TODO
```
MichaelAJay, line 51: this still on your radar?

packages/bitcore-wallet-client/src/lib/tsskey.ts
```diff
@@ -207,16 +207,21 @@ export class TssKeyGen extends EventEmitter {
...
     this.#requestPrivateKey = BitcoreLib.PrivateKey.fromString(this.#credentials.requestPrivKey);
-    this.backupKeyShare = !!params.backupKeyShare || true;
+
+    const baseXPrivKey = this.#key.get(params.password).xPrivKey;
+    const derivationPath = this.#credentials.getRootPath();
+    this.#xPrivKey = new BitcoreLib.HDPrivateKey(baseXPrivKey).deriveChild(derivationPath).toString();
+    const _seed = BitcoreLib.HDPrivateKey.fromString(this.#xPrivKey);
+    this.#seed = BitcoreLib.crypto.Hash.sha256(_seed.toBuffer());
+
+    this.backupKeyShare = !!params.backupKeyShare || false;
```
MichaelAJay: this.backupKeyShare assignment was previously extraneous with its default case, and is now too (although differently):

Let A = !!params.backupKeyShare
A || true = true
A || false = A

packages/bitcore-wallet-client/test/api.test.ts
```diff
@@ -2435,6 +2451,37 @@ describe('client API', function() {
...
+              message: 'hello 1-1'
+            };
+            helpers.createAndPublishTxProposal(clients[0], opts, async (err, x) => {
+              done(err);
+            });
+          });
+        });
+      });
+    });
+    it('Should return only main addresses', async function() {
```
MichaelAJay: These tests would be more useful if you included in the descriptions why they only return main addresses/all addresses.

`Should return only main addresses when noChange is true`

packages/bitcore-wallet-client/test/helpers.ts
```diff
@@ -192,17 +194,22 @@ export const helpers = {
...
+    try {
+      if (!opts.outputs) {
+        opts.outputs = [{
+          toAddress: opts.toAddress,
+          amount: opts.amount
+        }];
+      }
+      let txp = await client.createTxProposal(opts);
+      txp = await client.publishTxProposal({ txp });
+      if (cb) { cb(null, txp); }
```
MichaelAJay, line 207: Is this double return behavior what you want? Should you use `else` to return txp?

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1669,28 +1669,42 @@ export class WalletService implements IWalletService {
...
    * Get all addresses.
-   * @param {Object} opts
-   * @param {Numeric} opts.limit (optional) - Limit the resultset. Return all addresses by default.
-   * @param {Numeric} opts.skip (optional) - Skip this number of addresses in resultset. Useful for paging.
-   * @param {Boolean} [opts.reverse=false] (optional) - Reverse the order of returned addresses.
    * @returns {Address[]}
    */
-  getMainAddresses(opts, cb) {
+
+  getAddresses(opts: {
```
MichaelAJay: This would make a good Type (you define it inline in expressapp.ts).

packages/crypto-wallet-core/src/utils/bigint.ts
MichaelAJay, line 1: Consider defining a union type for the common `BigInt|Number|String` and remove jsdocs - favor TS type annotations.

packages/crypto-wallet-core/src/utils/bigint.ts
```diff
@@ -0,0 +1,142 @@
...
+};
+
+
+/**
+ * Divide BigInts with precision. This works for Numbers and strings as well since it converts the inputs to BigInts
+ * @param {BigInt|Number|String} numerator
+ * @param {BigInt|Number|String} denominator
+ * @returns {BigInt} Rounded BigInt
+ */
+export function div(numerator, denominator) {
```
MichaelAJay: returns a `number`, not a BigInt

packages/crypto-wallet-core/src/utils/bigint.ts
```diff
@@ -0,0 +1,142 @@
...
+
+
+/**
+ * Divide BigInts with precision. This works for Numbers and strings as well since it converts the inputs to BigInts
+ * @param {BigInt|Number|String} numerator
+ * @param {BigInt|Number|String} denominator
+ * @param {BigInt|Number|String} precision (optional) Defaults to the max number of decimals in the inputs
+ * @returns {Number}
+ */
+export function divToFloat(numerator, denominator, precision) {
```
MichaelAJay: Check for division by 0?

packages/crypto-wallet-core/src/utils/bigint.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (precision == null) {
+    precision = Math.max(numerator.toString().split('.')[1]?.length || 0, denominator.toString().split('.')[1]?.length || 0);
+  }
+  const scaleFactor = 10n ** BigInt(precision);
+  const scaledNumerator = BigInt(numerator) * scaleFactor;
```
MichaelAJay: You may consider ensuring numerator is an integer. BigInt(<not integer>) throws a RangeError, and that's not handled.

packages/bitcore-cli/src/commands/txhistory.ts
kajoseph: Not quite. It's negating `printRaw`, so it could be `conditional ? false : true;` or `conditional ? true : false` depending on the current value of `printRaw`

packages/bitcore-cli/src/commands/txproposals.ts
kajoseph: Not for the TSS project. Air gapping was an (incomplete) project from before my time. It's an interesting project and one that I'd like to come back to at some point, but not now.

packages/bitcore-wallet-client/test/helpers.ts
kajoseph: Yeah, I was going for making the promise return a first class citizen instead of prioritizing the cb over it. This way, it's just a one line delete change when we do remove cb support.

packages/bitcore-cli/src/cli.ts
kajoseph: yeah, this is a pretty naive check. Updated

packages/crypto-wallet-core/src/utils/bigint.ts
```diff
@@ -1,72 +1,96 @@
...
- * @param {BigInt|Number|String} precision (optional) Defaults to the max number of decimals in the inputs
- * @returns {Number}
+ * @param {BigIntLike} numerator
+ * @param {BigIntLike} denominator
+ * @param {BigIntLike} precision (optional) Default: 18
+ * @returns {number}
  */
-export function divToFloat(numerator, denominator, precision) {
+export function divToFloat(numerator: BigIntLike, denominator: BigIntLike, precision?: number) {
+  if (!denominator) throw new Error('Division by zero');
```
MichaelAJay: This check lets '0' through and shouldn't.

Some more below (in div())

Propose (isValidDenominator()) util.

packages/crypto-wallet-core/test/bigint.test.ts
MichaelAJay, line 1: Add div and divToFloat tests for denominator = '0' (string 0)

MichaelAJay (approved):

nitsujlangston (approved):

PR #3984, opened 2025-08-26
[BWC] Fix Solana Account Index Signing Bug

leolambo:


packages/bitcore-wallet-client/test/api.test.js
```diff
@@ -1408,6 +1439,36 @@ describe('client API', function() {
           );
         });
       });
+
+      it.only('should not sign SOL proposal with invalid path', () => {
```
kajoseph: .only

PR #3986, opened 2025-08-27
[BCN] Add fee data to blocks

MicahMaphet:


packages/bitcore-node/scripts/addBlockFees.js
```diff
@@ -0,0 +1,62 @@
...
+    if (remove) {
+      console.log(`Removing fee data from all blocks on ${chain} ${network}`);
+      await BitcoinBlockStorage.collection.updateMany({ chain, network }, { $unset: { feeData: {} } })
+      return;
+    }
+    console.log(`Adding fee data to all blocks on ${chain} ${network}`);
+    let block = await BitcoinBlockStorage.collection.findOne({ chain: chain, network: network, height: 0 });
+
+    while (true) {
+      if (!block.feeData?.feeTotal) {
```
kajoseph: Consider the case when feeTotal is 0

packages/bitcore-node/scripts/addBlockFees.js
```diff
@@ -0,0 +1,62 @@
...
+  .then(async () => {
+    if (remove) {
+      console.log(`Removing fee data from all blocks on ${chain} ${network}`);
+      await BitcoinBlockStorage.collection.updateMany({ chain, network }, { $unset: { feeData: {} } })
+      return;
+    }
+    console.log(`Adding fee data to all blocks on ${chain} ${network}`);
+    let block = await BitcoinBlockStorage.collection.findOne({ chain: chain, network: network, height: 0 });
+
+    while (true) {
```
kajoseph: A more conventional way would be to do `while (block.nextBlockHash)`

packages/bitcore-node/scripts/addBlockFees.js
```diff
@@ -0,0 +1,62 @@
...
+
+const chain = args[args.indexOf('--chain') + 1];
+const network = args[args.indexOf('--network') + 1];
+
+if (!['BTC', 'BCH', 'DOGE', 'LTC'].includes(chain) || !['mainnet', 'testnet', 'regtest'].includes(network)) {
+  usage('Invalid chain and/or network param(s).');
+}
+
+const remove = args.includes("--remove")
+const verbose = args.includes("--verbose")
```
kajoseph: We need to fix our linting, but our convention in this codebase is to use single quotes

packages/bitcore-node/scripts/addBlockFees.js
MicahMaphet: I wasn't able to find a good way to do that without skipping the last block. I could do while (block).

kajoseph: Ah, you could also use a `do..while` loop

MicahMaphet: I looked at it more carefully and completely redid the query. Adding feeData to 88,000 blocks (with maybe 50 non coinbase transactions) took 180 seconds with looping through block.nextBlockHash but now takes 30 seconds.

kajoseph: While I appreciate the rework and the performance boot, I think streaming is a better way to go. Keep in mind this is going to be run on a production server that's serving API requests. Doing .toArray() on the entire set of blocks from 0 - X is not memory efficient (almost 1M blocks for BTC...higher for other chains), and then doing a Promise.all on all 1M is going to peg the CPU, plus you're doing 2 db calls per block, so it'll likely peg the db as well.

When doing migration scripts, runtime is less important that resource management. 180s of runtime is more than sufficient. I've had migration scripts run for days.

It's also a good idea to put cooldowns in loop iterations so you aren't pegging the db with calls. Usually tens of ms is good...like 50ms or so.

MicahMaphet: The stream is way better. To my surprise the stream is faster even while taking up less time in user space.
With 145,000 promises: 50.77user 3.91system 0:58.69elapsed 93%CPU (48 seconds starting after db connection)
With a stream: 7.78user 3.35system 0:19.46elapsed 57%CPU (9 seconds starting after db connection)

MicahMaphet:
@kajoseph yes

PR #3992, opened 2025-09-08
Adds block/before-time functionality for SOL

SethMcGuire:


packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -415,6 +415,17 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
     return this.blockTransform(network, block, slot);
   }
 
+  async getBlockBeforeTime(params: GetBlockBeforeTimeParams): Promise<IBlock | null> {
+    const { network, time = new Date() } = params;
+    const { rpc } = await this.getRpc(network);
+    const beforeTimeSlot = await this._findSlotByDate(network, new Date(time));
+    if (beforeTimeSlot === null) {
+      throw new Error('Unable to find block before specified time');
```
kajoseph: I believe this will bubble up to the route and return a 500 error. Probably better to just return null

PR #3995, opened 2025-09-10
[BWS] Fix SOL token balance

kajoseph:
This PR:

* Adds the tokenAddress opt to the SOL getBalance call
* Adds better typing to the V8 class method params


leolambo:
is production BWS on v11 ?

leolambo (approved):
LGTM

SethMcGuire:
LGTM

PR #3997, opened 2025-09-11
[BCN] Add SPL Token program data

leolambo:
Adds program data to token info endpoint `/api/SOL/:network/token/:tokenAddress`
 e.g 
`{"name":"USDT","symbol":"USDT","decimals":6,"programType":"token","programAddress":"TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"}`

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
+++ b/packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
@@ -11,7 +11,7 @@ import { CacheStorage } from '../../../../models/cache';
 import { IBlock } from '../../../../types/Block';
 import { CoinListingJSON } from '../../../../types/Coin';
 import { IChainConfig, IProvider, ISVMNetworkConfig } from '../../../../types/Config';
-import { BroadcastTransactionParams, GetBalanceForAddressParams, GetBlockParams, GetCoinsForTxParams, GetEstimatePriorityFeeParams, GetWalletBalanceParams, IChainStateService, StreamAddressUtxosParams, StreamBlocksParams, StreamTransactionParams, StreamTransactionsParams, StreamWalletTransactionsParams, WalletBalanceType, GetBlockBeforeTimeParams } from '../../../../types/namespaces/ChainStateProvider';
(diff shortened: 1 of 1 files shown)


packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -744,7 +749,29 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
...
+      const result = await connection.getAccountInfo(tokenAddress).send();
+      const owner = result?.value?.owner?.toString?.();
+      if (!owner) {
+        throw new Error(`Mint account not found or unreadable: ${tokenAddress}`);
+      }
+      if (owner === TOKEN_PROGRAM_ADDRESS) {
+        programType = 'token';
+        programAddress = TOKEN_PROGRAM_ADDRESS;
+      }
+      if (owner === TOKEN_2022_ADDR) {
```
MichaelAJay: owner can't be both - use if / else if.

packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -744,7 +749,29 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
...
+        programType =  'token2022';
+        programAddress = TOKEN_2022_ADDR;
+      }
+      if (!programAddress) {
+        throw new Error(`Unknown token program owner for ${tokenAddress}: ${owner}`);
```
MichaelAJay: Consider logging directly instead of throwing an error here - this isn't exceptional behavior (it just means we didn't cover every type that owner could be - which of course you did.

packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -698,11 +698,16 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
  async getSPLTokenInfo(
     network: string, 
     tokenAddress: string
-  ): Promise<{ name: string; symbol: string; decimals: number }> {
-    const { umi } = await this.getRpc(network);
+  ): Promise<{ name: string; symbol: string; decimals: number; programType: string; programAddress: string; }> {
```
MichaelAJay: The way this method executes, it's possible that programType & programAddress could be undefined. In practice, it shouldn't happen since the token ought to belong to one of those two owners. But it is technically possible

PR #4004, opened 2025-09-26
[BCN] Adds balanceAtTime to XRP, EVM, and SOL

SethMcGuire:


packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -383,10 +383,85 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
...
+      const txList = await connection.getSignaturesForAddress(ata || address, { limit: 1, before: sigInBlock }).send();
+      if (!txList.length || !txList[0]?.signature) {
+        throw new Error(`Balance not found at ${args.time}`);
+      }
+      const tx = await rpc.getTransaction({ txid: txList[0].signature });
+      if (!tx) {
+        throw new Error(`Balance not found at ${args.time}`);
+      }
+      const index = tx.accountKeys?.findIndex(acct => acct == address);
+      if (!index || index === -1) {
```
leolambo: would throw error on 0. findIndex returns 0 for the first element.
should be:
  `if (index === undefined || index === -1) {`

packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -383,10 +383,85 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
...
+      }
+      const txList = await connection.getSignaturesForAddress(ata || address, { limit: 1, before: sigInBlock }).send();
+      if (!txList.length || !txList[0]?.signature) {
+        throw new Error(`Balance not found at ${args.time}`);
+      }
+      const tx = await rpc.getTransaction({ txid: txList[0].signature });
+      if (!tx) {
+        throw new Error(`Balance not found at ${args.time}`);
+      }
+      const index = tx.accountKeys?.findIndex(acct => acct == address);
```
leolambo: use `===`

packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -383,10 +383,85 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
...
+        confirmed: BigInt(prev.confirmed) + BigInt(cur.confirmed),
+        balance: BigInt(prev.balance) + BigInt(cur.balance)
+      }),
+      { unconfirmed: 0n, confirmed: 0n, balance: 0n }
+    );
+    return {
+      unconfirmed: Number(balance.unconfirmed),
+      confirmed: Number(balance.confirmed),
+      balance: Number(balance.balance)
+    };;
```
leolambo: remove extra `;`

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
```diff
@@ -216,12 +217,53 @@ export class BaseEVMStateProvider extends InternalStateProvider implements IChai
...
+    };
+  }
+
   async getBalanceForAddress(params: GetBalanceForAddressParams): Promise<WalletBalanceType> {
     const { chain, network, address, args } = params;
     const { web3 } = await this.getWeb3(network, { type: 'realtime' });
     const tokenAddress = args?.tokenAddress;
     const addressLower = address.toLowerCase();
     const hex = args?.hex === 'true' || args?.hex === '1';
+    let blockNumber: number;
```
leolambo: type should include `undefined` or handle undefined case

leolambo (approved):
Nice work

packages/bitcore-node/src/modules/ripple/api/csp.ts
```diff
@@ -91,16 +92,41 @@ export class RippleStateProvider extends InternalStateProvider implements IChain
...
     const lowerAddress = address.toLowerCase();
     const cacheKey = `getBalanceForAddress-${chain}-${network}-${lowerAddress}`;
     return CacheStorage.getGlobalOrRefresh(
       cacheKey,
       async () => {
         const client = await this.getClient(network);
         try {
-          const balance = await client.getBalance({ address });
+          let ledgerIndex: number | undefined;
+          if (args?.time) {
```
kajoseph, line 122: This gets cached under the same non-time key. Need to check if this is also true for the other chains affected in this PR (EVM, SVM)

kajoseph: Tesolving since the cache time is set to None/0 if args.time, which clears the 1 minute cache for the non-time balance. Not a big deal... At worst, we might get a few more RPC calls.

kajoseph (approved):

PR #4014, opened 2025-10-07
[BWS] Ref: Transak getSignedPaymentUrl

Gamboster:


packages/bitcore-wallet-service/src/externalservices/transak.ts
```diff
@@ -173,82 +193,92 @@ export class TransakService {
...
+      const WIDGET_API = keys.WIDGET_API;
+
+      if (
+        !checkRequired(req.body, requiredParams)
+      ) {
+        return reject(new ClientError("Transak's request missing arguments"));
+      }
+
+      let accessToken;
+        if (req.body.accessToken) {
```
kajoseph: lint: indentation

packages/bitcore-wallet-service/src/externalservices/transak.ts
```diff
@@ -173,82 +193,92 @@ export class TransakService {
...
+      try {
+        keys = this.transakGetKeys(req);
+      } catch (err) {
+        return reject(err);
+      }
+      const API_KEY = keys.API_KEY;
+      const WIDGET_API = keys.WIDGET_API;
+
+      if (
+        !checkRequired(req.body, requiredParams)
```
kajoseph: lint: no need for a new line here

packages/bitcore-wallet-service/src/externalservices/transak.ts
```diff
@@ -173,82 +193,92 @@ export class TransakService {
...
+        'network',
+        'cryptoCurrencyCode',
+        'partnerOrderId',
+        'partnerCustomerId',
+      ];
 
-    const headers = {
-      'Content-Type': 'application/json'
-    };
+      const env = _.cloneDeep(req.body.env);
```
kajoseph: We're trying to get rid of lodash. Would prefer to use something like `JSON.parse(JSON.stringify(req.body.env));`

However, I don't think this is necessary anyway. See other comment about where `env` is used

packages/bitcore-wallet-service/src/externalservices/transak.ts
```diff
@@ -173,82 +193,92 @@ export class TransakService {
...
+      ) {
+        return reject(new ClientError("Transak's request missing arguments"));
+      }
+
+      let accessToken;
+        if (req.body.accessToken) {
+          accessToken = req.body.accessToken;
+        } else {
+          try {
+            const accessTokenData = await this.transakGetAccessToken({ body: env });
```
kajoseph: I think this usage of `env` is incorrect. Deeper in the code, it looks like `req.body.env` is supposed to be a string (`production` or `sandbox`). So, this should be `transakGetAccessToken({ body: { env } });`.

However, it looks like there's also the need for the `req.body.context` string as well. Can we not just pass in `req`? It looks like we don't need to do the _.deepClone() above and can just call `transakGetAccessToken(req)` since req.body gets reassigned below anyway.

> This also applies to `transakGetOrderDetails`.

I suspect that the reason for the _.deepClone is because of this in `transakGetKeys`:
```
    delete req.body.env;
    delete req.body.context;
```
in which case it should probably be `const body = _.deepClone(req.body);`. However, I think we can handle this a bit better/cleaner. Do we even need to delete the env and context? If so, maybe we can add a param to `transakGetKeys` like `transakGetKeys(req, cleanBody?: boolean = true) {` OR we can just move this logic to be above the `transakGetKeys` call.

Gamboster: Thanks for digging deeper into this. While it may not be strictly necessary to remove `env` and `context`, since these are only parameters used by us, I prefer to avoid sending them in requests to our providers; they could cause errors or misinterpretations.
Thanks for detecting the error in `this.transakGetAccessToken({ body: env })`. I fixed it by sending "req".
I removed lodash and opted to add the `cleanBody` parameter as you suggested.

packages/bitcore-wallet-service/src/externalservices/transak.ts
Gamboster: I removed `_.cloneDeep`

packages/bitcore-wallet-service/src/externalservices/transak.ts
Gamboster: fixed

packages/bitcore-wallet-service/src/externalservices/transak.ts
Gamboster: fixed ty!

PR #4018, opened 2025-10-09
[BCN] Fix Class type errors

kajoseph:
This PR:

* Fixes typing errors thrown by ts-node when you try to run bitcore-node with ts-node (e.g. running tests)


kajoseph:
using tsx per https://github.com/bitpay/bitcore/pull/4027

PR #4021, opened 2025-10-10
[BCC] Lint bitcore-client

kajoseph:
This PR:

* Lints bitcore-client per the new eslint rules

### DIFF
--- a/packages/bitcore-client/src/storage.ts
+++ b/packages/bitcore-client/src/storage.ts
@@ -163,8 +163,7 @@ export class Storage {
     let open = true;
     for (const key of keys) {
       const { path } = key;
-      let { pubKey } = key;
-      pubKey = pubKey || new BitcoreLib.PrivateKey(key.privKey).publicKey.toString();
+      const pubKey = key.pubKey || new BitcoreLib.PrivateKey(key.privKey).publicKey.toString();
       let payload = {};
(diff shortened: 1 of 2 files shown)


packages/bitcore-client/src/storage.ts
```diff
@@ -163,7 +162,8 @@ export class Storage {
     const { name, keys, encryptionKey } = params;
     let open = true;
     for (const key of keys) {
-      let { pubKey, path } = key;
+      const { path } = key;
+      let { pubKey } = key;
```
MichaelAJay: const pubKey = key.pubKey || new BitcoreLib... would allow const delcaration.

packages/bitcore-client/src/wallet.ts
```diff
@@ -629,7 +629,8 @@ export class Wallet {
   }
 
   async signTx(params) {
-    let { tx, keys, utxos, passphrase, signingKeys, changeAddressIdx } = params;
+    const { tx, keys, passphrase, signingKeys, changeAddressIdx } = params;
+    let { utxos } = params;
```
MichaelAJay: This is fine - but we could easily prefer const with just a small update:
```
    const utxos = params.utxos || [];
    if (!utxos.length) {...};
```

Would help downstream understandability

PR #4023, opened 2025-10-13
[BCN] Fixes SPL balance at time bugs

SethMcGuire:


packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -446,15 +446,15 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
       if (!tx) {
         throw new Error(`Balance not found at ${args.time}`);
       }
-      const index = tx.accountKeys?.findIndex(acct => acct === address);
+      const index = tx.accountKeys?.findIndex(acct => acct === (ata || address));
```
kajoseph, line 449: I have not been seeing `accountKeys` on any txs. Is there a more reliable way of finding this?

SethMcGuire: You may need to npm i. Michael added accountKeys to the crypto-rpc `getTransaction` method a couple weeks ago.

PR #4028, opened 2025-10-17
[BCN] Try earlier blocks if SOL block is missing in blockBeforeTime

SethMcGuire:


packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -494,11 +494,22 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
...
+    let beforeTimeSlot = await this._findSlotByDate(network, new Date(time));
     if (!beforeTimeSlot) {
       return null;
     }
-    const block = await rpc.getBlock({ height: beforeTimeSlot });
+    let block;
+    for (let attempt = 0; attempt < 10; attempt++) { // retry in case of skipped or missing slots
+      try {
+        block = await rpc.getBlock({ height: beforeTimeSlot });
+      } catch (e: any) {
```
kajoseph, line 506: What if a block is found? For loop continues...

PR #4029, opened 2025-10-17
Eslint plugin import

MichaelAJay:
Adds and configures eslint-plugin-import.
Maintains import-sort rules which are auto-fixable.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -60,9 +60,10 @@ export default defineConfig([
       }],
       'import/order': ['error', {
         groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index', 'object', 'type'],
-        'newlines-between': 'always',
+        'newlines-between': 'never',
         alphabetize: { order: 'asc', caseInsensitive: false }
-      }]
+      }],
+      'import/newline-after-import': 'error'
     }
   },
   {


eslint.config.mjs
```diff
@@ -53,11 +55,13 @@ export default defineConfig([
...
-        ignoreDeclarationSort: false,
-        ignoreMemberSort: false,
-        memberSyntaxSortOrder: ['none', 'all', 'multiple', 'single'],
-        allowSeparatedGroups: false
+        ignoreDeclarationSort: true, // rule cannot be auto-fixed
+        ignoreMemberSort: false // import { b, a } from 'X' -> import { a, b } from 'X'
+      }],
+      'import/order': ['error', {
+        groups: ['builtin', 'external', 'internal', 'parent', 'sibling', 'index', 'object', 'type'],
+        'newlines-between': 'always',
```
kajoseph: Is this doing what you want? e.g. it turns the imports in `bitcore-node/src/config.ts` from
```typescript
import { cpus, homedir } from 'os';
import { ConfigType } from './types/Config';
import { merge } from './utils';
import parseArgv from './utils/parseArgv';
```

to

```typescript
import { cpus, homedir } from 'os';

import { ConfigType } from './types/Config';
import { merge } from './utils';
import parseArgv from './utils/parseArgv';
```

Perhaps you meant the `import/newline-after-import` rule?

kajoseph: I would expect `newlines-between` to be `never`

MichaelAJay: I can change newlines-between (groups) to be never. I kind of like the separation of groups. But if you don't - you're the one that looks at it the most.

kajoseph (approved):

PR #4031, opened 2025-10-20
Lint previously linted for imports

MichaelAJay:
From recently-completed packages which were fully linted, re-lint with new import linting rules.


packages/crypto-wallet-core/package.json
```diff
@@ -10,7 +10,8 @@
     "compile": "npm run clean && npm run build",
     "precommit": "echo 'TODO: npm run fix'",
```
kajoseph: shouldn't this be non-TODO'd? I thought you did it before when you linted CWC...

kajoseph (approved):
LGTM

PR #4039, opened 2025-10-28
Bitcore vault

MichaelAJay:


packages/bitcore-vault/src/securityManager.ts
```diff
@@ -0,0 +1,111 @@
...
+      if (SecurityManager.inspectorUrlExists()) {
+        return { result: false, reason: 'Inspector url detected' };
+      }
+      return { result: true };
+    } catch (err) {
+      return { result: false, reason: err.message };
+    }
+  }
+
+  /**
```
MicahMaphet: This indentation doesn't look right

packages/bitcore-vault/src/securityManager.ts
```diff
@@ -0,0 +1,111 @@
...
+import inspector from 'node:inspector';
+
+export class SecurityManager {
+  private secureHeapBaseAllocation: number;
+    
+  constructor() {}
+
+  public async runAllSecurityChecks(): Promise<{ result: boolean; reason?: string }> {
+    const quickChecks = this.runQuickSecurityCheck();
+    if (!quickChecks.reason) {
```
leolambo: this conditional is backwards. remove `!`

packages/bitcore-vault/src/securityManager.ts
```diff
@@ -0,0 +1,111 @@
...
+  public static checkInspectFlagsAtLaunch(): boolean {
+    try {
+      const argv = (process.execArgv || []).join(' ');
+      const nodeOpts = String(process.env.NODE_OPTIONS || '');
+      const rx = /--inspect(?:-brk)?\b|--inspect-port\b/;
+      const isDetected = rx.test(argv) || rx.test(nodeOpts);
+      if (isDetected) {
+        console.warn('[SecurityManager] Inspect flags detected on secure process');
+      }
+      return;
```
leolambo: return boolean

packages/bitcore-vault/src/vaultWalletProxy.ts
```diff
@@ -0,0 +1,364 @@
...
+    return this.sendMessage<{ success: boolean }>('removeWallet', { name: walletName }, { timeoutMs });
+  }
+
+  private async promptForPassphrase(opts: { prompt?: string; maxBytes?: number } = {}): Promise<Buffer> {
+    if (!this.publicKey) throw new Error('Vault not initialized (missing public key)');
+
+    const promptText = opts.prompt ?? 'Passphrase: ';
+    const maxBytes = Math.min(Math.max(opts.maxBytes ?? 256, 8), 4096);
+    const stdin = process.stdin as NodeJS.ReadStream & { setRawMode?: (mode: boolean) => void };
+    const plain = Buffer.allocUnsafe(maxBytes);
```
leolambo: why are.you using unsafe? consider plain `alloc()`

packages/bitcore-vault/src/vaultWallet.ts
```diff
@@ -0,0 +1,104 @@
...
+      crypto.randomFillSync(passphrase); // Overwrite passphrase memory
+      this.lock();
+    }
+  }
+
+  /**
+   * Override signTx to require vault access
+   */
+  public async signTx(params: any): Promise<any> {
+    return this.withVaultAccess(params.passphrase, super.signTx.bind(this), params);
```
leolambo: this one is interesting `withVaultAccess` unlocks using the passphrase. but the passphrase still needs to be sent via params?

packages/bitcore-vault/src/vaultWalletProxy.ts
```diff
@@ -0,0 +1,364 @@
...
+            : `exited with code ${code}`;
+
+          const msg = `SecureProcess ${exitReason}. Manual reinitialization required.`;
+          console.warn(msg);
+
+          const error = new Error(msg);
+          this.rejectAllPendingMessages(error);
+          this.secureProcess = null;
+          this.initializationPromise = null;
+
```
leolambo, line 99: should clear this.publicKey, this.walletAddresses

```
this.publicKey = null;
this.walletAddresses.clear();
```

packages/bitcore-vault/src/secureProcess.ts
```diff
@@ -0,0 +1,451 @@
...
+      this.securityQuickCheckInterval = null;
+    }
+  }
+
+  private checkSecureHeap(): boolean {
+    return SecurityManager.isSecureHeapEnabled();
+  }
+
+  private getPublicKey(): string {
+    return this.publicKey.export({ type: 'spki', format: 'pem' }).toString();
```
leolambo, line 249: add `this.publicKey` null check

packages/bitcore-vault/src/secureProcess.ts
```diff
@@ -0,0 +1,451 @@
...
+  }
+
+  private async addPassphrase(payload: { name: string; encryptedPassphrase: string }): Promise<{ success: boolean }> {
+    const { name, encryptedPassphrase } = payload;
+    const walletEntry = this.wallets.get(name);
+
+    if (!walletEntry) {
+      throw new Error(`Wallet not found: ${name}`);
+    }
+
```
leolambo, line 277: consider privateKey null check

```
    if (!this.privateKey) {
      throw new Error('Private key not available - secure process not initialized or already cleaned up');
    }
```

packages/bitcore-vault/src/vaultWalletProxy.ts
```diff
@@ -0,0 +1,364 @@
...
+              console.error('Secure process exited unexpectedly. Parent exiting as configured.');
+              process.exit(1);
+            } else {
+              console.error('Secure process exited unexpectedly. Parent continues as configured.');
+            }
+          }
+        });
+
+        await this.sendMessage<void>('initialize', {});
+        const publicKeyPem = await this.sendMessage<string>('getPublicKey', {});
```
leolambo, line 112: consider validating retrieved key

```
if (!publicKeyPem || typeof publicKeyPem !== 'string' || publicKeyPem.trim().length === 0) {
    throw new Error('Invalid public key received from secure process');
}
```

packages/bitcore-vault/src/vaultWallet.ts
MichaelAJay: Good catch. Updated.

packages/bitcore-vault/src/vaultWalletProxy.ts
MichaelAJay: Good call. `typeof publicKeyPem !== 'string' || !publicKey.trim().length` is functionally similar - any other falsy value (including undefined) than `""` wouldn't make it past the type check, and the trim().length takes care of `""`.

packages/bitcore-vault/src/vaultWalletProxy.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    // Instruct secure process to remove the wallet
+    return this.sendMessage<{ success: boolean }>('removeWallet', { name: walletName }, { timeoutMs });
+  }
+
+  private async promptForPassphrase(opts: { prompt?: string; maxBytes?: number } = {}): Promise<Buffer> {
+    if (!this.publicKey) throw new Error('Vault not initialized (missing public key)');
+
+    const promptText = opts.prompt ?? 'Passphrase: ';
+    const maxBytes = Math.min(Math.max(opts.maxBytes ?? 256, 8), 4096);
```
leolambo: RSA-2048 with OAEP padding and SHA-256 can encrypt a maximum of 190 bytes of data in a single operation. publicEncrypt on line 340 will throw for passphrases > 190 bytes. consider using 190 instead of 4096

packages/bitcore-vault/src/securityManager.ts
```diff
@@ -0,0 +1,108 @@
...
+      return { result: false, reason: err.message };
+    }
+  }
+
+  public static isSecureHeapEnabled(): boolean {
+    const { total } = secureHeapUsed();
+    return total > 0;
+  }
+    
+  public secureHeapAllocationCheck(): boolean {
```
leolambo: seems like dead code? also the base alloc >= current alloc is contradicting `currentAllocation >= this.secureHeapBaseAllocation` found on ln 70

packages/bitcore-cli/src/commands/txhistory.ts
```diff
@@ -125,6 +125,7 @@ export async function getTxHistory(
         const timestamp = new Date(tx.time * 1000);
         const time = compact ? Utils.formatDateCompact(timestamp) : Utils.formatDate(timestamp);
         const txid = compact ? Utils.compactString(tx.txid) : tx.txid;
+        tokenObj = tokenObj && !compact ? { ...tokenObj, precision: tokenObj.decimals.short.maxDecimals } : tokenObj;
```
kajoseph: Out of scope? Not sure what this is from. Seems like an artifact of the unsigned commit stuff

PR #4040, opened 2025-10-30
[BWS] Lint bitcore-wallet-service

MicahMaphet:
Lints bitcore-wallet-service with new eslint rules

### DIFF
--- a/packages/bitcore-wallet-service/scripts/migrateWalletsNetwork.js
+++ b/packages/bitcore-wallet-service/scripts/migrateWalletsNetwork.js
@@ -160,7 +160,7 @@ storage.connect(config.storageOpts, async (err) => {
         const resCache = await storage.db.collection(Storage.collections.CACHE).deleteMany({
           walletId: wallet.id
         });
-        if (resCache?.result?.n <= 0 && !resCache?.result?.ok) {
+        if (!resCache?.result?.ok) {
           console.log(JSON.stringify(resTxs));
         }
       } else {
--- a/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
+++ b/packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
@@ -96,12 +96,12 @@ export class EthChain implements IChain {
         return cb(err);
       }
       // getPendingTxs returns all txps when given a native currency
-      server.getPendingTxs(opts, (err, _txps) => {
+      server.getPendingTxs(opts, (err, txps) => {
         if (err) return cb(err);
         let fees = 0;
         let amounts = 0;
 
-        _txps.filter(txp => {
+        txps.filter(txp => {
           // Add gas used for tokens when getting native balance
           if (!opts.tokenAddress) {
             fees += txp.fee || 0;
--- a/packages/bitcore-wallet-service/src/lib/model/email.ts
+++ b/packages/bitcore-wallet-service/src/lib/model/email.ts
@@ -1,6 +1,25 @@
 import _ from 'lodash';
 import Uuid from 'uuid';
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+interface IEmail {
+  version: number;
+  createdOn: number;
+  id: number;
+  walletId: string;
+  copayerId: string;
+  from: string;
+  to: string;
+  subject: string;
+  bodyPlain: string;
+  bodyHtml: string;
+  status: string;
+  attempts: number;
+  lastAttemptOn?: number;
+  notificationId: string;
+  language: string;
+}
+
 export class Email {
   version: number;
   createdOn: number;
--- a/packages/bitcore-wallet-service/src/lib/model/wallet.ts
+++ b/packages/bitcore-wallet-service/src/lib/model/wallet.ts
@@ -337,7 +337,7 @@ export class Wallet implements IWallet<() => boolean> {
     );
   }
 
-  // / Only for power scan
+  // Only for power scan
   getSkippedAddress() {
     $.checkState(this.isComplete(), 'Failed state: this.isComplete() at <getSkipeedAddress()>');
 
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -5185,18 +5185,17 @@ export class WalletService implements IWalletService {
 
   moralisGetTokenAllowance(req): Promise<any> {
     return new Promise((resolve, reject) => {
+      if (!config.moralis) return reject(new Error('Moralis missing credentials'));
+      if (!checkRequired(req.body, ['address']) && !checkRequired(req.body, ['ownerAddress'])) {
+        return reject(new ClientError('moralisGetTokenAllowance request missing arguments'));
+      }
+      
+      const walletAddress = req.body.ownerAddress ?? req.body.address;
       const headers = {
         'Accept': 'application/json',
         'Content-Type': 'application/json',
         'X-Api-Key': config.moralis.apiKey,
       };
-
-      if (!config.moralis) return reject(new Error('Moralis missing credentials'));
-      if (!checkRequired(req.body, ['address']) && !checkRequired(req.body, ['ownerAddress'])) {
-        return reject(new ClientError('moralisGetTokenAllowance request missing arguments'));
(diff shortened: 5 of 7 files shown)


packages/bitcore-wallet-service/scripts/migrateWalletsNetwork.js
```diff
@@ -159,9 +160,7 @@ storage.connect(config.storageOpts, async (err) => {
         const resCache = await storage.db.collection(Storage.collections.CACHE).deleteMany({
           walletId: wallet.id
         });
-        if (resCache?.result?.n > 0) {
-          fixCacheCount++;
-        } else if (!resCache?.result?.ok) {
+        if (resCache?.result?.n <= 0 && !resCache?.result?.ok) {
```
kajoseph: We should get rid of the `resCache?.result?.n <= 0` condition here.

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
```diff
@@ -98,12 +96,12 @@ export class EthChain implements IChain {
...
       }
       // getPendingTxs returns all txps when given a native currency
-      server.getPendingTxs(opts, (err, txps) => {
+      server.getPendingTxs(opts, (err, _txps) => {
         if (err) return cb(err);
         let fees = 0;
         let amounts = 0;
 
-        txps = txps.filter(txp => {
+        _txps.filter(txp => {
```
kajoseph: this doesn't look correct. You shouldn't need to prepend the txps param with a _ since it's being used here, and this filter seems like a sideways way to go about summing `fees` and `amounts` since that's all it's doing and not meaningfully filtering.

packages/bitcore-wallet-service/src/lib/model/wallet.ts
```diff
@@ -337,7 +337,7 @@ export class Wallet implements IWallet<() => boolean> {
     );
   }
 
-  /// Only for power scan
+  // / Only for power scan
```
kajoseph: You can just get rid of the extra /

packages/bitcore-wallet-service/src/lib/chain/eth/index.ts
kajoseph: This doesn't look right. This is not effectively filtering, it's just a loop that's summing `fees` and `amounts`. I think we should make it a loop and remove the `_` prefix from txps.

packages/bitcore-wallet-service/src/lib/model/email.ts
```diff
@@ -1,24 +1,6 @@
 import _ from 'lodash';
+import Uuid from 'uuid';
 
-const Uuid = require('uuid');
-
-interface IEmail {
```
kajoseph, line 5: In other models (e.g. wallet) we've used this interface to help define the class below. e.g.
```
export class Email implements IEmail {
```

There may be some discrepancies between the Email class and the IEmail interface. In the case of the Wallet class, I had to do some extra workaround stuff, so if this email model file gets to be too much of a change then you can just punt on it for now and eslint ignore the no unused var rule for the IEmail interface.

packages/bitcore-wallet-service/test/integration/server.test.ts
```diff
@@ -5719,7 +5717,7 @@ describe('Wallet service', function() {
   });
 
   describe('cashAddr backwards compat', function() {
-    /// LEGACY MODE
+    // / LEGACY MODE
```
kajoseph: rm extra /

packages/bitcore-wallet-service/test/integration/server.test.ts
```diff
@@ -5773,7 +5771,7 @@ describe('Wallet service', function() {
       });
     });
 
-    /// CASHADDR MODE
+    // / CASHADDR MODE
```
kajoseph: rm extra /

packages/bitcore-wallet-service/test/storage.test.ts
```diff
@@ -259,15 +259,15 @@ describe('Storage', function() {
 
 
     it('should store a 5 txs on the cache and update status correctly', (done) => {
-      let tipIndex = 80; // current cache tip
-      let items = [
-        { txid: '1234', blockheight: 803 },    /// <=== Latests
+      const tipIndex = 80; // current cache tip
+      const items = [
+        { txid: '1234', blockheight: 803 },    // / <=== Latests
```
kajoseph: rm extra /

packages/bitcore-wallet-service/test/storage.test.ts
```diff
@@ -282,24 +282,24 @@ describe('Storage', function() {
     });
 
     it('should prevent to store txs on wrong order', (done) => {
-      let tipIndex = 80; // current cache tip
-      let items = [
-        { txid: '1234', blockheight: 803 },    /// <=== Latests
+      const tipIndex = 80; // current cache tip
+      const items = [
+        { txid: '1234', blockheight: 803 },    // / <=== Latests
```
kajoseph: rm extra /

packages/bitcore-wallet-service/test/storage.test.ts
```diff
@@ -360,33 +360,33 @@ describe('Storage', function() {
 
 
     it('should store a 5 txs on the cache and retreive them correctly', (done) => {
-      let tipIndex = 80; // current cache tip
-      let items = [
-        { txid: '1234', blockheight: 803 },    /// <=== Latests
+      const tipIndex = 80; // current cache tip
+      const items = [
+        { txid: '1234', blockheight: 803 },    // / <=== Latests
```
kajoseph: rm extra /

packages/bitcore-wallet-service/test/storage.test.ts
```diff
@@ -360,33 +360,33 @@ describe('Storage', function() {
...
     });
 
 
     it('should store a 10 txs on the cache and retreive them correctly', (done) => {
-      let tipIndex = 80; // current cache tip
-      let items = [
-        { txid: '1234', blockheight: 803 },    /// <=== Latests
+      const tipIndex = 80; // current cache tip
+      const items = [
+        { txid: '1234', blockheight: 803 },    // / <=== Latests
```
kajoseph: rm extra /

packages/bitcore-wallet-service/test/storage.test.ts
```diff
@@ -398,8 +398,8 @@ describe('Storage', function() {
 
         // time passes
         updateHeight = 2000;
-        let items2 = [
-          { txid: '124', blockheight: 1803 },    /// <=== Latests
+        const items2 = [
+          { txid: '124', blockheight: 1803 },    // / <=== Latests
```
kajoseph: rm extra /

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -5205,7 +5185,11 @@ export class WalletService implements IWalletService {
 
   moralisGetTokenAllowance(req): Promise<any> {
     return new Promise((resolve, reject) => {
-      let keys, headers;
+      const headers = {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+        'X-Api-Key': config.moralis.apiKey,
+      };
```
kajoseph: can you move this object back down to where it was but just declare it as a const down there? That way `config.moralis.apiKey` isn't evaluated before `!config.moralis` is checked (the next line).

packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -5322,14 +5300,18 @@ export class WalletService implements IWalletService {
 
   moralisGetERC20TokenBalancesWithPricesByWallet(req): Promise<any> {
     return new Promise((resolve, reject) => {
-      let keys, headers;
+      const headers = {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+        'X-Api-Key': config.moralis.apiKey,
+      };
```
kajoseph, line 5311: Same as my other comment - move this object back down below and make it const down there.

kajoseph (changes requested):
Nice work! Mostly minor additional cleanup comments, but a couple of more consequential changes requested.

PR #4041, opened 2025-10-30
[BCN] Added new eslint rule

MicahMaphet:
Added an ESLint rule to require semicolons as delimiters in interface and type member declarations, with semicolons required at the end of multiline members but omitted for the last member in single-line declarations. Implemented this rule in bitcore-node exclusively.


packages/bitcore-node/src/utils/parseArgv.ts
```diff
@@ -1,6 +1,6 @@
 type Required = 'string' | 'int' | 'number' | 'float'
 type Optional = Required | 'bool' | 'boolean';
-type Arg<T> = string | { arg: string; type: T; }
+type Arg<T> = string | { arg: string; type: T }
```
kajoseph, line 3: it's odd that it eslint isn't requiring a semicolon at the end of this line and line 1.

eslint.config.mjs
```diff
@@ -56,6 +56,16 @@ export default defineConfig([
       'spaced-comment': ['error', 'always', { exceptions: ['-', '+'] }],
       '@stylistic/object-curly-spacing': ['error', 'always'],
       '@stylistic/indent': ['error', 2, { SwitchCase: 1 }],
+      '@stylistic/member-delimiter-style': ['error', {
```
kajoseph, line 59: I probably would have put this in the typescript-only section of this config, but this is fine. We can always move it if it's a problem.

kajoseph (approved):

PR #4047, opened 2025-11-07
[BCN] Environmental variable option for bitcore.config.json path

MicahMaphet:
Instead of searching for bitcore.config.json in bitcore/packages/bitcore-node/bitcore.config.json, bitcore/bitcore.config.json, and ~/bitcore.config.json, it uses the environmental variable BITCORE_CONFIG, if the variable does not exits it uses bitcore/bitcore.config.json.

### DIFF
--- a/packages/bitcore-node/src/config.ts
+++ b/packages/bitcore-node/src/config.ts
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import { cpus, homedir } from 'os';
+import path from 'path';
 import logger from './logger';
 import { ConfigType } from './types/Config';
 import { merge } from './utils';
@@ -20,10 +21,10 @@ function findConfig(): ConfigType | undefined {
   const bitcoreConfigStat = fs.statSync(bitcoreConfigPath);
   
   if (bitcoreConfigStat.isDirectory()) {
-    if (!fs.existsSync(bitcoreConfigPath + '/bitcore.config.json')) {
+    if (!fs.existsSync(path.join(bitcoreConfigPath, 'bitcore.config.json'))) {
       throw new Error(`No bitcore config exists in directory ${bitcoreConfigPath}`);
     }
-    bitcoreConfigPath += '/bitcore.config.json';
+    bitcoreConfigPath = path.join(bitcoreConfigPath, 'bitcore.config.json');
   }
   logger.info('Using config at: ' + bitcoreConfigPath);
   


packages/bitcore-node/src/config.ts
```diff
@@ -1,38 +1,35 @@
...
 import { ConfigType } from './types/Config';
 import { merge } from './utils';
-import parseArgv from './utils/parseArgv';
 
-const program = parseArgv([], ['config']);
+const bitcoreConfigPath = process.env.BITCORE_CONFIG || '../../bitcore.config.json';
 
-function findConfig(): ConfigType | undefined {
-  let foundConfig;
-  const envConfigPath = process.env.BITCORE_CONFIG_PATH;
```
kajoseph, line 10: There already was an env var for the config. Better to not rename the existing env var as it breaks the pipeline.

Also, it would be good to retain the `program.config` below. Then you can have
```
const bitcoreConfigPath = program.config || process.env.BITCORE_CONFIG_PATH || '../../bitcore.config.json';
```

packages/bitcore-node/src/config.ts
```diff
@@ -1,38 +1,35 @@
...
+logger.info('Using config at: ' + bitcoreConfigPath);
+
+if (!fs.existsSync(bitcoreConfigPath)) {
+  throw new Error(`No bitcore config exists at ${bitcoreConfigPath}`);
+}
+
+const bitcoreConfigStat = fs.statSync(bitcoreConfigPath);
+
+if (bitcoreConfigStat.isDirectory()) {
+  throw new Error(`Provided bitcore config path: ${bitcoreConfigPath} is a directory`);
```
kajoseph: If `bitcoreConfigPath` is a dir, it'd be better to append `bitcore.config.json` (`path.join(bitcoreConfigPath, 'bitcore.config.json'`) and try that. So in pseudocode:

```
if dir {
  path = path + 'bitcore.config.json'
}

if !exists {
  throw error
}
```

packages/bitcore-node/src/config.ts
```diff
@@ -1,38 +1,47 @@
...
+    throw new Error(`No bitcore config exists at ${bitcoreConfigPath}`);
+  }
+  
+  const bitcoreConfigStat = fs.statSync(bitcoreConfigPath);
+  
+  if (bitcoreConfigStat.isDirectory()) {
+    if (!fs.existsSync(bitcoreConfigPath + '/bitcore.config.json')) {
+      throw new Error(`No bitcore config exists in directory ${bitcoreConfigPath}`);
     }
+    bitcoreConfigPath += '/bitcore.config.json';
```
kajoseph: Better to use `path.join` here. If there's a trailing `/` in the path (e.g. `BITCORE_CONFIG_PATH=/path/to/config/`), then this will be `/path/to/config//bitcore.config.json`...which works but is a little messy and may not work for all OSs. It'd be better to do:
```
bitcoreConfigPath = path.join(bitcoreConfigPath, 'bitcore.config.json);
```
Applies to line 23 as well ( `if (fs.existsSync(path.join(...` ).

PR #4048, opened 2025-11-07
[BWS] Environmental variable option for bws.config path 

MicahMaphet:
Instead of just checkout bitcore-wallet-service/bws.config.js,

- First look in BWS_CONFIG_PATH or BWS_CONFIG_PATH/bws.config.js. Throws errors on failure.
- Second look in BITCORE_CONFIG_PATH or BWS_CONFIG_PATH/bitcore.config.json. Does not throw errors on failure, just uses default paths.
- If there are no environmental variables, try bitcore/bitcore.config.json and lastly bitcore-wallet-service/bws.config.js

### DIFF
--- a/packages/bitcore-wallet-service/src/config.ts
+++ b/packages/bitcore-wallet-service/src/config.ts
@@ -478,7 +478,11 @@ const Config = (): any => {
   };
 
   // Override default values with bws.config.js' values, if present
-  const { BITCORE_CONFIG_PATH = '../../../../', BWS_CONFIG_PATH = '../../' } = process.env;
+  const {
+    BITCORE_CONFIG_PATH = path.join(__dirname, '../../../../'),
+    BWS_CONFIG_PATH = path.join(__dirname, '../../')
+  } = process.env;
+  
   let bwsConfig;
 
   try {


packages/bitcore-wallet-service/src/config.ts
```diff
@@ -476,12 +478,36 @@ const Config = (): any => {
...
+      } else if (fs.existsSync(BITCORE_CONFIG_PATH) && BITCORE_CONFIG_PATH.endsWith('.json')) {
+        bwsConfig = JSON.parse(fs.readFileSync(BITCORE_CONFIG_PATH, 'utf8'));
+      }
+      bwsConfig = bwsConfig?.bitcoreWalletService;
+    }
+
+    if (!bwsConfig && BWS_CONFIG_PATH) {
+      if (fs.existsSync(path.join(BWS_CONFIG_PATH, 'bws.config.js'))) {
+        // eslint-disable-next-line @typescript-eslint/no-require-imports
+        bwsConfig = require(path.join(BWS_CONFIG_PATH, 'bws.config.js'));
```
kajoseph, line 504: In testing your this, I realized it doesn't actually work for `bws.config.js` because of the discrepancy between how `require` and `fs.existsSync` handle the CWD. If I run bws from `bitcore/packages/bitcore-wallet-service`, then `fs.existsSync` will use a CWD of `bitcore/packages/bitcore-wallet-service` but the `require` uses a CWD of `bitcore/packages/bitcore-wallet-service/ts_build/src`. 

One way to fix this would be to use a "standard" starting point such as `__dirname` like so:
```javascript
const {
    BITCORE_CONFIG_PATH = path.join(__dirname, '../../../../'),
    BWS_CONFIG_PATH = path.join(__dirname, '../../')
  } = process.env;
```

packages/bitcore-wallet-service/src/config.ts
```diff
@@ -476,12 +478,36 @@ const Config = (): any => {
...
-  } catch {
-    logger.info('bws.config.js not found, using default configuration values');
+    if (BITCORE_CONFIG_PATH) {
+      if (fs.existsSync(path.join(BITCORE_CONFIG_PATH, 'bitcore.config.json'))) {
+        bwsConfig = JSON.parse(fs.readFileSync(path.join(BITCORE_CONFIG_PATH, 'bitcore.config.json'), 'utf8'));
+      } else if (fs.existsSync(BITCORE_CONFIG_PATH) && BITCORE_CONFIG_PATH.endsWith('.json')) {
+        bwsConfig = JSON.parse(fs.readFileSync(BITCORE_CONFIG_PATH, 'utf8'));
+      }
+      bwsConfig = bwsConfig?.bitcoreWalletService;
+    }
```
kajoseph, line 499: Can we add a log if we found bwsConfig here? e.g.
```javascript
  bwsConfig = bwsConfig?.bitcoreWalletService;
  if (bwsConfig) {
    logger.info(`Using JSON config from ${BITCORE_CONFIG_PATH}`);
  }
}
```

packages/bitcore-wallet-service/src/config.ts
```diff
@@ -476,12 +478,36 @@ const Config = (): any => {
...
+
+    if (!bwsConfig && BWS_CONFIG_PATH) {
+      if (fs.existsSync(path.join(BWS_CONFIG_PATH, 'bws.config.js'))) {
+        // eslint-disable-next-line @typescript-eslint/no-require-imports
+        bwsConfig = require(path.join(BWS_CONFIG_PATH, 'bws.config.js'));
+      } else if (fs.existsSync(BWS_CONFIG_PATH) && BWS_CONFIG_PATH.endsWith('.js')) {
+        // eslint-disable-next-line @typescript-eslint/no-require-imports
+        bwsConfig = require(BWS_CONFIG_PATH);
+      }
+    }
```
kajoseph, line 512: Can we add a log here, too? e.g.

```javascript
  }
  if (bwsConfig) {
    logger.info(`Using JS config from ${BWS_CONFIG_PATH}`);
  }
}
```

packages/bitcore-wallet-service/src/config.ts
kajoseph: This is still an issue

kajoseph (approved):

PR #4052, opened 2025-11-14
[REF] Changelly: using stablecoin_key for stablecoin swaps

gabrielbazan7:


packages/bitcore-wallet-service/src/externalservices/changelly.ts
```diff
@@ -20,7 +27,7 @@ export class ChangellyService {
 
     const keys = {
       API: config.changelly.v2.api,
-      SECRET: config.changelly.v2.secret
+      SECRET: stablecoins.includes(req.body.coinFrom) ? config.changelly.v2.STABLECOIN_KEY : config.changelly.v2.secret
```
kajoseph: Breaking change. We should fallback to `config.changelly.v2.secret` if `STABLECOIN_KEY` is not present.

packages/bitcore-wallet-service/src/config.ts
```diff
@@ -456,6 +456,7 @@ const Config = (): any => {
     //   },
     //   v2: {
     //     secret: 'changelly_secret_v2',
+    //     STABLECOIN_KEY: 'changelly_secret_stablecoin_v2'
```
kajoseph: It'd be nice to have some naming & case consistency here. Perhaps `secret_stablecoin` or `secret_alt` would be better.

packages/bitcore-wallet-service/src/externalservices/changelly.ts
```diff
@@ -18,9 +25,15 @@ export class ChangellyService {
       throw new Error('ClientError: Service v2 not configured.');
     }
 
+    const { api, secret, secret_stablecoin } = config.changelly.v2;
+
+    const apiKey = stablecoins.includes(req.body.coinFrom)
+    ? (secret_stablecoin ?? secret)
```
kajoseph: eslint is triggering a red squiggly here

PR #4053, opened 2025-11-17
[CLI] Improvements for imported, read-only, and fully encrypted wallet files

kajoseph:
This PR:

* Adds support for wallet files exported from the wallet app (`creds` => `credentials`)
  * NOTE: This is a breaking change since it now looks for `credentials` instead of `creds` in previously-created wallet files. You can manually change the "creds" to "credentials" in your file(s).
* Fixes a bug in bitcore-node where the API would hang forever (until timeout) if a wallet didn't have wallet addresses registered in BCN.
* Makes improvements in the tx `history` to not re-call the API when toggling formats (or exporting), as well as showing full decimal precision for expanded format.

### DIFF
--- a/packages/bitcore-cli/src/commands/txhistory.ts
+++ b/packages/bitcore-cli/src/commands/txhistory.ts
@@ -49,6 +49,7 @@ export async function getTxHistory(
     if (!tokenObj) {
       throw new Error(`Unknown token "${opts.tokenAddress || opts.token}" on ${wallet.chain}:${wallet.network}`);
     }
+    tokenObj = JSON.parse(JSON.stringify(tokenObj)); // make copy for precision modification below
   }
   const currency = tokenObj?.displayCode || wallet.client.credentials.coin;
 
@@ -68,6 +69,16 @@ export async function getTxHistory(
       printRaw = false; // reset printRaw when toggling format
     }
 
+    if (tokenObj) {
+      if (!compact) {
+        // Show full decimals in expanded view
+        tokenObj['_precision'] = tokenObj.precision; 
+        tokenObj.precision = tokenObj.decimals.short.maxDecimals;
+      } else if (tokenObj['_precision'] !== undefined) {
+        // Restore original precision when toggling back to compact view
+        tokenObj.precision = tokenObj['_precision'];
+      }
+    }
     printRaw = viewAction === ViewAction.TOGGLE_RAW ? !printRaw : printRaw;
     const exportToFile = !!opts.export || viewAction === ViewAction.EXPORT;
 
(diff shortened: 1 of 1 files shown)


packages/bitcore-cli/src/cli.ts
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+import 'source-map-support/register';
```
MichaelAJay, line 3: is this used?

packages/bitcore-cli/src/commands/txhistory.ts
```diff
@@ -70,12 +71,19 @@ export async function getTxHistory(
     printRaw = viewAction === ViewAction.TOGGLE_RAW ? !printRaw : printRaw;
     const exportToFile = !!opts.export || viewAction === ViewAction.EXPORT;
 
-    const history = await wallet.client.getTxHistory({
-      includeExtendedInfo: true,
-      tokenAddress: tokenObj?.contractAddress,
-      limit: opts.pageSize,
-      skip: (page - 1) * opts.pageSize
-    });
+    if (
```
MichaelAJay, line 85: For your consideration:
`![ViewAction.TOGGLE_FORMAT, ViewAction.TOGGLE_RAW, ViewAction.EXPORT].includes(viewAction)`

packages/bitcore-cli/src/commands/txhistory.ts
```diff
@@ -106,6 +114,7 @@ export async function getTxHistory(
         const timestamp = new Date(tx.time * 1000);
         const time = compact ? Utils.formatDateCompact(timestamp) : Utils.formatDate(timestamp);
         const txid = compact ? Utils.compactString(tx.txid) : tx.txid;
+        tokenObj = tokenObj && !compact ? { ...tokenObj, precision: tokenObj.decimals.short.maxDecimals } : tokenObj;
```
MichaelAJay: A few more LOC, but I feel like it would be much more declarative in an if block:
```
if (tokenObj && !compact) {
  tokenObj.precision = tokenObj.decimals.short.maxDecimals;
}
```

MichaelAJay: Why is this in the for loop?

packages/bitcore-cli/src/wallet.ts
```diff
@@ -162,26 +162,44 @@ export class Wallet implements IWallet {
...
+    }
+    const { secret } = await this.client.createWallet(
+      this.name,
+      args.copayerName,
+      m,
+      n,
+      {
+        chain,
+        network: network as Network,
+        ...Utils.getSegwitInfo(addressType)
```
MichaelAJay, line 197: Intended behavior is Utils.getSegwit could potentially overwrite chain and network?

kajoseph: yes, that's fine. This particular change (line 184 => lines 189-199) is only cosmetic.

packages/bitcore-cli/src/cli.ts
kajoseph: Yes. It results in thrown error stacks showing paths to the TS files instead of the `build/**/*.js` files.

packages/bitcore-cli/src/commands/txhistory.ts
kajoseph: Good Q. It shouldn't be.

packages/bitcore-cli/src/commands/txhistory.ts
kajoseph: I did consider that, but it actually doesn't work with TS since ViewAction is an `enum`. Thus, your line throws a TS error that `string` is incompatible with `ViewAction` type.

PR #4068, opened 2025-12-12
[BC/CWC] Wallet private key updates [1]

MichaelAJay:
### Encryption
- adds encryptBuffer - enables normalized buffer input
- adds decryptToBuffer - enables improved security by not decrypting directly to string
### Storage
- adds `addKeysSafe` - incoming keys' private key is encrypted, so not immediately available to be used to retrieve pubkey, which should be available anyway. Throws if missing a pubkey
### Wallet
- Adds version: 2 & version checks for backwards compatibility (note: to create an old Wallet would require version: 0 - so 1 is skipped - looking for feedback on numbering/versioning).
- `create` encrypts HDPrivateKey xprivkey and privateKey so they can be decrypted as buffers instead of serializing the whole masterKey and THEN encrypting
- `importKeys` does the same for signing keys
- `signTx` decrypts to buffers then uses deriver for chain-aware decoding. In the next phase, this decoding should be made unnecessary by passing the buffer all the way through to use, if possible
### Tests
- update prior tests to include versioning
- introduce new tests for version 2 specifics
### Derivation
- Add two passthrough methods on DeriverProxy to buffer & serialize private keys
- Adds `privateKeyToBuffer` and `privateKeyBufferToNativePrivateKey` to IDeriver & implement for each Deriver - no need to override for any of the extending classes (UTXOs to AbstractBitcoreLibDeriver & EVM to EthDeriver)

### DIFF
--- a/packages/bitcore-client/src/storage/textFile.ts
+++ b/packages/bitcore-client/src/storage/textFile.ts
@@ -217,22 +217,6 @@ export class TextFile {
 
   /**
    * Rewrite the address file in place for a specific migration version.
-   *
-   * Why a transformer?
-   * The TextFile backend is append-only. That is simple for normal writes, but it means
-   * migration cannot "overwrite" an existing address entry by appending a new one because
-   * `getKey()` still resolves the first matching line it encounters. For migration we
-   * therefore treat `addresses.txt` like a stream of JSONL records:
-   *
-   * 1. Read bytes from the existing file.
-   * 2. Parse each JSONL line into an object.
-   * 3. Pass each object through a Transform stream that replaces only the records that
-   *    belong to this wallet/address set.
-   * 4. Serialize the transformed objects back to JSONL.
-   * 5. Write the result to a temporary file and atomically rename it into place.
-   *
-   * This lets us preserve the file order and all unrelated records while updating the
-   * migrated addresses in a single pass.
    */
   async migrate(params: { version: number; name: string; keys: any[] }) {
     const { version, name, keys } = params;
@@ -253,18 +237,6 @@ export class TextFile {
 
     const tempAddressFileName = `${this.addressFileName}.v${version}.migrating`;
 
-    /**
-     * The transform stream is the "rewrite policy" for migration.
-     *
-     * Each record coming in is one parsed JSON object from the source file.
-     * We either:
-     * - push it through unchanged, or
-     * - replace it with the migrated version for the same wallet/address.
-     *
-     * The important thing to notice is that Transform streams do not need to change the
-     * number of records. They can simply emit a modified version of the incoming record,
-     * which makes them a nice fit for "rewrite this file in one pass" jobs like this.
-     */
     const migrationTransform = new stream.Transform({
       writableObjectMode: true,
       readableObjectMode: true,
@@ -280,7 +252,7 @@ export class TextFile {
           const replacement = replacementRecords.get(record.address);
           /**
            * If replacement not in map, indicates a prior-existing address which was not included in the keys to migrate.
-           * Fall back to prior.
+           * Warn & fall back to prior.
            */
           if (!replacement) {
             console.warn(`TextFile migration warning: found address ${record.address} for wallet ${name}, but no replacement key was provided. Keeping original record.`);


packages/crypto-wallet-core/src/derivation/eth/index.ts
```diff
@@ -55,4 +55,20 @@ export class EthDeriver implements IDeriver {
...
+  /**
+   * @param {any} privKey - expects hex-encoded string, as returned from EthDeriver.derivePrivateKey
+   * @returns {Buffer}
+   * @throws {Error} If privKey is not a Buffer (planned forwards compatibility) or string. Propagates all other errors
+   */
+  privateKeyToBuffer(privKey: any): Buffer {
+    if (Buffer.isBuffer(privKey)) return privKey;
+    if (typeof privKey !== 'string') throw new Error(`Expected string, got ${typeof privKey}`);
+    // Expects to match return from derivePrivateKey's privKey.
+    return Buffer.from(privKey, 'hex');
```
kajoseph, line 80: should strip leading 0x if present

packages/crypto-wallet-core/src/derivation/sol/index.ts
```diff
@@ -54,4 +54,22 @@ export class SolDeriver implements IDeriver {
       pubKey: Buffer.from(pubKey).toString('hex')
     } as Key;
   };
+
+  /**
+   * @param {any} privKey - expects base 58 encoded string
+   * @returns {Buffer}
+   * @throws {Error} If privKey is not a Buffer (planned forwards compatibility) or string. Propagates all other errors
+   * 
+   * TODO
```
kajoseph: What's there todo?

packages/crypto-wallet-core/src/derivation/btc/index.ts
```diff
@@ -33,6 +33,25 @@ export abstract class AbstractBitcoreLibDeriver implements IDeriver {
     pubKey = new this.bitcoreLib.PublicKey(pubKey);
     return new this.bitcoreLib.Address(pubKey, network, addressType).toString();
   }
+
+  /**
+   * @returns {Buffer} raw secpk1 private key buffer (32 bytes, big-endian)
+   * @throws {Error} If privKey is not a Buffer (planned forwards compatibility) or string. Propagates all other errors
+   */
+  privateKeyToBuffer(privKey: any): Buffer {
```
kajoseph: Why are we not more strict with the input type?

packages/bitcore-client/src/encryption.ts
```diff
@@ -23,13 +23,27 @@ export function encryptEncryptionKey(encryptionKey, password) {
...
+    payload.fill(0);
+    final.fill(0);
+    if (!toBuffer) {
+      // Don't fill output if it's what's returned directly
+      output.fill(0);
+    }
+  }
 }
 
 export function encryptPrivateKey(privKey, pubKey, encryptionKey) {
```
kajoseph, line 53: Should we mark as `@deprecated` and use `encryptBuffer` instead? We could also just make this a wrapper for `encryptBuffer`...

packages/bitcore-client/test/unit/wallet.test.ts
```diff
@@ -368,6 +374,117 @@ describe('Wallet', function() {
     });
   });
 
+  describe('signTx v2 key handling', function() {
+    let txStub: sinon.SinonStub;
+    afterEach(async function() {
+      txStub?.restore();
```
kajoseph: sandbox.resolve()

packages/bitcore-client/src/storage.ts
```diff
@@ -189,4 +192,68 @@ export class Storage {
     const { name, limit, skip } = params;
     return this.storageType.getAddresses({ name, limit, skip });
   }
+
+  /**
+   * New methods
+   * TODO: Deprecate above as necessary
```
kajoseph: Should we mark the above as `@deprecated`?

packages/bitcore-client/test/unit/wallet.test.ts
```diff
@@ -368,6 +374,117 @@ describe('Wallet', function() {
...
+        await Wallet.deleteWallet({ name: walletName, storageType });
+      });
+
+      it('should decrypt stored ciphertext and hand hex privKey to Transactions.sign', async function() {
+        const privHex = crypto.randomBytes(32).toString('hex');
+        const privBufForPubKey = CWC.Deriver.privateKeyToBuffer('ETH', privHex);
+        const pubKey = CWC.Deriver.getPublicKey('ETH', wallet.network, privBufForPubKey);
+        privBufForPubKey.fill(0);
+        const privBuf = CWC.Deriver.privateKeyToBuffer('ETH', privHex);
+        // v2 key encryption uses the key's pubKey as the IV salt (not the wallet pubKey)
```
kajoseph: ?? As far as I can tell, they both set the IV the same way.

packages/crypto-wallet-core/src/derivation/index.ts
```diff
@@ -119,6 +126,14 @@ export class DeriverProxy {
       return Paths.BTC.default + accountStr;
     }
   }
+
+  privateKeyToBuffer(chain, privateKey: Buffer | string): Buffer {
+    return this.get(chain).privateKeyToBuffer(privateKey);
+  }
+
+  privateKeyBufferToNativePrivateKey(chain: string, network: string, buf: Buffer): any {
```
kajoseph: Avoid `any`. Every fn returns a string - might as well type it.

packages/crypto-wallet-core/src/types/derivation.ts
```diff
@@ -14,4 +14,21 @@ export interface IDeriver {
...
+
+  /**
+   * Used to normalize output of Key.privKey
+   */
+  privateKeyToBuffer(privKey: any): Buffer;
+
+  /**
+   * Temporary - converts decrypted private key buffer to chain-native private key format
+   */
+  privateKeyBufferToNativePrivateKey(buf: Buffer, network: string): any;
```
kajoseph: Do we need this?

MichaelAJay: Sort of. It depends on how much we want THIS PR to do.

The point of this is that it pushes the buffer private keys all the way to the boundary - Wallet side.

Next would be updating the called libraries and sending buffers straight on through.

I think that should be a second effort after this one.

packages/bitcore-client/src/encryption.ts
```diff
@@ -46,13 +46,40 @@ export function decryptEncryptionKey(encEncryptionKey, password, toBuffer?: bool
   }
 }
 
+// @deprecated - Use encryptBuffer
```
kajoseph: Invalid jsdoc. Should be 
```
/** @deprecated - Use encryptBuffer */
```

packages/bitcore-client/src/storage.ts
```diff
@@ -116,6 +116,7 @@ export class Storage {
     return this.storageType.saveWallet({ wallet });
   }
 
+  // @deprecated
```
kajoseph: Invalid jsdoc. What fn replaced it?

(+more below)

packages/bitcore-client/src/encryption.ts
```diff
@@ -23,33 +23,101 @@ export function encryptEncryptionKey(encryptionKey, password) {
...
 export function encryptPrivateKey(privKey, pubKey, encryptionKey) {
-  const key = Buffer.from(encryptionKey, 'hex');
-  const doubleHash = Buffer.from(SHA256(SHA256(pubKey)), 'hex');
-  const iv = doubleHash.subarray(0, 16);
-  const cipher = crypto.createCipheriv(algo, key, iv);
-  const encData = cipher.update(privKey, 'utf8', 'hex') + cipher.final('hex');
-  return encData;
+  // Store buffers this method makes to sanitize
+  const createdBuffers: Buffer[] = [];
```
kajoseph: This function seems overcomplicated. Why not just:

```
export function encryptPrivateKey(privKey, pubKey, encryptionKey) {
  encryptionKey = Buffer.from(encryptionKey, 'hex');
  privKey = Buffer.from(privKey, 'utf8');
  return encryptBuffer(privKey, pubKey, encryptionKey);
}
```

packages/crypto-wallet-core/src/types/derivation.ts
kajoseph: This seems like some odd scope overlap. It's temporary but to be used in the future? Let's just save it for when we do the future work.

packages/bitcore-client/src/encryption.ts
```diff
@@ -23,33 +23,72 @@ export function encryptEncryptionKey(encryptionKey, password) {
...
+}
+
+function decryptToBuffer(encHex: string, pubKey: string, encryptionKey: Buffer): Buffer {
+  const iv = Buffer.from(SHA256(SHA256(pubKey)), 'hex').subarray(0, 16);
+  const decipher = crypto.createDecipheriv(algo, encryptionKey, iv);
+
+  const decrypted = decipher.update(encHex, 'hex');
+  const final = decipher.final();
+  try {
+    return Buffer.concat([decrypted, final]);
```
kajoseph, line 92: Include the `decipher.final()` in the try.
```
try {
  return Buffer.concat([decrypted, decipher.final()]);

File: packages/bitcore-client/src/encryption.ts [hunk 14]
```diff
@@ -23,33 +23,72 @@ export function encryptEncryptionKey(encryptionKey, password) {
...
   const iv = password_hash.subarray(32, 48);
   const decipher = crypto.createDecipheriv(algo, key, iv);
-  const decrypted = decipher.update(encEncryptionKey, 'hex', 'hex' as any) + decipher.final('hex');
-  return decrypted;
+  
+  const payload = decipher.update(encEncryptionKey, 'hex');
+  const final = decipher.final();
+  const output = Buffer.concat([payload, final]);
+  try {
+    return toBuffer ? output : output.toString('hex');
```
kajoseph, line 41: include `decipher.final()` in the try.

```
let final;
try {
  final = decipher.final();
  const output = Buffer.concat([payload, final]);
  return toBuffer ? output : output.toString('hex');
} finally {
  payload.fill(0);
  final?.fill(0);
}
```
Not sure if defining `final` outside the try is necessary. If this code is going to error, it's highly likely that it's due to an invalid checksum/tag in the `decipher.final()` call, so it may be fine to just do `Buffer.concat([payload, decipher.final()]);`. Whatever is decided also applies to the other encrypt/decrypt buffer methods.

packages/bitcore-client/src/wallet.ts
```diff
@@ -303,6 +328,123 @@ export class Wallet {
     return this;
   }
 
+  async migrateWallet(encryptionKey: Buffer): Promise<Wallet> {
+    /**
```
kajoseph, line 344: We should log that a migration is taking place. E.g. "Migrating wallet from version x to version y"

packages/bitcore-client/src/wallet.ts
```diff
@@ -303,6 +328,123 @@ export class Wallet {
...
+
+    /**
+     * 1: Wallet to .bak
+     */
+    const rawWallet = await this.storage.loadWallet({ name: this.name, raw: true });
+    if (!rawWallet) {
+      throw new Error('Migration failed - wallet not found');
+    }
+    
+    await writeFile(`${this.name}.bak`, rawWallet, 'utf8')
```
kajoseph: Should write to an expected directory instead of just dumping in the cwd. e.g. `~/.bitcore/bitcoreWallets/backup/` (next to where the text wallet files are saved)

packages/bitcore-client/src/wallet.ts
```diff
@@ -303,6 +328,123 @@ export class Wallet {
...
+    const storedEncryptedPassword = this.password; // Wallet.toObject() rehashes password - save and replace
+    const walletObj = this.toObject(false);
+    walletObj.password = storedEncryptedPassword;
+    await this.storage.saveWallet({ wallet: walletObj })
+      .catch(err => {
+        console.error('Migration failure: wallet not successfully saved', err);
+        throw new Error('Migration failure: wallet not successfully saved. Use backups to restore prior wallet and keys');
+      });
+
+    return this;
```
kajoseph, line 475: Log migration successful

packages/bitcore-client/src/wallet.ts
```diff
@@ -303,6 +328,123 @@ export class Wallet {
...
+      return this;
+    }
+
+    if (this.version > CURRENT_WALLET_VERSION) {
+      console.warn(`Wallet version ${this.version} greater than expected current wallet version ${CURRENT_WALLET_VERSION}`);
+      return this;
+    }
+
+    /**
+     * 1: Wallet to .bak
```
kajoseph, line 360: Log that a pre-migration wallet backup is being written to \<file>

kajoseph (commented):
Functionally, this seems to work well. Nice job!

In addition to the minor comments below, we should add test cases for the migration as well as tests for critical functions with the new version wallets (i.e. deriving, sending, sign message)

packages/bitcore-client/src/wallet.ts
```diff
@@ -218,12 +230,6 @@ export class Wallet {
       storageType
     });
 
-    if (!xpriv) {
-      console.log(mnemonic.toString());
-    } else {
-      console.log(hdPrivKey.toString());
```
kajoseph, line 224: We need to keep this in here - otherwise the user has no way of backing up their seed phrase. To prevent converting to string, we can pass the buffer directly to process.stdout:
```typescript
if (!xpriv) {
  process.stdout.write(Buffer.from(mnemonic.phrase));
  process.stdout.write(os.EOL);
} else {
  process.stdout.write(hdPrivKey.toBuffer());
  process.stdout.write(os.EOL);
}
```

PR #4070, opened 2025-12-13
Use TSX with Mocha

MicahMaphet:
## :spiral_notepad:  Description
Use tsx to run mocha test scripts in typescript packages. The purpose of this is to standardize bitcore typescript testing. No typescript tests will need to build.

## :package: Packages
The packages with typescript being added:

- `bitcore-cli`: Replaced ts-node with tsx.
- `bitcore-client`: Already uses tsx, added type checking.
- `bitcore-node`: Added tsx to unit tests. Integration tests already use tsx; added type checking. Modified test imports to work with tsx.
- `bitcore-wallet-client`: Replaced ts-node with tsx.
- `bitcore-wallet-service`: Replaced ts-node with tsx.
- `crypto-wallet-core`: Already uses tsx: added type checking.


packages/bitcore-wallet-service/package.json
```diff
@@ -93,7 +93,7 @@
     "mocha-lcov-reporter": "^1.3.0",
     "sinon": "^7.1.1",
     "supertest": "*",
-    "ts-node": "^8.0.3",
+    "tsx": "4.19.2",
```
kajoseph: Why is this an older version than the others and version locked?

PR #4082, opened 2026-01-05
Add marketstats endpoint

msalcala11:


packages/bitcore-wallet-service/src/externalservices/coingecko.ts
```diff
@@ -23,6 +37,209 @@ export class CoinGeckoService {
...
+      const cacheKey = `cgMarketStats:${vsCurrency}`;
+      const credentials = this.coinGeckoGetCredentials();
+      const headers = {
+        'Content-Type': 'application/json',
+        'Accept': 'application/json',
+        'x-cg-pro-api-key': credentials.API_KEY
+      };
+
+      this.storage.checkAndUseGlobalCache(cacheKey, Defaults.COIN_GECKO_MARKET_STATS_CACHE_DURATION, async (err, values, oldvalues) => {
+        if (err) logger.warn('Cache check failed', err);
```
kajoseph: This `err` won't be included in the log like with `console.warn`. The winston logger lib uses string iterpolation, so you'll need to do:
```
logger.warn('Cache check failed: %o', err);
```

I'd also recommend making this message a little more contextual:
```
logger.warn('Coingecko market stats cache check failed: %o', err);
```

(It looks like both critiques would also apply to previously existing logs below which I'm guessing you were just trying to follow)

msalcala11: Thanks, just made that logging change

PR #4105, opened 2026-02-12
Add tests to bitcore-cli

kajoseph:
### Description
Need more tests in bitcore-cli

### Changelog

* Adds tests to bitcore-cli
* Fixes a few bugs
   * BWC/S: Sending max amount should work for UTXO chains
   * CLI: Gate SOL from multi-party wallets
   * Various other bitcore-cli bugs
* Removes `tsx` dependency from BWC

### Testing Notes
`cd packages/bitcore-cli && npm run test`

<hr style="border-width:3px">

### Checklist


packages/bitcore-cli/src/commands/txproposals.ts
```diff
@@ -3,18 +3,28 @@ import os from 'os';
...
+
 export function command(args: CommonArgs) {
   const { wallet, program } = args;
   program
-    .description('View, sign, and reject transaction proposals for a wallet')
+    .description('View or perform actions on transaction proposals for a wallet')
     .usage('<walletName> --command txproposals [options]')
     .optionsGroup('Tx Proposals Options')
-    .option('--action <action>', 'Action to perform on transaction proposals: sign, reject, delete, broadcast')
+    .option('--action <action>', `Action to perform on transaction proposals: ${Object.entries(ViewAction).map(([k, v]) => `${k.toLowerCase().replace(v, `(${v})`)}`).join(' | ')}`)
```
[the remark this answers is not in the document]
kajoseph: Meh.

packages/bitcore-cli/src/commands/create/index.ts
```diff
@@ -13,7 +13,8 @@ export async function createWallet(args: CommonArgs<{ mnemonic?: string }>) {
 
   const chain = await getChain();
   const network = await getNetwork();
-  const isMultiParty = await getIsMultiParty();
+  // No solana support for multi-party wallets right now (TSS is ECDSA only)
+  const isMultiParty = chain === 'sol' ? false : await getIsMultiParty();
```
MichaelAJay, line 17: You `chain.toLowerCase()` in the file above - should do that here too just to be safe?

packages/bitcore-cli/src/commands/join/index.ts
```diff
@@ -12,6 +12,10 @@ export async function joinWallet(args: CommonArgs<{ mnemonic?: string }>) {
 
   const chain = await getChain();
 
+  if (chain === 'sol') {
```
MichaelAJay, line 15: same here - should be chain.toLowerCase()?

packages/crypto-wallet-core/src/utils/index.ts
```diff
@@ -91,4 +91,55 @@ export function toHex(input: number | string | bigint): string {
   }
 }
 
+export function difference(arr1: any[], arr2: any[]): any[] {
```
MichaelAJay: docs would be really nice here:
"Returns a new array of the elements in arr1 which are not in arr2"

packages/crypto-wallet-core/src/utils/index.ts
```diff
@@ -91,4 +91,55 @@ export function toHex(input: number | string | bigint): string {
...
+  arr1 = arr1 || [];
+  arr2 = arr2 || [];
+  const arr2Set = new Set(arr2);
```
MichaelAJay, line 107: Works fine for primitives, but Set.has() uses strict reference equality. Structurally identical objects are not considered equal - is that the desired behavior? That should maybe be specified in the docs too if so.

Run this script from bitcore root - 
```bash
node - <<'NODE'
const { difference } = require('./packages/crypto-wallet-core/ts_build/src/utils');

const arr1 = [{ a: 1 }, { b: 2 }];
const arr2 = [{ a: 1 }];
const diff = difference(arr1, arr2);

// Many people would expect diff to ONLY include { b: 2 }, but it includes { a: 1 } as well.

console.log(JSON.stringify(diff));
NODE
```

packages/crypto-wallet-core/src/utils/index.ts
MichaelAJay: This feels like a great candidate for a TS generic. Tightens up the usage contract.
```typescript
export function difference<T>(arr1: T[], arr2: T[]): T[]
```

packages/crypto-wallet-core/src/utils/index.ts
MichaelAJay: Here's a test which demonstrates the failure. That's assuming the intent is "structural equality" - same shape elements get filtered out.
```typescript
   it('should return objects not present in arr2 (structural equality)', () => {                                                                                                                                        
     const result = difference(                                                                                                                                                                                         
       [{ a: 1 }, { b: 2 }],                                                                                                                                                                                            
       [{ a: 1 }],                                                                                                                                                                                                      
     );                                                                                                                                                                                                                 
     expect(result).to.deep.equal([{ b: 2 }]);                                                                                                                                                                          
   });   
```

Ran this test - it fails.

```
      AssertionError: expected [ { a: 1 }, { b: 2 } ] to deeply equal [ { b: 2 } ]
      + expected - actual

       [
         {
      -    "a": 1
      -  }
      -  {
           "b": 2
         }
       ]
```

packages/crypto-wallet-core/src/utils/index.ts
```diff
@@ -91,4 +91,55 @@ export function toHex(input: number | string | bigint): string {
...
+  // Use a WeakSet to track already compared objects to handle circular references
+  const weakRefs = new WeakSet();
+  const isWeak = (val) => val !== null && (typeof val === 'object' || typeof val === 'function');
 
+  while (stack.length) {
+    const a = stack.pop();
+    const b = stack.pop();
+    if (a === b) continue;
+    if (a == null || b == null) return false;
+    if (typeof a !== 'object' || typeof b !== 'object') return false;
```
MichaelAJay, line 129: Should there be a specific a/b are Arrays check - or is that handled correctly downstream?

packages/crypto-wallet-core/src/utils/index.ts
```diff
@@ -91,4 +91,40 @@ export function toHex(input: number | string | bigint): string {
...
+      if (!(b instanceof Date) || a.getTime() !== b.getTime()) return false;
+      continue;
+    }
+    if (a instanceof RegExp) {
+      if (!(b instanceof RegExp) || a.source !== b.source || a.flags !== b.flags) return false;
+      continue;
+    }
+    for (const key in a) {
+      stack.push(a[key], b[key]);
+    }
```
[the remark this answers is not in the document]
MichaelAJay: @kajoseph I think Copilot's on to something here.  Add this test and run it:
```typescript
    it('should compare repeated references against each matching path', function() {
      const shared = { value: 1 };
      const obj1 = {
        first: { value: 1 },
        second: { value: 2 }
      };
      const obj2 = {
        first: shared,
        second: shared
      };
      expect(utils.isEqual(obj2, obj1)).to.equal(false); // this will PASS
      expect(utils.isEqual(obj1, obj2)).to.equal(false); // this will FAIL
    });
```

The cycle guard tracks individual objects, but deep equality needs to track compared object pairs. A repeated reference on one side can cause a later path comparison to be skipped even when the object on the other side differs.

Recommended fix: replace the single WeakSet with pair tracking, e.g. WeakMap<object, WeakSet<object>>, and only skip when the exact (a, b) pair has already been compared.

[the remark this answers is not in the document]
MichaelAJay: Here's an implementation that the above test passes for:

```typescript
export function isEqual(obj1: object, obj2: object): boolean {
  if (obj1 === obj2) return true;
  if (obj1 == null || obj2 == null) return false;
  if (typeof obj1 !== 'object' || typeof obj2 !== 'object') return false;
  const stack = [obj1, obj2];
  // Track compared pairs to handle circular references without skipping different paths.
  const weakRefs = new WeakMap<object, WeakSet<object>>();

  while (stack.length) {
    const a = stack.pop();
    const b = stack.pop();
    if (a === b) continue;
    if (a == null || b == null) return false;
    if (typeof a !== 'object' || typeof b !== 'object') return false;
    if (Object.getPrototypeOf(a) !== Object.getPrototypeOf(b)) return false;
    const weakRefsForA = weakRefs.get(a);
    if (weakRefsForA?.has(b)) continue;
    if (weakRefsForA) {
      weakRefsForA.add(b);
    } else {
      weakRefs.set(a, new WeakSet([b]));
    }
    if (a instanceof Date) {
      if (!(b instanceof Date) || a.getTime() !== b.getTime()) return false;
      continue;
    }
    if (a instanceof RegExp) {
      if (!(b instanceof RegExp) || a.source !== b.source || a.flags !== b.flags) return false;
      continue;
    }
    for (const key of Object.keys(a)) {
      if (!(key in b)) return false;
      stack.push(a[key], b[key]);
    }
    const addtlBKeys = difference(Object.keys(b), Object.keys(a));
    for (const key of addtlBKeys) {
      if (!(key in a)) return false;
      stack.push(a[key], b[key]);
    }
  }
  return true;
}
```

packages/bitcore-cli/src/commands/create/index.ts
kajoseph: Nah, getChain() always returns lowercase. I'm adding jsdocs to prompts.ts so that's more explicitly understood.

packages/bitcore-cli/src/commands/join/index.ts
kajoseph: Same as above - getChain() always returns a lowercased string

packages/bitcore-cli/src/wallet.ts
```diff
@@ -132,22 +132,38 @@ export class Wallet implements IWallet {
...
+    let secret;
+    let joinedWalletName;
+    if (joinSecret) {
+      const wallet = await this.client.joinWallet(joinSecret, copayerName, { chain });
+      joinedWalletName = wallet.name;
+    } else {
+      secret = await this.register({ copayerName });
+    }
+    // Update credentials with joined/registered info (e.g. walletId, publicKeyRing, etc)
+    this.#walletData.credentials = credentials;
```
MichaelAJay: If the `if/else` block above doesn't mutate `this.#walletData.credentials`, then this line is a no-op.

If the `if/else` block DOES mutate `this.#walletData.credentials`, then I believe it's worse than a no-op, because we'd be overwriting the mutation with pre-mutated value.

packages/crypto-wallet-core/src/utils/index.ts
kajoseph: This is intended. It's meant as a replacement for lodash's `difference` function which does shallow equality

packages/crypto-wallet-core/src/utils/index.ts
kajoseph: It's handled

packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -134,7 +133,7 @@ export class Verifier {
       const o2 = args.outputs[i];
       if (!strEqual(o1.toAddress, o2.toAddress)) return false;
       if (!strEqual(o1.script, o2.script)) return false;
-      if (o1.amount != o2.amount) return false;
+      if (o1.amount != o2.amount && !(args.sendMax && o2.amount == null)) return false;
```
MichaelAJay, line 137: A comment here would be useful:
"If the amounts aren't equal, only 'max spend, amount omitted' is exempt"

packages/bitcore-cli/src/wallet.ts
MichaelAJay: @kajoseph I know you didn't change this - but Constants.PUBLIC_API has lowercase keys, and you're searching on uppercase keys here. Latent bug - only discovered because getTokenFromChain is called with a contract address not in the currencies endpoint.

MichaelAJay (approved):

nitsujlangston (approved):

PR #4138, opened 2026-04-06
[BWS, BWC] JIT Nonce XRP support

leolambo:
### Description
Extends the deferred nonce / `prepareTx` flow to XRP. Previously only EVM chains could use `deferNonce` to skip nonce assignment at create time and assign a fresh sequence number right before signing. XRP uses the same nonce-based signing model (the `Sequence` field is part of the signed payload), so it benefits from the same JIT assignment.

Also adds `deferNonce` to the BWC `Txp` interface so downstream consumers (like the app) can check it without type errors.

### Changelog

* `prepareTx` chain gate now allows XRP alongside EVM chains
* Null nonce guard in XRP `getBitcoreTx` to prevent NaN when building unsigned txs with deferred nonce (matches existing EVM pattern)
* Added `deferNonce?: boolean` to the BWC `Txp` type interface
* Integration tests for the XRP deferred nonce flow (create, prepareTx, gap-free nonce, sign)

### Testing Notes
Run BWS integration tests:
```bash
cd packages/bitcore-wallet-service && npm run test:integration
```
New tests are in the `#prepareTx XRP (deferred nonce)` describe block.

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-wallet-client/src/lib/api.ts
+++ b/packages/bitcore-wallet-client/src/lib/api.ts
@@ -4208,6 +4208,7 @@ export interface Txp {
   encryptedMessage?: string; // is set equal to `message` before decryption in processTxps()
   network: string;
   nonce?: number;
+  deferNonce?: boolean;
   note?: Note;
   outputOrder: Array<number>;
   outputs?: Array<{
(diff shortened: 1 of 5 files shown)


kajoseph (commented):
LGTM

kajoseph (approved):

Issue #3883, opened 2025-04-03
Refactor libs' BN and Point away from monkey patches


kajoseph:
The monkey patches and prototype overrides in the libs' `crypto/bn.js` and `crypto/point.js` cause issues when trying to use the underlying libraries elsewhere in projects.

These need to be refactored to at least use ES2015+ class-based inheritance as opposed to prototype overrides. Ideally, we could get rid of the dependence on BN.js and elliptic's Point altogether.

PR #4144, opened 2026-04-16
fix(bitcore-lib): normalize elliptic BN in ECDSA.sign

wqy000wqy:
## Summary

This fixes an assertion failure in `bitcore-lib` `ECDSA.sign()` that can happen in bundled environments where `bitcore-lib` and `elliptic` end up using different `bn.js` instances.

## Root cause

`ECDSA.sign()` computes:

```js
Q = G.mul(k);
r = Q.x.umod(N);
```

In some bundled runtimes:

- `Q.x` comes from `elliptic`
- `N` comes from bitcore's BN wrapper

Even though both are BN-like values, they are not guaranteed to be compatible for direct arithmetic across instances, which can trigger internal assertions.

## Fix

Normalize the elliptic coordinate through bytes before modular arithmetic:

```js
r = BN.fromBuffer(Buffer.from(Q.getX().toArray('be', 32))).umod(N);
```

`Q.x` is produced by `elliptic`, while the subsequent arithmetic uses bitcore's BN wrapper. Normalizing through bytes preserves the math while avoiding cross-instance BN operations.

## Validation

- added a regression test in `test/crypto/ecdsa.js`
- verified `npx mocha test/crypto/ecdsa.js`
- verified `npx mocha test/message.js`

### DIFF
--- a/packages/bitcore-lib/lib/crypto/ecdsa.js
+++ b/packages/bitcore-lib/lib/crypto/ecdsa.js
@@ -205,7 +205,10 @@ const sign = function(hashbuf, privkey, opts) {
     k = randomK ? getRandomK() : getDeterministicK(hashbuf, privkey, badrs);
     badrs++;
     Q = G.mul(k);
-    r = Q.x.umod(N);
+    // elliptic points may carry coordinates backed by a different bn.js instance
+    // than bitcore's BN wrapper when bundled by tools like webpack. Normalize
+    // through bytes before doing BN arithmetic to avoid cross-instance asserts.
+    r = BN.fromBuffer(Buffer.from(Q.getX().toArray('be', 32))).umod(N);
     s = k.invm(N).mul(e.add(d.mul(r))).umod(N);
   } while (r.cmp(BN.Zero) <= 0 || s.cmp(BN.Zero) <= 0);
 
@@ -299,4 +302,3 @@ module.exports.__testing__ = {
   getRandomK,
   toLowS,
 };
-
--- a/packages/bitcore-lib/lib/crypto/ecdsa.js
+++ b/packages/bitcore-lib/lib/crypto/ecdsa.js
@@ -205,7 +205,10 @@ const sign = function(hashbuf, privkey, opts) {
     k = randomK ? getRandomK() : getDeterministicK(hashbuf, privkey, badrs);
     badrs++;
     Q = G.mul(k);
-    r = Q.x.umod(N);
+    // elliptic point coordinates may come from a different bn.js instance in
+    // bundled runtimes. Point#getX() normalizes back into bitcore's BN type
+    // before modular arithmetic.
+    r = Q.getX().umod(N);
     s = k.invm(N).mul(e.add(d.mul(r))).umod(N);
   } while (r.cmp(BN.Zero) <= 0 || s.cmp(BN.Zero) <= 0);
 
@@ -299,4 +302,3 @@ module.exports.__testing__ = {
   getRandomK,
   toLowS,
 };
-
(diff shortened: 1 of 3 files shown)


wqy000wqy:
`ci/circleci: bitcore-lib` passed successfully.

The only failing check appears to be `ci/circleci: bitcore-node`, specifically an EVM integration test in `test/integration/routes/address.test.ts`:

- `Address Routes`
- `EVM`
- `should get token transactions`

The failure is `expected [] to deeply equal [...]`, which looks like the test did not receive the expected token transaction fixtures/data in CI.

Since this PR only changes `packages/bitcore-lib/lib/crypto/ecdsa.js` and its regression test, this failure appears unrelated to the `bitcore-lib` ECDSA change.

kajoseph:
@wqy000wqy Thanks for the PR. You'll need to sign your commits before this can be merged - see [CONTRIBUTING.md](https://github.com/bitpay/bitcore/blob/master/CONTRIBUTING.md)

wqy000wqy:
I checked the existing issues before opening this PR.

This appears to be related to #3883, which describes broader compatibility problems around bitcore's BN/Point abstractions and their interaction with underlying `bn.js` / `elliptic` types. However, that issue does not appear to include a fix yet for this specific `ECDSA.sign()` failure.

I reproduced the problem against the latest `master` before preparing this change.

This PR is intended as a minimal, focused fix for the concrete failure in `bitcore-lib` `ECDSA.sign()`, by normalizing the elliptic coordinate into bitcore's BN type before modular arithmetic.

I have also updated the PR with a signed commit, and the latest commit is now marked as `Verified`.

packages/bitcore-lib/lib/crypto/ecdsa.js
```diff
@@ -205,7 +205,10 @@ const sign = function(hashbuf, privkey, opts) {
     k = randomK ? getRandomK() : getDeterministicK(hashbuf, privkey, badrs);
     badrs++;
     Q = G.mul(k);
-    r = Q.x.umod(N);
+    // elliptic points may carry coordinates backed by a different bn.js instance
+    // than bitcore's BN wrapper when bundled by tools like webpack. Normalize
+    // through bytes before doing BN arithmetic to avoid cross-instance asserts.
+    r = BN.fromBuffer(Buffer.from(Q.getX().toArray('be', 32))).umod(N);
```
MichaelAJay: I think we can fix this more directly with `Q.getX().umod(N)`. In bitcore, Point.getX() already returns an internal BN instance, as does Point.getN(), so the stable boundary you're suggesting is maintained.

The only thing this is predicated upon is that getX() continues to return an instance of bitcore's BN. A test in `packages/bitcore-lib/test/crypto/point.js` would be good to ensure that invariant holds.

MichaelAJay (changes requested):
Nice catch on the cross-BN issue. I left one small comment on the implementation: I think we can use the existing `Point#getX()` normalization boundary and avoid the extra round trip.

packages/bitcore-lib/lib/crypto/ecdsa.js
wqy000wqy: Good point, thanks. I updated the implementation to use `Q.getX().umod(N)` and added a test in `test/crypto/point.js` to lock in the invariant that `Point#getX()` returns bitcore's BN type.

I re-ran:

- `npx mocha test/crypto/ecdsa.js`
- `npx mocha test/crypto/point.js`

MichaelAJay (approved):
Thanks for the PR - nice work!

kajoseph (approved):

catwalker777:
Dennis Kading
840-205-9394

PR #4146, opened 2026-04-17
Add copayers to tss txps

kajoseph:
### Description
Accepted/broadcasted tx proposals for TSS wallets currently only have 1 copayer. This PR adds the other TSS participants as copayers to the txp so the client can see who participated in the TSS signing session.

### Changelog

* On pushing the signature to the server, gets TSS signing session and attaches other participants as copayers to the txp.
* Easier local https with a CA

### Testing Notes
You should be able to sign a txp with a TSS wallet as usual and the txp should have all signers listed in the copayers array

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -1,4 +1,3 @@
-import util from 'util';
 import {
   BitcoreLib as Bitcore,
   BitcoreLibCash as BitcoreCash,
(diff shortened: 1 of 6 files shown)


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -3357,6 +3358,27 @@ export class WalletService implements IWalletService {
             return cb(ex);
           }
 
+          if (wallet.tssKeyId) {
+            try {
+              // Add the other copayers to the txp.copayers array
+              // so the client can see who participated in the signing.
+              const addrDbString = Utils.getAddressNetworkForDbLookup(txp.from || txp.inputs[0]?.address, wallet.network);
+              const address = await util.promisify(storage.fetchAddressByWalletId).call(storage, wallet.id, addrDbString);
+              const tssSigSeshId = `${txp.id}:${address.path.replace(/\//g, '-')}`;
```
gabrielbazan7: I'm generating the session ID on the client like this:
`${txp.id}:${derivationPath.replace(/\//g, '-')}-input${i}`
Because of that, the fetch always returns null.

Could you update it to:
```const tssSigSeshId = `${txp.id}:${address.path.replace(/\//g, '-')}-input1`;```


Using `-input1` is enough here, I tested it and it works correctly.

Comment (kajoseph) [reply packages/bitcore-wallet-service/src/lib/server.ts]: Is `i` starting at 1? Or would `-input0` be better?

Comment (kajoseph) [reply packages/bitcore-wallet-service/src/lib/server.ts]: Would it be too much to ask to change the format to be `<txpId>:<input>:<path>`? So then we'd use

```javascript
 const tssSigSeshId = `${txp.id}:input0:${address.path.replace(/\//g, '-')}`;
```

And I'm actually second guessing the need to even have the address path in the id. I think it would work to just have `<txpId:<input>` like `${txp.id}:input{i}`

gabrielbazan7: Yes, no problem. I can make that change. 

```
${txp.id}:input${i};
``` 

with `i` starting at 0. Thanks !

gabrielbazan7 (approved):
ACK!

nitsujlangston (approved):

PR #4150, opened 2026-04-29
Fix various small bugs

kajoseph:
### Description
This fixes various small bugs I found while digging into various issues recently

### Changelog

* bitcore-cli: Fixes read-only wallet imports
* bitcore-cli: Adds read-only export functionality (to support exporting RO wallets)
* root: Fixes vscode debugger crashing - bump root typescript version to 5.7
* insight: Fixes bitcore libs version mismatch with `@bitpay-labs` depdendencies
* insight: Fixes block page crashing to a white page when the block tx call fails

### Testing Notes
* bitcore-cli: Export any read-only wallet from the app and import into bitcore-cli
* bitcore-cli: Export a wallet as read-only
* bitcore-cli: Export a read-only wallet -- it should not error and not ask if you want to export read-only.
* root: The vscode debugger should work
* insight: it should run + going to block page where txs API call returns empty should not crash to a white page

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-cli/src/wallet.ts
+++ b/packages/bitcore-cli/src/wallet.ts
@@ -301,17 +301,18 @@ export class Wallet implements IWallet {
       throw new Error('No wallet data to save. Wallet not created or loaded');
     }
     
-    let key;
+    let key: TssKey.TssKey | KeyType | undefined;
     if (!readOnly) {
       if (this.#walletData.key instanceof TssKey.TssKey) {
         key = new TssKey.TssKey(this.#walletData.key.toObj());
       } else {
         key = new Key({ seedType: 'object', seedData: this.#walletData.key.toObj() });
       }
-    }
-    if (key && (key.isPrivKeyEncrypted() || key.isKeyChainEncrypted?.())) {
-      const walletPassword = await getPassword('Wallet password:');
-      key.decrypt(walletPassword);
+
+      if (key.isPrivKeyEncrypted() || (key as TssKey.TssKey).isKeyChainEncrypted?.()) {
+        const walletPassword = await getPassword('Wallet password:');
+        key.decrypt(walletPassword);
+      }
     }
     
     let data: any = { key: key?.toObj(), credentials: this.#walletData.credentials.toObj() };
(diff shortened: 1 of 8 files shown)


packages/bitcore-cli/src/wallet.ts
```diff
@@ -294,24 +294,25 @@ export class Wallet implements IWallet {
...
+    readOnly?: boolean;
   }) {
-    const { filename, exportPassword } = args;
+    const { filename, exportPassword, readOnly } = args;
     if (!this.#walletData) {
       throw new Error('No wallet data to save. Wallet not created or loaded');
     }
     
     let key;
     if (this.#walletData.key instanceof TssKey.TssKey) {
```
MichaelAJay: Wrap prior if/else in `if (!readonly) {}`

packages/bitcore-cli/src/wallet.ts
```diff
@@ -294,24 +294,27 @@ export class Wallet implements IWallet {
...
     exportPassword?: string;
+    readOnly?: boolean;
   }) {
-    const { filename, exportPassword } = args;
+    const { filename, exportPassword, readOnly } = args;
     if (!this.#walletData) {
       throw new Error('No wallet data to save. Wallet not created or loaded');
     }
     
     let key;
```
MichaelAJay: It would be nice to explicitly type key:
`let key: TssKey.TssKey | Key | undefined`

The downside being that the conditional below (get pw & decrypt) would have to get bumped into each conditional branch (but then could be cleaned up below)
```
    let key: TssKey.TssKey | Key | undefined;
    if (this.#walletData.key instanceof TssKey.TssKey) {
      key = new TssKey.TssKey(this.#walletData.key.toObj());
      const walletPassword = await getPassword('Wallet password:');
      key.decrypt(walletPassword);
    } else if (!readOnly) {
      key = new Key({ seedType: 'object', seedData: this.#walletData.key.toObj() });
      const walletPassword = await getPassword('Wallet password:');
      key.decrypt(walletPassword);
    }
```

MichaelAJay (approved):

nitsujlangston (approved):

PR #4151, opened 2026-04-30
Remove lodash from BWS express app route modules

tmcollins4:
### Description
Removes the remaining lodash usage from the BWS express app route modules. This keeps the extracted route code behavior-equivalent while reducing lodash dependency surface in the `expressapp.ts` route area.

### Changelog

* Replace lodash `each`, `isNumber`, `isString`, and `flatten` usage in BWS route modules with native equivalents.
* Remove lodash imports from the blockchain, maintenance, walletdata, and wallets route modules.

### Testing Notes

* Ran direct lodash-vs-native replacement checks.
* Ran BWS lint and production build.
* Ran focused ExpressApp route tests.

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/routes/maintenance.ts
+++ b/packages/bitcore-wallet-service/src/lib/routes/maintenance.ts
@@ -1,5 +1,4 @@
 import express from 'express';
-import * as _ from 'lodash';
 import { WalletService } from '../server';
 import { Stats } from '../stats';
 import type * as Types from '../../types/expressapp';
@@ -84,8 +83,9 @@ export function registerMaintenanceRoutes(router: express.Router, context: Route
   router.get('/v1/txnotes/', (req, res) => {
     getServerWithAuth(req, res, server => {
       const opts: { minTs?: number } = {};
-      if (req.query.minTs && _.isNumber(+req.query.minTs)) {
-        opts.minTs = +req.query.minTs;
+      const minTs = +req.query.minTs;
+      if (req.query.minTs && typeof minTs === 'number') {
+        opts.minTs = minTs;
       }
       server.getTxNotes(opts, (err, notes) => {
         if (err) return returnError(err, res, req);
--- a/packages/bitcore-wallet-service/src/lib/routes/maintenance.ts
+++ b/packages/bitcore-wallet-service/src/lib/routes/maintenance.ts
@@ -83,9 +83,12 @@ export function registerMaintenanceRoutes(router: express.Router, context: Route
   router.get('/v1/txnotes/', (req, res) => {
     getServerWithAuth(req, res, server => {
       const opts: { minTs?: number } = {};
-      const minTs = +req.query.minTs;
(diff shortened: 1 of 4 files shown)


kajoseph (dismissed):

packages/bitcore-wallet-service/src/lib/routes/wallets.ts
```diff
@@ -212,7 +211,7 @@ export function registerWalletRoutes(router: express.Router, context: RouteConte
     if (res.headersSent) {
       return;
     }
-    return res.json(_.flatten(responses));
+    return res.json(responses.reduce((all, response) => all.concat(response), []));
```
kajoseph: this reduce be simplified to just `responses.flat()`

packages/bitcore-wallet-service/src/lib/routes/maintenance.ts
```diff
@@ -84,8 +83,9 @@ export function registerMaintenanceRoutes(router: express.Router, context: Route
   router.get('/v1/txnotes/', (req, res) => {
     getServerWithAuth(req, res, server => {
       const opts: { minTs?: number } = {};
-      if (req.query.minTs && _.isNumber(+req.query.minTs)) {
-        opts.minTs = +req.query.minTs;
+      const minTs = +req.query.minTs;
+      if (req.query.minTs && typeof minTs === 'number') {
+        opts.minTs = minTs;
```
MichaelAJay: The unary coercion of req.query.minTs will always coerce to a number - but it may be a nonsensical number (such as NaN).

req.query.minTs = 'abc' is going to cause trouble downstream (it did previously too):
// +'abc' = NaN

```
const minTs = +'abc';
// minTs = NaN
// typeof minTs = 'number'
```

I think the cleanest replacement would be updating the conditional altogether:
`if(Number.isFinite(minTs)`

There are some other cases that should maybe be cleaned up while we're here:
`+[]` = 0;
`+{}` = NaN;
`+""` = 0;

(Same if we replace the unary coercion +x with Number(x))

To mitigate all of these issues, consider:
```
// if req.query.minTs isn't a non-empty string, then skip
if (typeof req.query.minTs === 'string' && req.query.minTs !== '') {
  const minTs = Number(req.query.minTs); // or +req.query.minTs;
  if (Number.isFinite(minTs) {
    opts.minTs = minTs;
  }
}
```

Not quite sure of the spirit of the request - but may be worth looking into.

kajoseph: I think `if (NaN` results in falsey, so the if statement check should handle the NaN case.

MichaelAJay: that's true, but the check is on
`if (req.query.minTs && typeof minTs === 'number')` - so a check on the request object directly.

It's not req.query.minTs that's NaN, it's minTs.

MichaelAJay (approved):

kajoseph (approved):

PR #4154, opened 2026-05-04
Fix bitcore-lib brittle point handling

MichaelAJay:
### Description
Updating packages for bitcore-lib caused tests to fail which exposed a brittle point handling in PublicKey._transformX() and Schnorr.verify().

Utilizing the internal BN module mitigates this issue.

### Changelog

* In PublicKey._transformX(), if x is a buffer, convert it to BN
* In Schnorr.verify(), convert public key to BN

#### PublicKey._transformX()

* Baseline: 244676.15 ns/op
* Inline fix: 249094.65 ns/op

#### Schnorr.verify()

* Baseline: 1820761.76 ns/op
* Inline fix: 1795178.86 ns/op

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-lib/lib/publickey.js
+++ b/packages/bitcore-lib/lib/publickey.js
@@ -1,13 +1,13 @@
 'use strict';
 
-var BN = require('./crypto/bn');
-var Point = require('./crypto/point');
-var Hash = require('./crypto/hash');
-var JSUtil = require('./util/js');
-var Network = require('./networks');
-var _ = require('lodash');
-var $ = require('./util/preconditions');
+const _ = require('lodash');
+const BN = require('./crypto/bn');
+const Hash = require('./crypto/hash');
+const Point = require('./crypto/point');
 const TaggedHash = require('./crypto/taggedhash');
+const Network = require('./networks');
+const JSUtil = require('./util/js');
+const $ = require('./util/preconditions');
 
 /**
  * Instantiate a PublicKey from a {@link PrivateKey}, {@link Point}, `string`, or `Buffer`.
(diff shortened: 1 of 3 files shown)


packages/bitcore-lib/lib/publickey.js
```diff
@@ -189,6 +189,9 @@ PublicKey._transformDER = function(buf, strict) {
  */
 PublicKey._transformX = function(odd, x) {
   $.checkArgument(typeof odd === 'boolean', 'Must specify whether y is odd or not (true or false)');
+  if (Buffer.isBuffer(x)) {
```
kajoseph: The JSDoc says x should be a `Point`, not a buffer. Where is x being passed in as a buffer?

I also see in the fn just above this that x is a `new BN(xbuf)` in several places. Would updating those to `BN.fromBuffer(xbuf)` fix this instead?

It seems like the JSDoc is wrong and x should be a `BN`, not a `Point`.

MichaelAJay: Interpreter.verifyTaprootCommitment() -> PublicKey.fromX(<bool>, x is buffer-life) -> PublicKey._transformX(..., x)

MichaelAJay: Updating in all PublicKey methods which call _transformX would be fine.

MichaelAJay: The JSDocs are indeed wrong - whether through drift or just generally wrong - no code path ends up calling _transformX with a Point.

The only callsites:
* _transformDER (you pointed this one out already)
* PublicKey.fromX (passthrough)

kajoseph (approved):
LGTM

PR #4155, opened 2026-05-04
[BCN] Streaming improvements

leolambo:
### Description
CSPs were carrying `req` and `res` through the whole call stack so that piping into the response could happen deep inside the provider. The HTTP framing moves back to the route layer via a new `streamJsonArray` helper, CSPs return `Readable` streams, and the three duplicate piping helpers (`ExternalApiStream.onStream`, `Storage.apiStream`, `Storage.stream`) are gone.

### Changelog

* Add `streamJsonArray` helper in `routes/apiUtils.ts`; supports JSON array (default) and JSONL via a `stream.jsonl` flag the helper auto-detects, so routes stay chain-agnostic
* Migrate the six stream entry points (`address`, `tx`, `block`, wallet `addresses`/`transactions`/`utxos`) plus the Gnosis multisig route across internal, base EVM, Moralis, MultiProvider, SVM, and Ripple to return streams
* Delete the legacy `onStream`/`apiStream`/`Storage.stream` helpers along with their `express` imports
* Restore eager cursor cleanup on client disconnect: the helper destroys the stream and settles the awaiting promise on req/res close, and EVM wallet-tx hooks cursor close to the final piped stream via a cleanup-callback list so the hook fires regardless of pipe-chain depth

### Testing Notes
* `npm run test:unit` in `bitcore-node` (281 passing; 8 new tests for `streamJsonArray` and EVM cursor-cleanup)
* Hit `/api/<chain>/<network>/address/:addr`, `/tx`, `/block`, and the wallet endpoints; output format matches the prior version
* ERC-20 path: `/address/:addr?tokenAddress=...` returns a JSON array of transfer objects (see Behavioral Notes)

### Behavioral notes
Two intentional output changes worth calling out for any byte-level consumer:
* The ERC-20 transfer endpoint now streams as a multi-line JSON array (`[\n{..},\n{..}\n]`) instead of `res.json()`'s compact form. Same JSON value, only inter-element whitespace differs.
* Empty-wallet 400 body changed from `text/plain "No addresses found for wallet"` to the JSON `{error, message}` shape used by every other 4xx path.

<hr style="border-width:3px">

### Checklist


packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
```diff
@@ -617,48 +586,37 @@ export class BaseSVMStateProvider extends InternalStateProvider implements IChai
...
-          logger.error('Error mid-stream (streamBlocks): %o', result.error?.log || result.error);
-        }  
-        return resolve();
-      } catch (err: any) {
-        logger.error('Error streaming blocks: %o', err.stack || err.message || err);
-        reject(err);
+        nextBlock = thisNextBlock;
+        block.nextBlockHash = nextBlock?.hash;
+        block.confirmations = Number(BigInt(height) - BigInt(block.height) + 1n);
+        stream.push(block);
```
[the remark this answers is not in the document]
leolambo: is pre-existing but drop the flag to be accurate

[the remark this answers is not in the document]
leolambo: intended, the jsonl endpoints emit newline-delimited objects, which isn't a valid JSON document, so the previous `application/json` label was inaccurate

packages/bitcore-node/src/routes/apiUtils.ts
```diff
@@ -10,3 +11,140 @@ export function respondWithError(res: Response, err: any) {
...
+      // instead of hanging until the stream eventually emits 'close' (which may not happen
+      // on a destroyed pipeline if upstream never settles).
+      safeResolve({ success: false, error: new Error('client disconnected') });
+    };
+
+    req.on('close', onAbort);
+    // ndjson in jsonl mode so JSON-aware clients (supertest, fetch().json()) don't try to
+    // parse a stream of newline-delimited objects as a single JSON document.
+    res.type(jsonl ? 'application/x-ndjson' : 'json');
+    res.on('close', onAbort);
```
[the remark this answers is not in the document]
leolambo: this is intentional and documented

kajoseph (commented):
There's one test failure that looks like it's relevant to this PR

packages/bitcore-node/src/routes/apiUtils.ts
```diff
@@ -10,3 +11,140 @@ export function respondWithError(res: Response, err: any) {
   }
   return res.status(500).send(err.message || err);
 }
+
+export interface StreamJsonArrayOpts {
+  jsonl?: boolean;
+}
+
+export interface StreamJsonArrayResult {
```
MichaelAJay: Consider a stricter union type here:
```typescript
export type StreamJsonArrayResult = 
  | { success: true }
  | { success: false; error: any; reason: 'client-disconnect' | 'late-error' | 'not-supported' | 'stream-error' | 'closed-before-end' }
```

The logging messaging for streamJsonArray() callers all references "Error mid-stream" - which could be sharpened up.

Even adding an `errMsg` property to the false case and explicitly setting it here could standardize the logging when required.

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
```diff
@@ -59,6 +60,10 @@ export interface BuildWalletTxsStreamParams {
   transactionStream: TransformWithEventPipe;
   populateEffects: PopulateEffectsForAddressTransform;
   walletAddresses: string[];
+  // _buildWalletTransactionsStream pushes teardown callbacks here (e.g. cursor.close).
+  // streamWalletTransactions runs them when the FINAL piped stream closes/ends, so the
+  // hook lives on the stream the route actually destroys on disconnect.
+  cleanups?: Array<() => void>;
```
MichaelAJay: Two issues with `cleanups`:
1) (minor) This interface should describe the inputs required to build the stream
2) `_buildWalletTransactionsStream` could just directly return a function which defines its disposal needs instead of pushing into an array defined in its caller (parent), who then also operates on that array.

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
MichaelAJay, line 1: Example of inverting the cleanup architecture:
```typescript
async _buildWalletTransactionsStream(...): Promise<{ stream: ..., dispose: () => void | Promise<void> }>

async streamWalletTransactions(...) {
  ...
  // Child directly defines its disposal function to caller instead of passing it in to parent's state
  const { stream: builtStream, dispose } = await this.__buildWalletTransactionsStream(...);

  // add transforms to builtStream
  const finalStream = builtStream.eventPipe(...)...;

  // events can occur in sequence. Guard before invoking dispose to ensure idempotency
  // perhaps this belongs in a helper file instead of in-line
  let disposed = false;
  const disposeOnce = () => {
    if (disposed) return;
    disposed = true;
    try {
       const result = dispose();
       if (typeof result?.catch === 'function') { // handle async cleanup failure }
    } catch (e) { // handle synchronous cleanup failures }
  }
  
  // Idempotent from the caller's perspective on the final stream
  finalStream.on('end', disposeOnce);
  finalStream.on('close', disposeOnce);
  finalStream.on('error', disposeOnce);
  ...
}
```

packages/bitcore-node/src/providers/chain-state/evm/api/csp.ts
```diff
@@ -731,22 +727,15 @@ export class BaseEVMStateProvider extends InternalStateProvider implements IChai
...
-      }
-    };
-
-    const { req, res } = params;
-    req.on('close', cleanupCursor);
-    res.on('close', cleanupCursor);
+    streamParams.cleanups?.push(() => {
+      if (cursorClosed) return;
+      cursorClosed = true;
+      try { cursor.close(); } catch { /* already closed */ }
```
MichaelAJay: cursor.close() returns a Promise - it needs to be awaited for the catch block to handle it.

packages/bitcore-node/src/routes/apiUtils.ts
```diff
@@ -10,3 +11,140 @@ export function respondWithError(res: Response, err: any) {
...
+          res.write('[\n');
+        } else {
+          res.write(',\n');
+        }
+      }
+      if (isFirst) {
+        isFirst = false;
+      }
+      if (typeof data !== 'string' && !Buffer.isBuffer(data)) {
+        data = JSON.stringify(data);
```
MichaelAJay: As of how this is called currently, this is safe. But for future-proofing it would be useful to go ahead and handle unexpected JSON.stringify errors here at the source. Also, if JSON.stringify does error, we probably don't want to toggle `isFirst` off, so move this above the `if (isFirst){...}` block.

packages/bitcore-node/src/routes/apiUtils.ts
```diff
@@ -45,6 +94,8 @@ export function streamJsonArray(
     // a client disconnect races a stream end/error or a stream 'close' event follows destroy().
     const safeResolve = (result: StreamJsonArrayResult) => { if (!settled) { settled = true; resolve(result); } };
     const safeReject = (err: any) => { if (!settled) { settled = true; reject(err); } };
+    const fail = (reason: StreamJsonArrayFailureReason, error?: any) =>
```
MichaelAJay, line 98: Just a stylistic point, but "unsuccessfulSafeResolve" would be a great indicator that this is just a convenience wrapper on safeResolve w/ a specified false result body.

Alternatively, it may overly obfuscate that we're actually just safeResolve'ing - the only difference is we'd build the StreamJsonArrayResult (just the extra success: false) instead of just passing reason and error. TS gives us good help if as soon as we put success: false in the object.

packages/bitcore-node/src/providers/chain-state/internal/internal.ts
```diff
@@ -430,8 +429,10 @@ export class InternalStateProvider implements IChainStateService {
       .find(query)
       .sort({ blockTimeNormalized: 1 })
       .addCursorFlag('noCursorTimeout', true);
-    const listTransactionsStream = new this.WalletStreamTransform(wallet);
-    transactionStream.pipe(listTransactionsStream).pipe(res);
+    const listTransactionsStream: any = transactionStream.pipe(new this.WalletStreamTransform(wallet));
```
MichaelAJay: transactionStream.close() returns a Promise, so we need to await here like we did elsewhere.

packages/bitcore-node/src/routes/apiUtils.ts
```diff
@@ -10,3 +13,222 @@ export function respondWithError(res: Response, err: any) {
...
+        try {
+          payload = jsonStringify(data);
+        } catch (err) {
+          return onStreamError(err);
+        }
+      }
+      if (!jsonl) {
+        res.write(isFirst ? '[\n' : ',\n');
+      }
+      isFirst = false;
```
MichaelAJay, line 202: Is backpressure important here? Opus recommended adding (after isFirst reassign)
```
      if (!res.write(payload) && !closed) {
        // The BTC wallet-tx and gnosis paths piped straight into res before this refactor, so a
        // slow client throttled the mongo cursor instead of buffering the response in memory.
        stream.pause();
        res.once('drain', () => { if (!closed) stream.resume(); });
      }
```

packages/bitcore-node/src/modules/multiProvider/api/csp.ts
MichaelAJay, line 252: Running some tests, caught an issue here.


> small, pre-existing but now deterministic: MultiProvider empty address result → 500 instead of []. In _buildAddressTransactionsStream, when _preflightStream resolves {success:true, firstItem:null} (empty result set), it returns txStream, which has already emitted 'end'. Confirmed on Node 22.16 that a Readable is destroyed===true after 'end', so streamJsonArray's guard rejects with stream destroyed before piping began → route 500. Master hung instead of 500, so it's not a regression.


Proposed fix:
```typescript
if (preflight.firstItem == null) {
  provider.health.recordSuccess();
  return Readable.from([], { objectMode: true });
}
```

MichaelAJay (approved):

kajoseph (approved):

PR #4156, opened 2026-05-05
Modules: loadModules and register functions instead of classes

MicahMaphet:
### Description
The `Modules` class logic was unnecessarily coupled. Made modules logic more isolated by replacing the `Modules` singleton with the `loadModules` function and removing convoluted inheritance. Modules no longer export classes, but instead export `register` functions.

The goal is entirely readability. The module classes had no need for internal state, all they do is call a few functions once. 

### Changelog

* Single `loadModules` function instead of `Modules` singleton
* Instead of the modules being classes, they export `register` functions
* Import `P2P`, `Libs`, `CSP`, and Verification from modules instead of passing from `Modules`
* Blew away `RippleEventAdapter`
* Blew away effectively unused interface `IService` and unused class `BaseModule`
* Updated tests accordingly
* `test/unit/modules.test.ts` `verifyModules` was redone to account for loosing `Modules` internal state

### Checklist


packages/bitcore-node/src/modules/bitcoin-cash/index.ts
```diff
@@ -1,15 +1,15 @@
...
 
-export default class BCHModule extends BaseModule {
-  constructor(services: BaseModule['bitcoreServices'], chain: string, network: string, _config: IUtxoNetworkConfig) {
-    super(services);
-    services.Libs.register(chain, '@bitpay-labs/bitcore-lib-cash', '@bitpay-labs/bitcore-p2p-cash');
-    services.P2P.register(chain, network, BitcoinP2PWorker);
-    services.CSP.registerService(chain, network, new BCHStateProvider());
-    services.Verification.register(chain, network, VerificationPeer);
-  }
+export default function register({ chain, network }: ChainNetwork) {
```
kajoseph: We should create a type for this register function.

`src/types/Module.ts`
```typescript
export type RegisterModule = (params: ChainNetwork) => void;
```

Then in here and other module files, import the type and 
```typescript
const register: RegisterModule = ({chain, network }) => {
  ...
};

export default register;
```

Then we can better declare the type for the `register` var in `src/modules/index.ts` > `loadModules()`
```typescript
  ...
  const register: RegisterModule = require(modulePath).default || require(modulePath);
  ...
```

kajoseph (approved):
Nice work!

PR #4158, opened 2026-05-06
[BCN] Add RPC overrides for MultiProvider EVM CSP + fixes

leolambo:
### Description

Multi-provider chains don't sync the local Mongo block and tx collections, so base CSP methods that read from them either return empty data or compute wrong values (`getLocalTip` returning null, `getFee` falling back to a 0 quartile, `/block` listing nothing). Mirror the override set already on `MoralisAdapter` so the multi-provider CSP serves these from the underlying RPC instead. Also fixes two bugs found while testing the wallet history endpoint.

### Changelog

* Override `getLocalTip` to fetch the latest block via realtime RPC, cached 5s per network to debounce hot route handlers
* Override `getFee` to use `rpc.estimateFee` instead of the empty `EVMTransactionStorage` gas-price history
* Override `streamBlocks` and `_getBlocks` to walk the resolved range via `web3.eth.getBlock` and serialize via `EVMBlockStorage.convertRawBlock`
* Use `transfer.rawContract.value` for asset-transfer amounts in `AlchemyAdapter._transformAssetTransfer`; the `transfer.value` field is a decimal display value and was leaking fractional, non-wei amounts into stored txs
* Route `_buildWalletTransactionsStream` to `streamERC20Transfers` when `args.tokenAddress` is set; previously it always called `streamAddressTransactions`, so wallet history for a specific token returned native transfers instead
* Cover the new overrides and bugfixes with unit tests in `test/unit/modules/base/csp.test.ts` and `test/unit/adapters/alchemy.test.ts`

### Testing Notes

* `/api/:chain/:network/block/tip` should return the live tip block, not null
* `/api/:chain/:network/fee/4` should return a non-zero feerate
* `/api/:chain/:network/block?limit=5` should list the last 5 blocks
* `/api/:chain/:network/wallet/:pubKey/transactions` (no `tokenAddress`) native transfers only, value in wei
* `/api/:chain/:network/wallet/:pubKey/transactions?tokenAddress=0x...`  only that token's transfers, value in token base units

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-node/src/modules/multiProvider/api/csp.ts
+++ b/packages/bitcore-node/src/modules/multiProvider/api/csp.ts
@@ -345,22 +345,33 @@ export class MultiProviderEVMStateProvider extends BaseEVMStateProvider {
     const { walletAddresses } = streamParams;
     const chainId = await this.getChainId({ network });
     const providers = this.getProvidersForNetwork(network);
+    const tokenAddress = (args as any)?.tokenAddress;
 
     let activeProvider = providers.find(p => p.health.isAvailable());
     if (!activeProvider) {
       throw new AllProvidersUnavailableError('walletTransactions', this.chain, network);
     }
-    logger.debug(`MultiProvider: wallet stream using ${activeProvider.adapter.name} for ${this.chain}:${network} (${walletAddresses.length} addresses)`);
+    logger.debug(`MultiProvider: wallet stream using ${activeProvider.adapter.name} for ${this.chain}:${network} (${walletAddresses.length} addresses)${tokenAddress ? ` token=${tokenAddress}` : ''}`);
 
     for (const address of walletAddresses) {
       try {
-        const txStream = activeProvider.adapter.streamAddressTransactions({
-          chainId,
-          chain: this.chain,
-          network,
-          address,
-          args: { order: 'ASC', ...args } as any
-        });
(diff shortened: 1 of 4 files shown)


kajoseph (commented):
I'm still getting ERC20 token transfer txs when querying for native transfer txs. Note, this is from the Alchemy provider.
```
◇  Tx History (page 1) ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                          │
│  [Wed May 06 2026 15:56:24 EDT] 0xcf4b76ea5f6c630f7739236c39d915d1bafa468f4882cc020b06d513daea82a2 <= received 0.0000000000012 ETH (0 fee) (4863 confs)  │
│  [Wed May 06 2026 14:18:00 EDT] 0x6bba71cbf51b20480cab4d459453d7adf226c430a8f3eb14ca4eb47cb738773b <= received 0.01 ETH (0 fee) (5317 confs)             │
│                                                                                                                                                          │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
For _sends_, want to include token transfers as 0 ETH transfers b/c we spent on gas, but for _receives_ we don't want them to show up in the native token tx history.

kajoseph (approved):
Approving and merging. The above comment can be addressed in a subsequent PR

PR #4160, opened 2026-05-07
Fix bitcore-lib browser test issues

MichaelAJay:
### Description
Locally, when Lodash was removed, it regenerated the package-lock file. Although the elliptic package was pinned, elliptic's dependencies were not pinned, so upon regeneration, elliptic's dependencies transitively caused test failure in browser test.

This was mitigated by updating bitcore-lib to use Point x-coordinate accessors (`point.getX()`) instead of directly accessing `point.x`. This issue was latent, and could have happened at any time, so this fix is appropriate.

In commit `f68b2ca86f4e520f2458ce4d9f96760ceb9fddb3`, benchmark scripts are checked in to the bitcore-lib package, along with a summary of execution (run on a Mac M3). These are then removed in the following commit.

### Changelog

* Use internal Point class `getX` accessor pattern instead of directly retrieving the point class's x coordinate.

### Testing Notes
-
<hr style="border-width:3px">

### Checklist


kajoseph (approved):
LGTM

PR #4161, opened 2026-05-08
Ethereum Stop Sequence

MicahMaphet:
### 🛑 Description
Previously when using ethereum on `bitcore-node`, the server would never stop. This was caused by the ethereum rpcs not stopping. This solves this by tearing down rpcs after the worker stops all the services

### 🪵 Changelog

* Tear down rpcs after the worker stops the services in all, api, and p2p

### 🧪 Testing Notes
Tested with a `bitcore.config.json` that only uses ethereum.
Run, and stop on `master` and `eth-stop`:
```
$ npm run p2p
[press ^C after ETH:regtest up to date message]
info :: 2026-05-08T06:27:30.011Z :: ETH:regtest up to date.
^Cinfo :: 2026-05-08T06:27:33.530Z :: Shutting down P2P pid 55016
```
<hr style="border-width:3px">

### Checklist


packages/bitcore-node/src/providers/chain-state/evm/p2p/p2p.ts
```diff
@@ -383,6 +383,7 @@ export class EVMP2pWorker extends BaseP2PWorker<IEVMBlock> {
     this.multiThreadSync.stop();
     logger.debug(`Stopping worker for chain ${this.chain} ${this.network}`);
     await this.disconnect();
+    setTimeout(BaseEVMStateProvider.teardownRpcs, 100);
```
kajoseph: Why the setTimeout?

MicahMaphet: It would crash the other services if done immediately.

kajoseph: Then it seems this is in the wrong place. We should call `teardownRpcs` after the other services are stopped.

kajoseph (commented):
Hmm... So, this works. _BUT_, importing `BaseEVMStateProvider` in these files means that EVM indexes will be created in every instance of bitcore-node, regardless of the chains configured. Consequently, if someone's running an instance of bitcore-node only for UTXO chains, the indexes [here](https://github.com/bitpay/bitcore/blob/337b27dacfbfeb83e282d5a172d892ea95aa9cae/packages/bitcore-node/src/providers/chain-state/evm/models/transaction.ts#L70) will still be created since `BaseEVMStateProvider` imports `packages/bitcore-node/src/providers/chain-state/evm/models/transaction.ts`.

I suppose `BaseEVMStateProvider` needs a registry of services that needs to be tracked when teardownRpcs is called.

PR #4170, opened 2026-05-15
[BWS/BWC] Feat: MoonPay embedded buy new endpoints

Gamboster:
Description
New endpoints are required to create/revoke a session in the Moonpay embedded flow. 

Changelog
I added two necessary endpoints: `createSession` and `revokeActiveSession`.
Additionally, there's a minor refactor for `getIpFromReq`.
The respective unit tests have been added.

### DIFF
--- a/packages/bitcore-wallet-service/src/externalservices/moonpay.ts
+++ b/packages/bitcore-wallet-service/src/externalservices/moonpay.ts
@@ -26,12 +26,14 @@ export class MoonpayService {
       SELL_WIDGET_API: string;
       API_KEY: string;
       SECRET_KEY: string;
+      SECRET_KEY_EMBEDDED: string | undefined;
     } = {
       API: config.moonpay[env].api,
       WIDGET_API: config.moonpay[env].widgetApi,
       SELL_WIDGET_API: config.moonpay[env].sellWidgetApi,
       API_KEY: config.moonpay[env].apiKey,
-      SECRET_KEY: config.moonpay[env].secretKey
+      SECRET_KEY: config.moonpay[env].secretKey,
+      SECRET_KEY_EMBEDDED: config.moonpay[env].secretKeyEmbedded
     };
 
     return keys;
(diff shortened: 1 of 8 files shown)


packages/bitcore-wallet-service/src/lib/common/utils.ts
```diff
@@ -282,13 +282,25 @@ export const Utils = {
...
+        ip = req.headers['x-real-ip'].split(',')[0];
+      }
     }
-    if (req.ip) return req.ip;
-    if (req.connection && req.connection.remoteAddress) return req.connection.remoteAddress;
-    return '';
+    if (!ip && req.ip) ip = req.ip;
+    if (!ip && req.connection && req.connection.remoteAddress) ip = req.connection.remoteAddress;
+
+    ip = ip.trim();
```
kajoseph: What if `!ip` is still true here?

Also, could you explain the need/use-case for this refactor?

Gamboster: `getIpFromReq` is a function I added for externalServices that require including the IP address in some requests.
Moonpay's `createSession` function for using Embedded buy requires the IP address, so I reviewed this function.
`trim()` is primarily for removing any leading, trailing, whitespace, or line terminator characters from the string.  I'll cover the `!ip` case.

I read that the IPv6-to-IPv4 normalizer could reduce potential future comparison errors by servers (since in some cases IPv4 users are masked as IPv6 by Node.js). However, I don't think it's strictly necessary. I'll remove it for now.

packages/bitcore-wallet-service/src/externalservices/moonpay.ts
```diff
@@ -435,4 +438,85 @@ export class MoonpayService {
...
+      }
+
+      const deviceIp = Utils.getIpFromReq(req);
+      if (!deviceIp) {
+        return reject(new ClientError('Could not determine device IP address'));
+      }
+
+      const headers = {
+        'Content-Type': 'application/json',
+        Authorization: "Api-Key " + SECRET_KEY,
```
kajoseph: lint

kajoseph (approved):

PR #4173, opened 2026-05-19
[BCN] Handle missing Solana token ATA in txhistory

msalcala11:
### Description
Fixes a BitPay app txhistory failure observed with zero-balance Solana token wallets.

Zero-balance Solana token wallets can still have transaction history. A user may have received or sent a token before, then sent the balance back to zero or closed the associated token account. In that state, the wallet should still be able to show historical activity.

**Observed failure mode:** When the app fetched txhistory for one of these token wallets, BWS returned a plain-text `Error getting ATA address` response instead of an empty or historical transaction list.

**Root cause:** BWS forwards `/v1/txhistory/` requests with `tokenAddress` to bitcore-node wallet transaction streaming. In the Solana token path, bitcore-node resolves the wallet address's associated token account before listing signatures. If there is no currently initialized ATA, `getConfirmedAta` throws and the stream fails.

**The fix:** This change keeps `getConfirmedAta` as the first lookup, so invalid mints and other non-missing-ATA failures keep their existing behavior. For the missing or uninitialized ATA case, it derives the deterministic ATA address and uses it to find any prior token-account activity. If prior activity exists, txhistory returns the corresponding transaction history even if the ATA was later closed.

### Changelog

* Fall back to the derived Solana ATA when SPL token txhistory has no currently initialized ATA.
* Preserve existing error behavior for invalid mints and other non-missing-ATA failures.
* Add regression coverage for `SOL + tokenAddress + missing current ATA`.

### Testing Notes

From `packages/bitcore-node`, run:

```bash
npm run tsc && BCN_LOG_LEVEL=none npm exec -- mocha 'build/test/integration/solana/csp.test.js' --grep 'derived ATA'
npm run test:integration
```

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
+++ b/packages/bitcore-node/src/providers/chain-state/svm/api/csp.ts
@@ -1,4 +1,5 @@
 import { CryptoRpc } from '@bitpay-labs/crypto-rpc';
+import { SOL_ERROR_MESSAGES } from '@bitpay-labs/crypto-rpc/lib/sol/error_messages';
 import { instructionKeys } from '@bitpay-labs/crypto-rpc/lib/sol/transaction-parser';
 import { fetchDigitalAsset, mplTokenMetadata } from '@metaplex-foundation/mpl-token-metadata';
 import { createUmi } from '@metaplex-foundation/umi-bundle-defaults';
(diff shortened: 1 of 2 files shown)


kajoseph (commented):
ACK

kajoseph (approved):

PR #4177, opened 2026-05-27
More bitcore-cli tweaks and improvements

kajoseph:
### Description
Fixes various bitcore-cli bugs and makes some flow improvements

### Changelog

* TSS Creation flow:
  * Adds helper instructions for the pub key/join code exchange
  * Replaces mnemonic backup with keyshare file backup
* Creation flow: skips address type prompt if only 1 address type available (e.g. DOGE)
* Creation flow: updated wallet password setting prompts to better reflect the `Unlock wallet` prompt. This is to assist user comprehension when exporting a wallet, which asks for an encryption password immediately followed by an unlock password (non read-only export).
* Improves text modding functions (colorText, boldText, etc.) to undo the mods rather than resetting to `\x1b[0m` (which also removes any previous text formatting)
* Fixed `opts.readonly` being false in exportWallet()
* Blocks `--register` from working on a TSS wallet since that'll register the wallet as just a single-sig wallet
* Fixes tests
* Updates the bin from `wallet` (very generic) to `bitcore-cli` (more specific)

### Testing Notes

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-cli/src/wallet.ts
+++ b/packages/bitcore-cli/src/wallet.ts
@@ -197,12 +197,20 @@ export class Wallet implements IWallet {
     if (!this.client) {
       await this.getClient({ mustExist: true });
     }
-    const { chain, coin, network, m, n, addressType } = this.client.credentials;
+    const { chain, coin, network, m, n, addressType, tssKeyId } = this.client.credentials;
     if (coin && coin !== chain) {
       // temporarily set it to chain for registration.
       // coin != chain for token wallets exported from the app.
       this.client.credentials.coin = chain;
     }
+
+    if (tssKeyId) {
+      // This function will only register the wallet as a single-sig (non-TSS) wallet (BAD!).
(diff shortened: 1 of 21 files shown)


packages/bitcore-cli/src/wallet.ts
```diff
@@ -333,7 +342,7 @@ export class Wallet implements IWallet {
       }
 
       if (key.isPrivKeyEncrypted() || (key as TssKey.TssKey).isKeyChainEncrypted?.()) {
-        const walletPassword = await getPassword('Wallet password:');
+        const walletPassword = await getPassword('Unlock wallet:');
```
MichaelAJay, line 345: Fine as is, but "Unlock wallet with wallet password", on the same terminal screen as "File encryption password", would be even more friendly.

MichaelAJay (approved):

nitsujlangston (approved):

PR #4184, opened 2026-06-09
Bitcore-cli XRP updates

kajoseph:
### Description
Bitcore-CLI is missing some important XRP functionality - namely, sending with a destination tag and setting account flags.

### Changelog

* Adds support for setting transaction flags for XRP accounts.
* Adds support for sending XRP with a dest tag.
* Upgrades @clack/prompts out of alpha. (Note, this replaced picocolors with fast-wrap-ansi for the terminal coloring which required a few test/CI config changes)

### Testing Notes
* Main menu should show a `Flags` advanced option for XRP wallets only
* `Flags` should show you current account flags
* `Flags` should allow you to choose account flags to set
* Choosing flags to set in `Flags` should flow cleanly into the tx creation flow (with account and amount pre set).
* The --help for the `flags` command should also show transaction options (e.g. `bin/wallet <name> --command flags --help`)
* `Send` should prompt you for an optional destination tag for XRP wallets only

<hr style="border-width:3px">

### Checklist


packages/crypto-wallet-core/src/transactions/xrp/index.ts
```diff
@@ -45,16 +58,14 @@ export class XRPTxProvider {
...
-          throw new Error('Invalid tfAccountSet flag');
-        }
         const accountSetTx: xrpl.AccountSet = {
           TransactionType: 'AccountSet',
           Account: from,
-          Flags: (isNaN(flags) ? xrpl.AccountSetTfFlags[flags] : flags) as number, // in testing, only the number values take effect.
+          Flags: this.transformFlags<xrpl.AccountSetFlagsInterface>(flags),
           Fee: fee.toString(),
           Sequence: nonce
         };
```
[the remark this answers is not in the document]
kajoseph: Hello??

packages/bitcore-wallet-service/src/lib/blockchainexplorers/v8.ts
```diff
@@ -43,6 +43,13 @@ function v8network(bwsNetwork, chain = 'btc') {
   return bwsNetwork;
 }
 
+function scrubResponseError(err) {
+  if (err.response) {
+    return new Error(`HTTP ${err.response.statusCode} - ${err.response.statusMessage}: ${err.response.body}\n${err.response.request.method} ${err.response.request.href}`);
+  }
+  return new Error(err);
```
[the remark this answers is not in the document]
kajoseph: This is by design since the HTTP errors inherit `Error`, so `err instanceof Error` is true and thus returning err doesn't scrub anything. Not sure what Copilot's saying about producing "[object Object]" as that has not been the case when I do `new Error(<err as an Error>)`.

The stack is replaced but it should only be one additional frame since err's stack would only be one frame up.

packages/crypto-wallet-core/src/utils/index.ts
```diff
@@ -159,4 +160,12 @@ export function isEqual(obj1: object, obj2: object): boolean {
...
   }
   return true;
+}
+
+export function normalizeXrpFlag(flag: string | number): string {
+  const normalizedFlag = isNaN(parseInt(flag as string)) ? (flag as string): xrpl.AccountSetTfFlags[flag as string];
+  if (!xrpl.AccountSetTfFlags[normalizedFlag]) {
+    throw new Error(`Invalid XRP flag: ${flag}`);
+  }
+  return normalizedFlag;
```
MichaelAJay: This normalizer is used by transformFlags, which is used to transform the xrpl.PaymentFlagsInterface case (payment/default in create). It will error if it receives one of the xrpl.PaymentFlagsInterface flags.

I think a payment test in the transactions.test.ts file (below) with string flags would demonstrate this error.

packages/bitcore-cli/src/commands/flags.ts
MichaelAJay, line 1: ## Issue
In direct `--command flags` flow: `getOrSetFlags()` calls `flags.command()`, which delegates to `transaction.command()` & registers/parses transaction options on the global Commander `program`. If `--flags` is provided, it then calls `createTransaction(args)` while `opts.command` is still truthy, so `createTransaction()` calls `transaction.command()` again and Commander attempts to register the same options twice.

## To observe
1) Create fresh XRP wallet & fund
2) Run with `--command flags --flags tfRequireDestTag`
3) Observe error: "Cannot add option..."

## Notes
The first parse in `getOrSetFlags()` is necessary to read in `--flags`. So `createTransaction()` should be flexible enough to not try to parse again.

MichaelAJay: Recommendation (confirmed it works) - look for a 'skipCommandParse' flag in the commands/transaction.ts createTransaction() method, and add that :
```
  if (opts.command && !opts.skipCommandParse) {
    Object.assign(opts, command(args));
  }
```

MichaelAJay:
Found a bug:

Environment: Local
Flow:
1) Send XRP (this is validation step). We're looking for typical "Send" behavior - namely that the first field is Recipient's Address.
2) Interact with the Flag menu - I observed the behavior both with toggling DestTag on, off, RequireAuth on, but NOT when choosing "None". Carry out tx.
3) Try to send XRP again. Note that the Recipeient's Address (and amount) are skipped - it goes straight to Destination Tag and doesn't come back around.

No matter what the flag state is, restarting the session fixes it.

packages/crypto-wallet-core/test/utils.test.ts
```diff
@@ -304,4 +305,88 @@ describe('Utils', function() {
       expect(utils.isEqual(left, right)).to.equal(false);
     });
   });
+
```
MichaelAJay, line 308: It would be useful to test valid PaymentFlags w/o enum - and for thoroughness, maybe consider the other cross case as well. In addition, what if 'normalizeXrpFlag' is called with flags from BOTH enums - it will throw for the first flag which isn't from the appropriate enum.
```
   it('should throw when a valid PaymentFlags string is passed without the enum', function() {                                                                                
     // tfPartialPayment is valid for PaymentFlags, but without the enum the function                                                                                         
     // falls back to AccountSetTfFlags, so this should throw                                                                                                                 
     expect(() => utils.normalizeXrpFlag('tfPartialPayment')).to.throw('Invalid XRP flag: tfPartialPayment');                                                                 
   }); 
```

Relatedly, it may be useful to refactor normalizeXrpFlag just a bit so that the error message includes a string representation of the enum set used to attempt the normalization.

That way the message isn't: "Invalid XRP flag: tfPartialPayment" (but tfPartialPayment IS valid a flag...)
It's: "Invalid XRP flag 'tfPartialPayment' for enum 'AccountSetTfFlags'"

MichaelAJay (approved):

nitsujlangston (approved):

PR #4190, opened 2026-07-06
[BWS] Feat: security updates for Transak

Gamboster:
Description
New security features were required by Transak:

```
1. CORS protection on your APIs
Restrict any of your APIs that call Transak to your own front-end domains, and block requests from any other origin. Never use a wildcard.
Access-Control-Allow-Origin: https://YOUR_DOMAIN.com

2. Forward the end-user IP header
Send the end user's originating IP on every call to Transak's APIs, matching the IP from which the user accesses the widget. If you are behind a CDN, use its client IP header (e.g. Cloudflare's cf-connecting-ip).
x-user-ip: your.users.ip

3. Send your partner API key in the header
Send your partner API key from your backend on every call to Transak's APIs. Your key is available in the Transak Partner Dashboard.
x-api-key: your-api-key-here
```

Changelog
In this pull request, `origin` checking via CORS was added, in addition to adding the `x-user-ip` and `x-api-key` headers to all requests to Transak.

### DIFF
--- a/packages/bitcore-wallet-service/bws.example.config.js
+++ b/packages/bitcore-wallet-service/bws.example.config.js
@@ -325,7 +325,7 @@ module.exports = {
   //     publicKey: 'simplex_web_public_key_here',
   //   }
   // },
-  // thorswap : {
+  // thorswap: {
   //   sandbox: {
   //     api: 'https://dev-api.thorswap.net',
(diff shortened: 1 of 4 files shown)


kajoseph (approved):
tACK

PR #4192, opened 2026-07-07
[BWS] Feat: security updates for Moonpay

Gamboster:
Description
New security features were required by Moonpay:

> We've introduced an allowedIpAddress parameter as part of your widget URL signing. When set, MoonPay validates that the user opening the widget is on the same IP that was signed, adding a layer of protection against session hijacking and unauthorized routing through intermediary servers.

Changelog
This PR includes the hashed user's IP address within the URL we generated for the Moonpay on-ramp checkout.


kajoseph (approved):
tACK

PR #4193, opened 2026-07-07
[BWS] Skip in-flight nonces when preparing deferred txps

leolambo:
### Description
Broadcasted-but-unmined txps weren't counted when `prepareTx` assigns a nonce: they're not in the confirmed tx count and no longer pending. So bulk signing deferred txps reused the same nonce until the previous tx mined.

### Changelog

* Scan broadcasted txps in `_prepareTx` and skip any nonce at or above the confirmed count
* Reject signing a deferred txp whose nonce matches a broadcasted txp. Non-deferred txps with an explicit nonce can still be signed, so speed-up replacements keep working
* Add integration tests for the broadcast window, the bulk sign flow, and the speed-up case

### Testing Notes

1. Create multiple EVM payouts. 
2. In the app, open the pending proposals for the ops wallet, select them all and hit Sign selected. 
3. Each proposal should broadcast with an incrementing nonce. 

Before this fix, every proposal after the first reused the same nonce unless the previous tx had already mined, so only one of them would confirm.

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/common/defaults.ts
+++ b/packages/bitcore-wallet-service/src/lib/common/defaults.ts
@@ -19,6 +19,9 @@ export const Defaults = {
   // TODO: should allow different gap sizes for external/internal chains
   SCAN_ADDRESS_GAP: 30,
 
+  // Lookback window (in seconds) for scanning broadcasted txps during nonce assignment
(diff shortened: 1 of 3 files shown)


kajoseph (approved):
Nice work

PR #4199, opened 2026-07-13
[BCN] fix gnosis token history casing gap and return 400 for invalid ethmultisig addresses

msalcala11:
### Description

Two fixes for Gnosis multisig token history:

1. The transaction query matches `effects.*` addresses case-sensitively against checksummed values, but effect addresses weren't checksummed at ingestion until 1df63364f, and `effects.*` was never backfilled (`fixEvmTxToFromAddressesCase.js` only rewrites top-level `to`/`from`). Docs synced before then store lowercase effect addresses, so the query silently drops them before the case-insensitive matching from [#4186](https://github.com/bitpay/bitcore/pull/4186#issuecomment-4712660189) can see them. The query now accepts both casings.
2. Since #4186, a malformed request address makes `toChecksumAddress` throw and the endpoint returns a 500 (previously an empty 200). It now returns a 400.

### Changelog

* Match both casings for `effects.contractAddress`, `effects.from`, and `effects.to` in the Gnosis transaction query via `$in` (still uses the existing indexes).
* Return a 400 for invalid addresses on `/ethmultisig/transactions/:multisigContractAddress`, validating with the same validation logic as `toChecksumAddress`.
* Update the Gnosis unit test query assertions to match.

### Testing Notes

* `cd packages/bitcore-node && npm run test:unit` — the "EVM token wallet history filtering" suite covers both the query shape and row emission. On Node below 22.12, run with `NODE_OPTIONS=--experimental-require-module`.
* `GET /api/ETH/mainnet/ethmultisig/transactions/notanaddress` returns 400 (previously 500); valid addresses in either casing stream normally.

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-node/src/providers/chain-state/evm/api/gnosis.ts
+++ b/packages/bitcore-node/src/providers/chain-state/evm/api/gnosis.ts
@@ -160,6 +160,10 @@ export class GnosisApi {
     const transactionQuery = getCSP(chain, network).getWalletTransactionQuery(params);
     delete transactionQuery.wallets;
     delete transactionQuery['wallets.0'];
+    // effects.* addresses were stored raw (lowercase) until 1df63364f and were
+    // never backfilled (fixEvmTxToFromAddressesCase.js only rewrites top-level
+    // to/from), so match either casing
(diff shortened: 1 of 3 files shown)


kajoseph (approved):
Nice work!

PR #4206, opened 2026-07-24
Handle tss txp deletion; Add cli session restoration; Bug fixes & improvements

kajoseph:
### Description
If walletA is in the process of signing a txp and walletB deletes the txp, then walletA will be spinning indefinitely. This PR deletes corresponding TSS signing sessions when deleting a txp. This will cause walletA's session to error with a "Session not found" error, leaving it up to the client to handle appropriately (as seen in CLI).

This PR adds support for TSS session restoration in case a session is interrupted.

This PR adds confirmation for password setting (would hate for someone to mistype their password only to find out later).

This PR makes the `.toObj` method make true copies of TssKey data. Without this, there was a race condition where exporting at the end of tss key creation may mean the unencrypted keyshare is saved to the wallet file. To that end, I also added a sanity check in the `save` method to ensure sensitive data does not get saved to disk.

This PR adds cleanup methods to the TSS keygen classes to zero out memory buffers that contain sensitive data.

This PR fixes a variety of CLI bugs.

### Changelog

* Deletes TSS signing sessions with txp removal.
* Adds TSS signing session restoration.
  * TODO: would be prudent to add a cleanup function to CLI to remove any zombie state files
* Adds confirmation prompts when setting passwords
* Adds cleanup methods & makes true copies of data for TssKey's `.toObj()`
* Adds prompt to encrypt wallet on load if not encrypted
* BUG - Fixes tokens not being saved to persistent opts in EVM wallets
* BUG - Fixes copayer0 of TSS wallets not getting other copayers. This would result in copayer0 being unable to verify txps created by other copayers.
* BUG - Sorts txps by nonce/creation date
* Increases PBKDF2 iterations

### Testing Notes

<hr style="border-width:3px">

### Checklist


packages/bitcore-cli/src/filestorage.ts
```diff
@@ -35,4 +35,12 @@ export class FileStorage {
   exists() {
     return fs.existsSync(this.filename);
   }
+
+  async getStatePath() {
```
MichaelAJay: A getter that mutates the filesystem is surprising. Consider renaming. Or at least providing a JSDoc specifying the recursive path creation.

packages/bitcore-cli/src/prompts.ts
```diff
@@ -75,26 +75,53 @@ export async function getPassword(
...
+        clearOnError: hidden
+      });
+      if (prompt.isCancel(password2)) {
+        throw new UserCancelled();
+      }
+      confirmed = password === password2; 
+    } else {
+      confirmed = true;
+    }
+  } while (!confirmed && !noRetry);
```
MichaelAJay: "while ... and not no retry" - consider cleaning up
```
noRetry -> retry
while (!confirmed && retry)
```

packages/bitcore-cli/src/prompts.ts
MichaelAJay, line 1: Thinking about the UX implications of the changes here.
In a web form, the "Confirm Password" field is just non-validated if it doesn't agree with the "Password" field. Obviously we can't do that for CLI operation - but is it cleanest to make the user retype in initial password on each pass?

Pass 1:
Password = "ABC"
Confirm = "ABV" (fat finger)

Pass 2:
Password = "ABC"...

Functionally fine, but personally I'd rather get a chance at retyping the Confirm, with an escape hatch - "Try new password".

packages/bitcore-cli/src/tss.ts
```diff
@@ -16,61 +18,84 @@ export async function sign(args: {
...
-  } catch (err) {
-    if (err.message?.startsWith('TSS_ROUND_ALREADY_DONE')) {
-      const sig = await tssSign.getSignatureFromServer();
-      if (!sig) {
-        throw new Error('It looks like the TSS signature session was interrupted. Try deleting this proposal and creating a new one.');
+  // Restore a previously-interrupted TSS session if it exists
+  if (fs.existsSync(storedSessionFile)) {
+    const storedSession = Encryption.decryptWithPassword(fs.readFileSync(storedSessionFile, 'utf8'), password);
+    await tssSign.restoreSession({ session: storedSession.toString(), password });
+
```
MichaelAJay, line 52: I think it would be nice to zero out storedSession at this point - if for no other reason than having that security consistency present in the codebase

packages/bitcore-cli/src/wallet.ts
```diff
@@ -326,9 +328,31 @@ export class Wallet implements IWallet {
 
     this.lockLoadedWallet();
     if (doNotComplete) return key;
-
+    
     const status = await this.client.openWallet();
     let needsSave = status?.wallet?.status === 'complete';
+
+    if (!key.isPrivKeyEncrypted()) {
+      if (loadingFromCache) {
```
MichaelAJay, line 336: Is it possible that it's loading from cache and is NOT a mistake? What happens if it's reloaded and has the same state?

packages/bitcore-cli/src/wallet.ts
```diff
@@ -361,6 +386,15 @@ export class Wallet implements IWallet {
 
         data = Encryption.encryptWithPassword(JSON.stringify(data), password, WALLET_ENCRYPTION_OPTS);
       }
+
```
MichaelAJay, line 383: This checks what it shouldn't have - should it also check what it SHOULD have?
```
data?.key?.xPrivKeyEncrypted
...etc
```

packages/bitcore-cli/src/wallet.ts
kajoseph: Yes, it's possible but not desirable - that's why I'm logging here. I'm not exactly sure _when_ this would happen, so I'm hoping to discover cases when/if it would happen and address those cases

packages/bitcore-cli/src/wallet.ts
kajoseph: No, since the absence of the xpriv/mnemonic is a valid state (readonly wallet)

MichaelAJay:
Observed a reproducible bug while functional testing.

```bash
◇  Enter the amount to send:
│  .00001
│
■  The number 1000.0000000000001 cannot be converted to a BigInt because it is not an integer
```

0.00001 * 1e8 evaluates to `1000.0000000000001` (as above).

Proposed change in `amountToSats`:\
```javascript
chain = chain.toLowerCase();
let multiplier = opts?.toSatoshis ?? 1e18; // Default to EVM chain
switch (chain) {
   case 'btc':
   case 'bch':
   case 'doge':
   case 'ltc':
     multiplier = 1e8;
     break;
   case 'xrp':
     multiplier = 1e6;
     break;
   case 'sol':
     multiplier = 1e9;
     break;
 }

return BigInt(Math.round(amount as number * multiplier));
```

packages/bitcore-wallet-client/src/lib/tsskey.ts
```diff
@@ -146,7 +161,9 @@ export class TssKey extends Key implements ITssKey {
     this.keychain.privateKeyShareEncrypted = JSON.stringify(Encryption.encryptWithPassword(this.keychain.privateKeyShare, password, opts));
     this.keychain.reducedPrivateKeyShareEncrypted = JSON.stringify(Encryption.encryptWithPassword(this.keychain.reducedPrivateKeyShare, password, opts));
     // remove the private data
+    this.keychain.privateKeyShare.fill(0);
```
kajoseph, line 162: I think I'd rather keep it as-is since your suggestion would leave a TssKey instance in a fractured state, complicating error handling.

packages/bitcore-wallet-client/src/lib/tsskey.ts
```diff
@@ -645,7 +666,7 @@ export class TssKeyGen extends EventEmitter {
         }
       });
       if (password) {
-        key.encrypt(password);
+        key.encrypt(password, { iter: 500_000 });
```
MichaelAJay: A regression test to ensure sufficient iterations would be good.

packages/bitcore-cli/src/wallet.ts
```diff
@@ -42,7 +43,7 @@ import type {
 const Client = API;
 
 const WALLET_ENCRYPTION_OPTS = {
-  iter: 5000
+  iter: 800_000
```
MichaelAJay, line 46: a regression test to ensure adequate iterations would be good

packages/bitcore-wallet-client/src/lib/tsskey.ts
kajoseph: I'm actually removing this. I added this in a moment when I thought 310k iterations was the OWASP recommendation, but I checked and it's 600k for SHA-256, so 500k is obviously not ideal.

packages/bitcore-wallet-client/src/lib/tsskey.ts
MichaelAJay: Misread the properties you were fill/unassigning - my bad.

packages/bitcore-wallet-client/src/lib/tsskey.ts
MichaelAJay: That seems good, although an integration test to ensure `encrypt` has an adequate number of iterations would still benefit overall security.

MichaelAJay (approved):

nitsujlangston (approved):

PR #4207, opened 2026-07-28
Add lavamoat allow-scripts

MichaelAJay:
### Description
@lavamoat/allow-scripts is added in an effort to mitigate supply chain attack vulnerabilities. All scripts are explicitly disallowed with the exception of bcrypt.

The top-level `compile` script is heavily modified: the original `lerna run compile` command allowed packages to compile in parallel based only on Lerna's declared dependency graph. That graph did not capture several build-time relationships created by generated TS output.

Of particular note, a build-time circular dependency between bitcore-node and bitcore-client was mitigated with the split-out `compile`. `bitcore-node` cannot fully compile until `bitcore-client` compiles. `bitcore-client` cannot compile until `bitcore-node` has generated its production build output. LavaMoat disabled the implicit npm `precompile` hook that previously generated `bitcore-node` production build before `bitcore-client` compiled.

#### DEVELOPMENT.md - Lavamoat lifecycle script behavior
   - When a managed package has any `lavamoat.allowScripts` entries,                                                      
     `allow-scripts run` will also execute that package's own                                                             
     `install`/`postinstall`/`prepublish`/`prepare` hooks unconditionally.                                                
   - The `allowScripts` policy only gates **dependency** lifecycle scripts.                                               
   - Adding a top-level lifecycle script to a package with an active                                                      
     `allowScripts` config will cause it to run during `npm run setup`.                                                   
                                                                                                                          
   ### Why this matters                                                                                                   
                                                                                                                          
   During clean-room validation of `npm run setup`, it was observed that                                                      
   `@bitpay-labs/bitcore-client` was the only workspace package triggering                                                
   dependency lifecycle scripts (specifically `bcrypt#5.1.0`). The LavaMoat                                               
   output made it appear as though the package's own lifecycle events were                                                
   running — but the distinction between dependency-gated and package-level                                               
   unconditional execution was unclear. This warning codifies the behavior                                                
   before someone adds an explicit lifecycle hook and is surprised by                                                     
   unintended execution during setup. It appears to be the case that if there are no dependencies explicitly allowed (`lavamoat.allowScripts`) then a package's lifecycle scripts are not run.

### Changelog

* Add .npmrc
* Install @lavamoat/allow-scripts
* Update all package.json files with a `lavamoat.allowScripts` configuration
* Only bcrypt is allowed - all other scripts are not required
* Split out top-level `compile` script.

### Testing Notes


<hr style="border-width:3px">

### Checklist


package.json
kajoseph: this does not run anymore when running `npm ci` from root. Is there a way to still allow this script to run?

MichaelAJay: This is intentional. We could run postinstall from the CircleCI build (I don't think we should) - but lavamoat's allow-scripts authored `.npmrc` specifically so we get full suppression of npm lifecycle scripts.

The reconfiguration of the CircleCI build config now addresses what `postinstall` was doing.

For the monorepo, we have to carry out the "Run approved dependency lifecycle scripts" from the CircleCI config between the `bootstrap` (which must set `npm_config_ignore_scripts=true`) and `compile` steps.

We could turn that into a script and just kick off `postinstall` directly from the CircleCI build (although I would press for naming it something that specifies its intentionality more) - but `postinstall` would need to become:
`npm run bootstrap (with scripts ignored) && npm run <run approved lifecycle scripts script> && npm run compile`

kajoseph: I'm not sure what you talking about with the CircleCI build. I'm talking about when I pull down a branch and run `npm ci` this PR makes it so it doesn't also run the postinstall script to bootstrap (i.e. symlink neighbor packages) and rebuild the packages. I understand that the point of lavamoat is suppress npm lifecycle scripts, but there are mechanisms for exceptions (e.g. like for bcrypt in this PR). Can we not add an exception for this particular script, too?

kajoseph: It seems like maybe we need another script anyway as the exceptions are never run. Don't we need to run `npx allow-scripts run` via lerna to really complete the npm install cycle? Something like

`package.json`
```json
"scripts": {
  "setup": "npm ci && lerna exec 'npx allow-scripts run' && npm run postinstall",
```

```sh
npm run setup
```

All this begs for a new `DEVELOPMENT.md` or `SETUP.md` doc to really explain what to do

kajoseph: I added the following scripts to my package.json

```json
"setup": "npm ci && npm run allow-scripts:run && npm run postinstall",
"allow-scripts:config": "lerna exec 'npx allow-scripts auto --skip-versions'",
"allow-scripts:run": "lerna exec 'npx allow-scripts run'",
```

Note the `--skip-versions` option on the config script. That'll open up the auto-script configs to be version agnostic.

Furthermore, running `npm run allow-scripts:run` as this branch currently sits produces some allow-script config discrepancies in BCN.

MichaelAJay: I made some changes.
1) There is now a `setup` script. I removed `postinstall` - its naming convention could lead to confusion, since the process we're trying to carry out specifically avoids reguarly-named lifecycle hooks.
2) I added `DEVELOPMENT.md`
3) As for the script changes - I think that disallowed scripts can skip versioning, but that allowed scripts should be version-controlled.

DEVELOPMENT.md
```diff
@@ -0,0 +1,163 @@
...
+When adding or updating a dependency:
+
+1. Update its manifest and lockfile.
+2. Install without lifecycle scripts:
+
+   ```sh
+   npm ci --ignore-scripts
+   npm run bootstrap:inert
+   ```
+
```
kajoseph: lockfiles are automatically generated/updated by npm. Does step 1 mean to run `npm i` from the package? And `npm ci` depends on the lockfile matching the manifest, so it depends on `npm i` having been run at some point before.

It might be worth having an `update` script so this can all be condensed to essentially:
1. Manually update/add the dependency in the relevant package.json
2. From root, run `npm run update`
3. Review every policy change...

DEVELOPMENT.md
```diff
@@ -0,0 +1,163 @@
...
+3. Populate new lifecycle paths as versionless denials:
+
+   ```sh
+   npm run allow-scripts:config
+   ```
+
+4. Review every policy change. Approve a script only when Bitcore requires its
+   generated artifact, and use an exact versioned `true` entry.
+5. Run `npm run allow-scripts:validate`.
+6. Run `npm run setup` from clean dependency trees and confirm Linux CI.
```
kajoseph: Why do we need to run this? And what does "confirm Linux CI" mean? How do I do that? Does that just mean "make sure the CI pipeline succeeds" (which would imply the user created a PR)?

kajoseph (commented):
I think the update dependency instructions should be simplified. Currently, this PR turns "update the dependency and run npm i" into a confusing 6 step process.

DEVELOPMENT.md
MichaelAJay: This is for the "Update dependencies" section - I guess it's not too clear, but assumedly if you're updating a dependency, downstream you'll be created a PR. If that's not clear or it's overly obvious I'll clean it up.

PR #4208, opened 2026-07-29
Proposal integrity and sendmax

tmcollins4:
### Description

Ensures transaction proposals preserve client-requested fixed fees, explicit UTXO inputs, and XRP destination tags. Also fixes send-max fee-level forwarding and DOGE input reporting.

Update the test UTXO fixture to use BWS’s canonical vout field for input outpoint verification.

### Changelog

- Verify fixed fees, explicit input outpoints/values, and XRP destination tags in BWC.
- Forward the requested fee level during send-max calculation.
- Return only DOGE inputs selected within the maximum transaction size.
- Update focused unit and integration coverage.

<hr style="border-width:3px">

### Checklist
- [x] I have read [CONTRIBUTING.md](https://github.com/bitpay/bitcore/tree/master/CONTRIBUTING.md) and verified that this PR follows the guidelines and requirements outlined in it.

### DIFF
--- a/packages/bitcore-wallet-client/src/lib/verifier.ts
+++ b/packages/bitcore-wallet-client/src/lib/verifier.ts
@@ -23,9 +23,13 @@ export class Verifier {
   private static normalizeAtomicValue(value) {
     if (typeof value === 'bigint') return value >= 0n ? value : null;
     if (typeof value === 'number') {
-      return Number.isSafeInteger(value) && value >= 0 ? BigInt(value) : null;
+      // Preserve the integer value actually represented by JavaScript, including unsafe Numbers.
+      return Number.isInteger(value) && value >= 0 ? BigInt(value) : null;
     }
-    if (typeof value === 'string' && /^(0|[1-9]\d*)$/.test(value)) {
+    if (
+      typeof value === 'string' &&
+      (/^(0|[1-9]\d*)$/.test(value) || /^0x[0-9a-fA-F]+$/.test(value))
+    ) {
       return BigInt(value);
     }
     return null;
@@ -40,8 +44,8 @@ export class Verifier {
   }
 
   private static optionalAtomicValuesEqual(value1, value2) {
-    const value1Missing = value1 === undefined || value1 === null;
-    const value2Missing = value2 === undefined || value2 === null;
+    const value1Missing = value1 == null;
+    const value2Missing = value2 == null;
     if (value1Missing || value2Missing) return value1Missing && value2Missing;
     return this.atomicValuesEqual(value1, value2);
   }
--- a/packages/bitcore-wallet-client/src/lib/verifier.ts
+++ b/packages/bitcore-wallet-client/src/lib/verifier.ts
@@ -212,9 +212,9 @@ export class Verifier {
       return false;
     if (typeof args.feePerKb === 'number' && txp.feePerKb != args.feePerKb)
       return false;
-    if (args.fee !== undefined && !this.atomicValuesEqual(args.fee, txp.fee))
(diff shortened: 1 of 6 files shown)


packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -20,6 +20,67 @@ const BCHAddress = BitcoreLibCash.Address;
 export class Verifier {
   private static _useRegtest: boolean = false;
 
+  private static normalizeAtomicValue(value) {
+    if (typeof value === 'bigint') return value >= 0n ? value : null;
+    if (typeof value === 'number') {
+      return Number.isSafeInteger(value) && value >= 0 ? BigInt(value) : null;
+    }
+    if (typeof value === 'string' && /^(0|[1-9]\d*)$/.test(value)) {
```
kajoseph: What about hex strings (assume 0x prefix)?

packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -20,6 +20,67 @@ const BCHAddress = BitcoreLibCash.Address;
 export class Verifier {
   private static _useRegtest: boolean = false;
 
+  private static normalizeAtomicValue(value) {
+    if (typeof value === 'bigint') return value >= 0n ? value : null;
+    if (typeof value === 'number') {
+      return Number.isSafeInteger(value) && value >= 0 ? BigInt(value) : null;
```
kajoseph: return null if is a very large, unsafe number? what are the implications of that?

packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -20,6 +20,67 @@ const BCHAddress = BitcoreLibCash.Address;
...
+  private static atomicValuesEqual(value1, value2) {
+    const normalizedValue1 = this.normalizeAtomicValue(value1);
+    const normalizedValue2 = this.normalizeAtomicValue(value2);
+    return normalizedValue1 !== null &&
+      normalizedValue2 !== null &&
+      normalizedValue1 === normalizedValue2;
+  }
+
+  private static optionalAtomicValuesEqual(value1, value2) {
+    const value1Missing = value1 === undefined || value1 === null;
```
kajoseph: can be shortened to `const value1Missing = value1 == null;`

kajoseph (commented):
I have some questions about the handling of large, unsafe numbers (applicable with EVM)

packages/bitcore-wallet-client/src/lib/verifier.ts
tmcollins4: Good catch. I will add in support for that.

packages/bitcore-wallet-client/src/lib/verifier.ts
tmcollins4: My thought was this: An unsafe JavaScript number may already be rounded before the verifier receives it, so converting it to BigInt could preserve the wrong value. Returning null makes verification fail safely. Large values would need to be sent as a bigint or decimal/hex string to be compared exactly.

kajoseph: When creating a txp, if I give a `fee` of any unsafe integer (e.g. 12345678901234567), this function will automatically return null which means the upstream `checkProposalCreation` will automatically fail (even if the unsafe integers happen to equate).

In practice, I'm not sure how common this actually is to pass in a specific `fee` (vs `feeLevel` or `feePerKb`) and also as a Number. And this probably would only happen with EVM (which uses much larger numbers than other chains)

kajoseph (approved):
Nice work!

PR #4218, opened 2026-08-12
Add TSS versioning & expiration

kajoseph:
### Description
Adds versioning and expiration to TSS sessions to ensure a) there's some finality to sessions, and b) there's clarity around upgrades

### Changelog

* Adds versioning to TSS
* Adds expiration to TSS sessions

### Testing Notes
Tests have been added for both cases

<hr style="border-width:3px">

### Checklist


packages/bitcore-wallet-service/src/lib/model/tsskeygen.ts
```diff
@@ -103,14 +103,34 @@ export class TssKeyGenModel implements ITssKeyGenModel {
...
+    /** TSS keygen version */
+    version: number;
+    /** Time limit for the session in minutes, after which the session expires */
+    timeLimit?: number;
+  }): TssKeyGenModel {
+    const { id, message, n, copayerId, passwordHash, version } = params;
     const { partyId } = message;
+    const timeLimit = Number(params.timeLimit);
     $.checkArgument(partyId === 0, 'Key generation session must be started by partyId 0');
+    $.checkArgument(!timeLimit || (Number.isFinite(timeLimit) && timeLimit > 0), 'Time limit must be a positive number');
```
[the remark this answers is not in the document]
kajoseph: this is fine as `0` is a common value for turning "off" the limit. In this case it will fallback to the max time limit of 60 minutes

packages/bitcore-wallet-service/src/lib/model/tsssign.ts
```diff
@@ -90,16 +101,24 @@ export class TssSigGenModel implements ITssSigGenModel {
...
 
 
-  static create(params: { id: string; message: ITssSigMessageObject; m: number; copayerId: string }): TssSigGenModel {
-    const { id, message, m, copayerId } = params;
+  static create(params: { id: string; message: ITssSigMessageObject; m: number; copayerId: string; version: number; timeLimit?: number }): TssSigGenModel {
+    const { id, message, m, copayerId, version } = params;
     const { partyId } = message;
+    const timeLimit = Number(params.timeLimit);
+    $.checkArgument(!timeLimit || (Number.isFinite(timeLimit) && timeLimit > 0), 'Time limit must be a positive number');
+
```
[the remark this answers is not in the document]
kajoseph: I've adjusted this so that "0" falls back to the MAX_TIME_LIMIT, which is the common usage for passing 0 to a limit.

packages/bitcore-wallet-service/src/lib/model/tsskeygen.ts
```diff
@@ -151,4 +173,11 @@ export class TssKeyGenModel implements ITssKeyGenModel {
     }
     return mostRecentRound;
   }
+
+  isExpired(): boolean {
+    if (!this.timeLimit) {
+      return false;
```
MichaelAJay, line 179: Should we fail-close here instead?

kajoseph: I think it's fine as-is. There may be a use case for having an unlimited session. We can always shore this up later by falling back to a constant, if we want.

packages/bitcore-wallet-service/src/lib/common/defaults.ts
```diff
@@ -434,6 +434,15 @@ export const Defaults = {
...
 
-  TSS_KEYGEN_SCHEME_VERSION: 1,
-  TSS_SIGGEN_SCHEME_VERSION: 1,
+  // In general, the major version represents server upgrades or major security upgrades, and the minor version represents client upgrades.
+  // Example1: If the server has an API response overhaul, that's a breaking change for older clients - it should increment the major version.
+  //           This will result in an UPGRADE_NEEDED error for older clients.
+  // Example2: If the client upgrades the encryption scheme, older clients will still be compatible with the server and
+  //           other older clients - it should increment the minor version. Mismatching client versions will still break the process, but
+  //           the difference from a major version bump is that matching older clients will still work.
+  TSS_KEYGEN_SCHEME_VERSION: 1.1,
```
MichaelAJay, line 443: Would it be useful to point to the MAX values here (from constants.ts) - do you ever see a case where the defaults wouldn't be the maxes?

packages/bitcore-wallet-service/src/lib/model/tsssign.ts
```diff
@@ -133,4 +157,11 @@ export class TssSigGenModel implements ITssSigGenModel {
     }
     return mostRecentRound;
   }
+
+  isExpired(): boolean {
+    if (!this.timeLimit) {
+      return false;
```
MichaelAJay, line 163: should we fail closed?

packages/bitcore-wallet-service/src/lib/routes/middleware/authTssRequest.ts
MichaelAJay, line 47: What would happen if someone accidentally called authTssRequest from a path that doesn't match one of the partial paths not listed above?

The thrown error would be `Session not found`, but really it was an invalid call if the method should only handle routes including `/tss/keygen` or `/tss/sign/`. The error would be misleading.

An `else` clause above could sharpen that up: throw error with a message indicating invalid usage of the method. Then we'd know for certain if we saw a "Session not found" error that the session was actually looked for.

packages/bitcore-wallet-service/src/lib/common/constants.ts
```diff
@@ -226,6 +226,9 @@ export const Constants = {
 
   // May be different from the defaults if we rollback
   // or want to beta test new versions.
-  TSS_KEYGEN_SCHEME_VERSION_MAX: 1,
-  TSS_SIGGEN_SCHEME_VERSION_MAX: 1,
+  TSS_KEYGEN_SCHEME_VERSION_MAX: 1.1,
+  TSS_SIGGEN_SCHEME_VERSION_MAX: 1.1,
+  
+  TSS_KEYGEN_SCHEME_MIN_SERVER_VERSION: 1,
+  TSS_SIGGEN_SCHEME_MIN_SERVER_VERSION: 1,
```
MichaelAJay: These are exact pairs - right? MIN/MAX? Recommend naming consistency - tss.ts processMessage (below) illustrates the inconsistency nicely.

packages/bitcore-wallet-service/src/lib/model/tsssign.ts
kajoseph: Same as above

packages/bitcore-wallet-service/src/lib/common/defaults.ts
kajoseph: Yes, in the case we need to rollback from a bad version or want to test a new version before setting to default

packages/bitcore-wallet-service/src/lib/routes/middleware/authTssRequest.ts
```diff
@@ -40,6 +40,8 @@ export function authTssRequest(): express.RequestHandler {
         session = await storage.fetchTssSigSession({ id });
         partyId = session?.participants.find(p => p.copayerId === copayerId)?.partyId;
         pubKey = partyId == null ? null : session?.rounds[0]?.find(r => r.fromPartyId === partyId)?.messages.publicKey;
+      } else {
+        throw Errors.NOT_AUTHORIZED.withMessage('Invalid TSS request path');
```
MichaelAJay: Not a critical issue, but this wouldn't be a client-side error - it would be a developer error: server-side misconfiguration.

packages/bitcore-cli/src/tss.ts
```diff
@@ -89,6 +92,10 @@ export async function sign(args: {
         tssSign.unsubscribe({ clearEvents: true });
         spinner.cancel('TSS session not found. It may have been deleted by another party.');
         return reject(new ProcessCancelled());
+      } else if (e instanceof Errors.TSS_SESSION_EXPIRED) {
+        tssSign.unsubscribe({ clearEvents: true });
+        spinner.cancel(e.message);
+        return reject(new ProcessCancelled());
```
MichaelAJay, line 105: In both of these terminal cases, consider cleaning up `storedSessionFile`, which is only cleaned up in the `complete` handler currently.

MichaelAJay (approved):

nitsujlangston (approved):

PR #4224, opened 2026-08-20
Fix unauthorized wallet join via alternate public-key fields

MichaelAJay:
### Description

Closes IS-1391 — unauthorized wallet join via `hardwareSourcePublicKey` / `clientDerivedPublicKey`.

`WalletService.joinWallet()` (`POST /v2/wallets/:id/copayers/`) previously returned early into `_addCopayerToWallet()` when a join request included a truthy `hardwareSourcePublicKey` or `clientDerivedPublicKey` and no request-side `opts.tssKeyId`. Normal join requests do not carry `opts.tssKeyId`, so this branch ran before wallet-secret verification (`copayerSignature`) and bypassed the upgrade, chain, network, legacy-wallet, duplicate-copayer, wallet-full, and TSS-participant checks.

Anyone who knew the ID of a wallet with an available join slot could therefore add an attacker-controlled copayer and obtain a valid copayer ID and authenticated request key without proving knowledge of the wallet’s join secret.

This PR removes that early return. Every accepted join now requires a valid HD `xPubKey`, which is the canonical identity input for the service-layer join protocol. `hardwareSourcePublicKey` and `clientDerivedPublicKey` remain available as ancillary copayer data, but their presence cannot suppress `xPubKey` validation or bypass join authorization.

The PR also makes `addAccess()` fail closed when it encounters malformed historical copayer state.

#### Compatibility note (intentional protocol restriction)

Service-layer joins that provide only an alternate public key, or provide a raw, placeholder, or otherwise invalid HD value in `xPubKey`, are now rejected.

The repository’s generated-client join path (`_doJoinWallet`) supplies both a real `xPubKey` and a real `copayerSignature`, including when either ancillary field is present. No generated-client flow exercised by this repository depends on the removed shortcut.

Upstream or out-of-repository consumers that relied on xpub-less joins must be inventoried before this change is considered impact-free. Such clients should not be accommodated by restoring an authorization bypass; supporting a different canonical identity requires a versioned end-to-end protocol.

### Changelog

* Removed the authorization-free early return from `joinWallet()`. Alternate public-key fields no longer call `_addCopayerToWallet()` before authorization and wallet validation.
* Required and parsed `xPubKey` for every service-layer join. Ancillary fields can no longer suppress required-field, HD parsing, or network validation.
* Ensured joins carrying ancillary fields are subject to the normal upgrade, chain, network, legacy-wallet, wallet-secret, and duplicate-copayer checks. Non-TSS joins remain subject to the wallet-full check, while TSS joins must pass the TSS participant check.
* Ensured the existing TSS participant check is reached after authorization. That check uses the persisted `wallet.tssKeyId` and the corresponding key-generation session’s participant list; the removed shortcut was incorrectly conditioned on request-side `opts.tssKeyId`, which normal join requests do not carry.
* Eliminated two historical failure modes for an absent join-time `xPubKey`: an undefined-input hashing exception on the default chain and a colliding constant derived ID on non-default chains.
* Made `addAccess()` return `NOT_AUTHORIZED` when the target copayer is missing or its persisted `xPubKey` is absent or malformed, instead of allowing a property-access or HD-parsing exception to escape.
* Added regression coverage in `test/integration/server.test.ts` and `test/model/copayer.test.ts`.

### Testing Notes

Run from `packages/bitcore-wallet-service`:

```bash
npm test

# Or target the affected suites:
npm run compile && npx mocha 'ts_build/test/integration/server.test.js' 'ts_build/test/model/copayer.test.js'
```

New and relevant assertions reviewers should validate
- Wrong-signature joins with either ancillary field are rejected as Bad request, and the persisted wallet remains unchanged with no copayer added.
  Tests in test/integration/server.test.ts:
  - should fail to join with wrong signature even when hardwareSourcePublicKey is present
  - should fail to join with wrong signature even when clientDerivedPublicKey is present
- Valid-signature joins with ancillary data succeed, return the persisted copayer ID, and retain the supplied metadata.
  Tests in test/integration/server.test.ts:
  - should join existing wallet with a valid signature and hardwareSourcePublicKey present
  - should join existing wallet with a valid signature and clientDerivedPublicKey present
- Missing join-time xPubKey values are rejected even when an ancillary field is present, and no copayer is persisted.
  Tests in test/integration/server.test.ts:
  - should reject a join with hardwareSourcePublicKey present when xPubKey is absent
  - should reject a join when xPubKey, hardwareSourcePublicKey and clientDerivedPublicKey are all missing or empty
  - should reject a TSS join with clientDerivedPublicKey present when xPubKey is absent, even if the ancillary-derived ID is a keygen session participant
- Malformed join-time xPubKey values are rejected regardless of ancillary metadata, with no copayer persisted in the new ancillary-field cases.
  Tests in test/integration/server.test.ts:
  - should fail to join with invalid xPubKey
  - should fail to join with a malformed xPubKey even when hardwareSourcePublicKey is present
  - should fail to join with a malformed xPubKey even when clientDerivedPublicKey is present
- A join carrying ancillary metadata cannot bypass the wallet-full check. A third join into a full m=1, n=2 wallet returns WALLET_FULL, and the wallet remains at two copayers.
  Test in test/integration/server.test.ts:
  - should fail to join a full wallet with a valid signature and hardwareSourcePublicKey present
- A TSS joiner with a valid copayerSignature is rejected with TSS_NON_PARTICIPANT when its xpub-derived copayer ID is absent from the key-generation session, and no copayer is persisted.
  Test in test/integration/server.test.ts:
  - should reject a join from a copayer who is not a TSS keygen session participant, even with clientDerivedPublicKey supplied
- addAccess() returns NOT_AUTHORIZED and leaves wallet and copayer-lookup state unchanged when historical copayer state is missing or malformed.
  Tests under #add access with historical copayer state in test/integration/server.test.ts:
  - should return NOT_AUTHORIZED when the target copayer is missing from the wallet
  - should return NOT_AUTHORIZED when the target copayer has no xPubKey
  - should return NOT_AUTHORIZED when the target copayer has a malformed xPubKey
- When an xPubKey is supplied, the copayer ID is derived from it regardless of ancillary fields. Supplied ancillary fields are persisted without selecting or changing identity.
  Tests under #create identity in test/model/copayer.test.ts:
  - should derive the ID from xPubKey alone
  - should prefer xPubKey when hardwareSourcePublicKey is also present
  - should prefer xPubKey when both alternate credentials are also present
  - should derive the same xpub-based ID on a non-default chain
  - should persist the supplied ancillary fields and leave unsupplied ones absent
  - should persist both ancillary fields without changing the xpub-derived ID
- Existing default-chain model behavior rejects creation when neither xPubKey nor either ancillary field is supplied.
  Test in test/model/copayer.test.ts:
  - should reject creation without xPubKey and without ancillary fields

Control-flow points not covered by new alternate-field-specific regression tests should be reviewed directly: chain mismatch, xpub network mismatch, duplicate-copayer detection, upgrade gates, and legacy-wallet rejection. The removed early return previously bypassed all of these checks.

The TSS tests explicitly clear their `tss_keygen` fixture IDs because `helpers.beforeEach()` does not currently include `tss_keygen` in its collection wipe list. This keeps the tests repeatable despite that pre-existing cleanup gap.

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -1144,9 +1144,15 @@ export class WalletService implements IWalletService {
         if (err) return cb(err);
         if (!wallet) return cb(Errors.NOT_AUTHORIZED);
 
-        const xPubKey = wallet.copayers.find(c => c.id === opts.copayerId).xPubKey;
+        // Malformed historical copayer state must fail closed at this boundary,
+        // never escape as an uncaught exception.
+        const target = wallet.copayers.find(c => c.id === opts.copayerId);
+        if (!target?.xPubKey) return cb(Errors.NOT_AUTHORIZED);
 
-        if (!this._verifyRequestPubKey(opts.requestPubKey, opts.signature, xPubKey)) {
+        try {
+          const isValid = this._verifyRequestPubKey(opts.requestPubKey, opts.signature, target.xPubKey);
+          if (!isValid) return cb(Errors.NOT_AUTHORIZED);
+        } catch (e) {
           return cb(Errors.NOT_AUTHORIZED);
         }
 
--- a/packages/bitcore-wallet-service/src/lib/server.ts
+++ b/packages/bitcore-wallet-service/src/lib/server.ts
@@ -1216,9 +1216,17 @@ export class WalletService implements IWalletService {
     }
     if (!Utils.checkValueInCollection(opts.chain, Constants.CHAINS)) return cb(new ClientError('Invalid coin'));
 
+    // SECURITY: xPubKey is required for every join, including hardwareSourcePublicKey/
+    // clientDerivedPublicKey ones. Copayer.xPubToCopayerId() (called below both for the TSS
+    // participant check and inside Copayer.create()) hashes opts.xPubKey directly; passing it
+    // undefined throws for the default coin and produces a colliding constant hash for every
+    // other coin. bitcore-wallet-client's only join path (_doJoinWallet) always sends a real
+    // xPubKey regardless of these two fields, so this doesn't restrict any flow this repo's
+    // own client exercises - see the TSS-participant comment below for how that was confirmed.
+    if (!checkRequired(opts, ['xPubKey'], cb)) return;
+
     let xPubKey;
     if (!opts.hardwareSourcePublicKey && !opts.clientDerivedPublicKey) {
-      if (!checkRequired(opts, ['xPubKey'], cb)) return;
       try {
         xPubKey = Bitcore_[opts.chain].HDPublicKey(opts.xPubKey);
       } catch {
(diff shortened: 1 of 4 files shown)


packages/bitcore-wallet-service/src/lib/server.ts
```diff
@@ -1207,7 +1211,17 @@ export class WalletService implements IWalletService {
    * @param {boolean} [opts.dryRun] Simulate the action but do not change server state.
    */
   joinWallet(opts, cb) {
-    if (!checkRequired(opts, ['walletId', 'name', 'requestPubKey', 'copayerSignature'], cb)) return;
+    // xPubKey is the canonical copayer identity in the current protocol. Alternate public-key
+    // fields are ancillary metadata; supporting an xPubKey-less copayer requires a versioned
+    // end-to-end identity protocol rather than a local identity fallback.
+    // SECURITY: xPubKey is required for every join, including hardwareSourcePublicKey/
+    // clientDerivedPublicKey ones. Copayer.xPubToCopayerId() (called below both for the TSS
+    // participant check and inside Copayer.create()) hashes opts.xPubKey directly; passing it
```
leolambo: since this only requires xPubKey at the caller, Copayer.create and xPubToCopayerId still accept a missing one and hash undefined into the id. consider adding a check for every caller without removing the exemption you left for the future hardware path

```js
static xPubToCopayerId(coin, xpub) {
  $.checkArgument(xpub, 'Missing xPubKey for copayer id derivation');
  const str = coin == Defaults.COIN ? xpub : coin + xpub;
  ...
}
```

MichaelAJay: Added the checkArgument to xPubToCopayerId.
I’d prefer not to restore the xPubKey exemption in joinWallet yet. Until the no-xPubKey flow has end-to-end support, allowing it through initial validation only for it to fail later during copayer ID derivation seems misleading and incomplete.
Supporting that flow properly will require two explicit identity paths:
A. xPubKey exists, and the copayer ID is derived from it.
B. xPubKey is absent and either hardwareSourcePublicKey or clientDerivedPublicKey exists, with an explicitly defined alternative for deriving the copayer ID and performing the associated validation.
Once path B is implemented throughout the join flow, we can restore the exemption safely.

leolambo (approved):

leolambo (approved):

kajoseph (approved):

PR #4225, opened 2026-08-20
Fix/ltc and doge check paypro

MichaelAJay:
### Description

Fixes IS-1413.

`Verifier.checkPaypro` previously verified destination addresses only for BTC and BCH. LTC and DOGE PayPro proposals could therefore substitute a different destination while preserving the invoiced amount and still pass co-signer verification.

This change makes PayPro verification fail closed for all supported multisig/UTXO chains: BTC, BCH, DOGE, and LTC. It compares the complete, order-independent set of transaction outputs against the PayPro instructions, including both destination and amount.

Addresses are parsed and canonicalized with each chain’s own bitcore library. This preserves equivalent address representations while ensuring malformed addresses cannot pass merely because their raw strings match.

### Changelog

* Verify PayPro destinations for DOGE and LTC in addition to BTC and BCH.
* Compare every transaction output against the complete PayPro instruction set.
* Match outputs and instructions by normalized destination and atomic amount, independent of ordering.
* Parse and canonicalize destinations with the corresponding BTC, BCH, DOGE, or LTC address library.
* Preserve equivalent BTC and LTC Bech32 casing.
* Preserve equivalent Litecoin legacy `3...` and modern `M...` P2SH representations.
* Preserve BCH cashaddr/legacy address equivalence.
* Reject malformed addresses on either the transaction proposal or PayPro instruction side without throwing.
* Reject unsupported chains and malformed PayPro data instead of accepting by default.
* Reject mismatched output counts, per-output amounts, and transaction totals.
* Preserve supported legacy transaction-proposal behavior.
* Log actionable verification failures with the transaction proposal ID when available.
* Add unit and caller-boundary regressions covering matching, substituted, equivalent, and malformed PayPro destinations.
* Update the standalone reproduction to use valid BTC, DOGE, and LTC address fixtures.
* Move checkTxProposal above its helpers and simplify return statement

### Testing Notes

Verified from `packages/bitcore-wallet-client`, with a local MongoDB instance
already running and reachable at `localhost:27017` (Docker container
`mongodb`, image `mongo:7.0`, running since 2026-07-13 — not started for this
change, just already part of the local dev environment):

* `npm run compile` — passed, no errors.
* `../../node_modules/.bin/mocha ts_build/test/verifier.test.js --grep 'checkPaypro|checkTxProposal'` — 47 passing, 0 failing.
* `npm run lint` — passed, no errors.
* `git diff --check` — passed, no whitespace errors.
* `npm test` (full package suite, including the MongoDB-backed integration tests) — **589 passing, 12 pending, 0 failing**, exit code 0. Coverage: Statements 75.29%, Branches 62.03%, Functions 72.58%, Lines 76.99%; all thresholds met, no database-related failures.

The focused coverage includes:

* Matching and substituted destinations for BTC, BCH, DOGE, and LTC.
* BTC and LTC Bech32 case equivalence.
* Litecoin legacy `3...` and modern `M...` P2SH equivalence.
* Malformed proposal and PayPro destinations for BTC, DOGE, and LTC, including asymmetric cases (malformed on only the proposal side, and only the PayPro side).
* BCH malformed-address handling and cashaddr/legacy compatibility.
* Legacy transaction proposals.
* Reordered, duplicate, missing, and mismatched output/instruction sets.
* Caller-boundary acceptance and rejection through `Verifier.checkTxProposal`.

For manual verification:

1. Create a PayPro proposal for an LTC or DOGE multisig wallet and confirm the matching merchant destination and amount are accepted.
2. Substitute the destination while retaining the invoiced amount and confirm co-signer verification rejects the proposal with `SERVER_COMPROMISED`.
3. For LTC, confirm equivalent supported address representations do not cause a false rejection.

<hr style="border-width:3px">

### Checklist


packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -10,7 +12,32 @@ import log from './log';
...
 const $ = singleton();
-const BCHAddress = BitcoreLibCash.Address;
+
+// Each supported multisig/UTXO chain canonicalizes destinations through its
+// own bitcore-lib fork's Address class rather than raw string/lowercase
+// comparison, so equivalent encodings (BTC/LTC Bech32 case, LTC legacy
+// `3...`/modern `M...` P2SH, BCH cashaddr/legacy) compare equal while
+// unparseable addresses throw and are treated as a verification failure
+// instead of matching merely because their raw strings match.
+const ADDRESS_LIB_BY_CHAIN: Record<string, { Address: new (address: string) => { toString(): string } }> = {
```
leolambo, line 24: this map only has btc, bch, doge, ltc, so every other chain falls through to the unsupported-chain return below and checkPaypro gives false, which becomes SERVER_COMPROMISED. old checkPaypro returned true for eth/xrp/erc20, and getTxProposals throws on the first failing txp, so a single eth or usdc paypro proposal now breaks the whole fetch for the 9 account based coins the ticket listed.

packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -305,53 +360,372 @@ export class Verifier {
...
+      if (amount === null) {
+        return {
+          valid: false,
+          invalidEntryIndex: index,
+          invalidField: 'amount'
+        };
+      }
 
-    if (opts.paypro && !this.checkPaypro(txp, opts.paypro)) return false;
+      normalizedEntries.push({ toAddress: entry.toAddress, amount, raw: entry });
```
leolambo, line 603: there's a potential backdoor if a `script` is provided via` packages/bitcore-wallet-client/src/lib/common/utils.ts:389`  (if (o.script) → addOutput). we might want to refuse any PayPro output that has a script at all or validate the script path as well.

packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -305,53 +360,372 @@ export class Verifier {
     return true;
   }
 
+  /**
+   * Checks a PayPro-funded transaction proposal against the signed PayPro
+   * response it was supposed to pay. Should never be used to check multiTx
+   * or multiSendContractAddress txps
```
leolambo: this precondition isn't enforced anywhere. A txp with multiSendContractAddress (or multiTx) set still passes the per-output comparison here, but buildTx signs a single call to the multisend contract instead of the outputs we just verified. Can we return false when either field is set?

packages/bitcore-wallet-client/src/lib/verifier.ts
MichaelAJay: addressed in the merged implementation

PR #4226, opened 2026-08-21
[BWS] Chore: unnecessary simplexGetEvents endpoint removed

Gamboster:
### Description
Ticket reference: [RN-2874](https://bitpayprod.atlassian.net/browse/RN-2874)

The `simplexGetEvents` function and its BWS endpoint were removed, as they were not currently used by either the app or the web.

### Changelog

Only unnecessary code removed.

### DIFF
--- a/packages/bitcore-wallet-service/src/externalservices/simplex.ts
+++ b/packages/bitcore-wallet-service/src/externalservices/simplex.ts
@@ -248,34 +248,4 @@ export class SimplexService {
       );
     });
   }
-
-  simplexGetEvents(req): Promise<any> {
-    return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


kajoseph (commented):
There is a `simplexGetEvents` method in BWC that maps to this endpoint. This method/endpoint was _never_ used by any previous versions of the app/web? In other words, removing this won't break any historical versions of the app, will it?

Gamboster:
> There is a `simplexGetEvents` method in BWC that maps to this endpoint. This method/endpoint was _never_ used by any previous versions of the app/web? In other words, removing this won't break any historical versions of the app, will it?

Yes, I removed the BWC method as well. Thanks for noticing. Indeed, no version of the app/web used that function (So no version should break).

PR #4232, opened 2026-08-28
[BWS][BWC][CWC] Bind prePublishRaw fallback to the proposal

leolambo:
### Description
[IS-1422](https://bitpayprod.atlassian.net/browse/IS-1422?issueKey=IS-1422&subProduct=jira-core) · Bug Bounty 2643811

The transaction verifier falls back to `txp.prePublishRaw` when it can't match the creator's signature against the locally rebuilt tx. It only checked that the signature was valid over `prePublishRaw`, not that `prePublishRaw` was the same proposal, so a compromised wallet service could keep a valid signature, swap the destination, and still pass the check. The fallback now recovers the one field the server changes at publish (the Solana blockhash or the EVM/XRP nonce) from `prePublishRaw`, puts it back on the current proposal, and requires a byte-for-byte match, so any other change fails as SERVER_COMPROMISED.

### Changelog

* Recover the server-mutated field from `prePublishRaw` and require the rebuilt proposal to match it exactly, so a tampered destination or amount fails as SERVER_COMPROMISED.
* Reject `prePublishRaw` on UTXO chains, which never mutate a field at publish.
* Guard the server-side `publishTx` fallback the same way, so a tampered stored proposal can't reuse an old signature.
* Add `getMutableFields` to the SOL, EVM, and XRP providers to pull that field back out of a raw tx.

### Testing Notes
New tests at each layer:

* crypto-wallet-core: `getMutableFields` recovery for SOL, EVM, and XRP
* bitcore-wallet-client: `checkPrePublishRaw` accepts a refreshed field and rejects a tampered destination
* bitcore-wallet-service: `isPrePublishRawBound` plus an end-to-end `publishTx` that rejects a tampered stored proposal

Run `npm test` in each of those three packages.

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-wallet-client/src/lib/verifier.ts
+++ b/packages/bitcore-wallet-client/src/lib/verifier.ts
@@ -1,7 +1,8 @@
 import {
   BitcoreLib as Bitcore,
   BitcoreLibCash,
-  Utils as CWCUtils
+  Utils as CWCUtils,
+  Transactions
 } from '@bitpay-labs/crypto-wallet-core';
 import { singleton } from 'preconditions';
 import { Constants, Utils } from './common';
@@ -278,14 +279,24 @@ export class Verifier {
     log.debug(`[TXP ${txp.id}] Regenerating & verifying tx proposal hash -> Hash: ${hash}, Signature: ${txp.proposalSignature}`);
   
     const verified = Utils.verifyMessage(hash, txp.proposalSignature, creatorSigningPubKey);
-    if (!verified && !txp.prePublishRaw) {
-      log.debug(`[TXP ${txp.id}] Invalid proposal signature, no prePublishRaw to fall back to`);
-      return false;
-    }
-    
-    if (!verified && txp.prePublishRaw && !Utils.verifyMessage(txp.prePublishRaw, txp.proposalSignature, creatorSigningPubKey)) {
-      log.debug(`[TXP ${txp.id}] Invalid proposal signature, even with prePublishRaw fallback`);
-      return false;
+    if (!verified) {
+      // Local rebuild != creator's signature. Legit only when BWS mutated a field at publish (SVM recent
+      // blockhash, or EVM/XRP deferred nonce): the creator signed the pre-publish serialization, stored as
+      // txp.prePublishRaw. Fall back to it only if the signature is valid over it AND it is bound to this
+      // proposal -- else a hostile server could pair a valid (prePublishRaw, proposalSignature) with a
+      // tampered destination. See checkPrePublishRaw.
+      if (!txp.prePublishRaw) {
+        log.debug(`[TXP ${txp.id}] Invalid proposal signature, no prePublishRaw to fall back to`);
+        return false;
+      }
+      if (!Utils.verifyMessage(txp.prePublishRaw, txp.proposalSignature, creatorSigningPubKey)) {
+        log.debug(`[TXP ${txp.id}] Invalid proposal signature, even with prePublishRaw fallback`);
+        return false;
+      }
+      if (!this.checkPrePublishRaw(chain, txp)) {
+        log.warn(`[TXP ${txp.id}] prePublishRaw is not bound to this proposal; possible server tampering`);
+        return false;
+      }
     }
 
     if (Constants.UTXO_CHAINS.includes(chain)) {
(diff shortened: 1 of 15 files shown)


packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -305,6 +316,51 @@ export class Verifier {
...
+      log.warn(`[TXP ${txp.id}] prePublishRaw present on chain ${chain} that cannot mutate at publish; refusing fallback`);
+      return false;
+    }
+    try {
+      const prePublishRaw = Array.isArray(txp.prePublishRaw) ? txp.prePublishRaw : [txp.prePublishRaw];
+      // Recover the mutable field (blockhash / nonce) from the pre-publish serialization the creator signed;
+      // the stored proposal carries the refreshed value, so it isn't readable off txp directly.
+      const provider: any = Transactions.get({ chain });
+      const mutableFields = provider.getMutableFields(prePublishRaw[0]);
+      if (!mutableFields || Object.values(mutableFields).every(v => v == null)) {
```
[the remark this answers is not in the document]
leolambo: invalid since this is gated above by chain so only valid chains hit `provider.getMutableFields`. the  CWC providers map every one of those chains to a provider that implements the method

packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -305,6 +316,51 @@ export class Verifier {
     return true;
```
MichaelAJay, line 316: I think the problem is larger than playing whack-a-mole with known false cases. Consider how to make this return true only for explicit trust situations, instead of for all situations we can't explicitly think of.

packages/bitcore-wallet-client/test/verifier.test.ts
```diff
@@ -222,4 +223,61 @@ describe('Verifier', function() {
       }).should.be.true;
     });
   });
+
```
MichaelAJay, line 226: Do we need XRP?

packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -278,41 +278,59 @@ export class Verifier {
...
+   * function terminates on an explicit proven-accept verdict rather than a fall-through.
+   *
+   * @param {Object} credentials
+   * @param {string} chain - lower-cased chain of the proposal
+   * @param {Object} txp - the transaction proposal
+   */
+  static checkUtxoAddresses(credentials, chain, txp) {
+    if (!Constants.UTXO_CHAINS.includes(chain)) {
+      return true;
+    }
```
MichaelAJay: This, in isolation, would let something through like `chain = 'foo'`. Maybe define a new constant in constants.js: 'SUPPORTED_NON_UTXO_CHAINS' or 'ACCOUNT_CHAINS' = [...EVM_CHAINS, ...SVM_CHAINS, ...RIPPLE_CHAINS]. Or just use that in place:
`if ([...EVM_CHAINS, ...SVM_CHAINS, ...RIPPLE_CHAINS].includes(chain)) return true`

packages/bitcore-wallet-client/src/lib/verifier.ts
```diff
@@ -278,41 +278,59 @@ export class Verifier {
...
+  /**
+   * For UTXO chains, verifies the change and escrow addresses belong to this wallet. Returns true for
+   * non-UTXO chains (they carry no such addresses to prove). Split out of checkTxProposalSignature so that
+   * function terminates on an explicit proven-accept verdict rather than a fall-through.
+   *
+   * @param {Object} credentials
+   * @param {string} chain - lower-cased chain of the proposal
+   * @param {Object} txp - the transaction proposal
+   */
+  static checkUtxoAddresses(credentials, chain, txp) {
```
MichaelAJay, line 319: Overall, this is still a deny-list. It's filtering out what, today, it CAN'T be. It would be much stronger if we filtered in only what it SHOULD be.

Consider the specific cases which are true today:
- If txp.sendMax, then ensure no change address. If not, ensure change address & check it
- AND if txp.instantAcceptanceEscrow, then ensure escrowAddress & check it. If not, ensure no escrowAddress.

Recommended form:
```javascript
// explicit chain allow
if (
  [...Constants.EVM_CHAINS,
  ...Constants.SVM_CHAINS,
  ...Constants.RIPPLE_CHAINS].includes(chain)
) {
  return !txp.changeAddress && !txp.escrowAddress;
  // or just true
}

// followed by unknown chain deny
if (!Constants.UTXO_CHAINS.includes(chain)) {
  return false;
}

// conditional utxo chain allows
const validChange =
  (!txp.changeAddress ||
    this.checkAddress(credentials, txp.changeAddress)) &&
  (Boolean(txp.changeAddress) || Boolean(txp.sendMax));

const validEscrow =
  !txp.escrowAddress ||
  this.checkAddress(credentials, txp.escrowAddress, txp.inputs);

return validChange && validEscrow;
```

As of today, the conditional utxo chain allowances are directly logically equivalent to the currently existing logic, but are fallthrough-false instead of true.

MichaelAJay: While what is above is logically equivalent to your code - should we also be checking txp.instantAcceptanceEscrow?
```javascript
// replaces validEscrow assignment above
const validEscrow =
  (!txp.instantAcceptanceEscrow || !!txp.escrowAddress) &&
  (!txp.escrowAddress ||
    this.checkAddress(credentials, txp.escrowAddress, txp.inputs));
```

PR #4243, opened 2026-09-09
Long poll TSS sessions

kajoseph:
### Description
Add a long poll for the TSS session get endpoints to reduce both latency and the number of API calls

### Changelog

* Add long poll logic to the TSS get endpoints
* Reduce noise for TSS signing sessions

### Testing Notes

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/bitcore-cli/src/tss.ts
+++ b/packages/bitcore-cli/src/tss.ts
@@ -87,8 +87,10 @@ export async function sign(args: {
     let rejected = false;
     const reject = (err) => { if (!rejected) { rejected = true; _reject(err); } };
 
+    let connErrs = 0;
     tssSign.subscribe();
     tssSign.on('roundsubmitted', (round) => {
+      connErrs = 0;
       storeSession(tssSign.exportSession());
       spinner.message(`Round ${round} submitted`);
     });
@@ -103,6 +105,14 @@ export async function sign(args: {
         spinner.cancel(e.message);
         rmSessionState();
         return reject(new ProcessCancelled());
+      } else if (e instanceof Errors.CONNECTION_ERROR) {
+        // Reduce the noise of transient errors
+        connErrs++;
+        if (connErrs > 10) {
+          prompt.log.warn(e.message);
+          connErrs = 0;
+        }
+        return;
       }
       prompt.log.error('Unexpected error during TSS signing: ' + (e.stack || e));
     });
--- a/packages/bitcore-cli/src/tss.ts
+++ b/packages/bitcore-cli/src/tss.ts
@@ -49,6 +49,15 @@ export async function sign(args: {
     tssKey: walletData.key as TssKeyType
   });
 
+  // Reduces the noise of transient connection errors
+  const connResilience = (count: number, e: Error): number => {
+    if (count > 10) {
+      prompt.log.warn(e.message);
+      return 0;
+    }
+    return ++count;
+  };
+
   // Restore a previously-interrupted TSS session if it exists
   if (fs.existsSync(storedSessionFile)) {
     const storedSession = Encryption.decryptWithPassword(fs.readFileSync(storedSessionFile, 'utf8'), password);
@@ -57,27 +66,36 @@ export async function sign(args: {
 
   // ...otherwise, start a new TSS session
   } else {
-    try {
-      await tssSign.start({
-        id,
-        messageHash,
-        derivationPath,
-        password
-      });
-      storeSession(tssSign.exportSession());
-    } catch (err) {
-      if (err.message?.startsWith('TSS_ROUND_ALREADY_DONE')) {
-        const sig = await tssSign.getSignatureFromServer();
-        if (!sig) {
-          throw new Error('It looks like the TSS signature session was interrupted. Try deleting this proposal and creating a new one.');
+    let isTransientError = false;
+    let connErrs = 0;
+    do {
+      try {
+        await tssSign.start({
+          id,
+          messageHash,
+          derivationPath,
+          password
+        });
+        storeSession(tssSign.exportSession());
+      } catch (err) {
+        isTransientError = false; // reset
+        if (err.message?.startsWith('TSS_ROUND_ALREADY_DONE')) {
+          const sig = await tssSign.getSignatureFromServer();
+          if (!sig) {
+            throw new Error('It looks like the TSS signature session was interrupted. Try deleting this proposal and creating a new one.');
+          }
+          return {
+            signature: transformISignature(sig),
+            publicKey: sig.pubKey
+          };
+        } else if (err instanceof Errors.CONNECTION_ERROR) {
+          isTransientError = true;
+          connErrs = connResilience(connErrs, err);
+        } else {
+          throw err;
         }
-        return {
-          signature: transformISignature(sig),
-          publicKey: sig.pubKey
-        };
       }
-      throw err;
-    }
+    } while (isTransientError);
   }
 
   const spinner = prompt.spinner({ indicator: 'timer', onCancel: () => { tssSign.unsubscribe(); } });
@@ -107,14 +125,11 @@ export async function sign(args: {
         return reject(new ProcessCancelled());
       } else if (e instanceof Errors.CONNECTION_ERROR) {
         // Reduce the noise of transient errors
-        connErrs++;
-        if (connErrs > 10) {
-          prompt.log.warn(e.message);
-          connErrs = 0;
-        }
+        connErrs = connResilience(connErrs, e);
         return;
+      } else {
+        prompt.log.error('Unexpected error during TSS signing: ' + (e.stack || e));
       }
-      prompt.log.error('Unexpected error during TSS signing: ' + (e.stack || e));
     });
     tssSign.on('complete', async () => {
       try {
--- a/packages/bitcore-wallet-client/src/lib/errors/spec.ts
+++ b/packages/bitcore-wallet-client/src/lib/errors/spec.ts
@@ -147,7 +147,7 @@ export const errorSpec: IErrorSpec[] = [
       'We were unable to parse your payment. Please try again or contact your wallet provider.'
   },
   {
-    name: 'NO_TRASACTION',
+    name: 'NO_TRANSACTION',
     message:
       'Your request did not include a transaction. Please try again or contact your wallet provider.'
   },
--- a/packages/bitcore-wallet-service/src/lib/routes/tss.ts
+++ b/packages/bitcore-wallet-service/src/lib/routes/tss.ts
@@ -37,17 +37,33 @@ export class TssRouter {
     });
 
     router.get('/v1/tss/keygen/:id/:round', authTssRequest(), async function(req, res) {
+      let interval: NodeJS.Timeout;
       try {
         const { id, round } = req.params as { [key: string]: string };
+        const { maxWaitTime } = req.query as { [key: string]: string };
         const copayerId = req.headers['x-identity'];
         if (round === 'secret') {
           const secret = await TssKeyGen.getBwsJoinSecret({ id, copayerId });
           return res.json({ secret });
         }
-        const { messages, publicKey } = await TssKeyGen.getMessagesForParty({ id, round: parseInt(round), copayerId });
-        return res.json({ messages, publicKey });
+
+        // Validate access and fetch session before committing to a streaming response, so that errors like
+        //   "session not found" or "not a participant" can still be returned with a proper
+        //   HTTP status instead of silently becoming an empty 200 (see below).
+        const session = await TssKeyGen.getSessionForCopayer({ id, copayerId });
+
+        // Keep the connection alive while waiting for the change stream to return a result.
+        // Headers must be finalized before writing the first heartbeat byte.
+        res.writeHead(200, { 'Content-Type': 'application/json' });
+        interval = setInterval(() => res.write('\n'), 1000);
+        req.on('close', () => clearInterval(interval));
+
+        const { messages, publicKey } = await TssKeyGen.getMessagesForParty({ session, round: parseInt(round), copayerId, maxWaitTimeSec: parseInt(maxWaitTime) });
+        return res.end(JSON.stringify({ messages, publicKey }));
       } catch (err) {
         return returnError(err ?? 'unknown', res, req);
+      } finally {
+        clearInterval(interval);
       }
     });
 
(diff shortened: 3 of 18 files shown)


packages/bitcore-wallet-service/src/lib/tss.ts
```diff
@@ -1,36 +1,140 @@
...
+  const sessionUpdateHandler = async (message: INotification) => {
+    if (message.type !== messageType) {
+      return;
+    }
+    if (message.id !== session.id) {
+      return;
+    }
+    try {
+      const _session = await fetchSession({ id: session.id });
+      if (isComplete(_session)) {
```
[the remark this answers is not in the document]
kajoseph: This is not an issue since session is either complete or incomplete. Updating to the latest when it's still incomplete is moot.

packages/bitcore-wallet-service/src/lib/tss.ts
```diff
@@ -1,36 +1,140 @@
...
+        messageBroker.offMessage(sessionUpdateHandler);
+        events.emit('session', _session);
```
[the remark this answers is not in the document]
kajoseph: I simply moved the `clearTimeout` to the end of the whole function so it clears it in all cases. Putting it here to guard a race condition where the timeout returns the stale session _just before_ this emits the update session will, at worst, result in the API returning that the round is incomplete and the client will try again. In other words, the worst case scenario is a little extra latency.

On the other hand, I fear that removing the timer prematurely could result in a forever promise.

packages/bitcore-wallet-client/src/lib/errors/spec.ts
```diff
@@ -147,7 +147,7 @@ export const errorSpec: IErrorSpec[] = [
       'We were unable to parse your payment. Please try again or contact your wallet provider.'
   },
   {
-    name: 'NO_TRASACTION',
+    name: 'NO_TRANSACTION',
```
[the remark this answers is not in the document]
kajoseph: Instances used in bitpay-app are identified. A separate PR to bitpay-app will be created

packages/bitcore-wallet-service/src/lib/tss.ts
```diff
@@ -1,36 +1,178 @@
...
+    void notify(message).catch(err => {
+      logger.error('Error handling TSS session update: %o', err);
+    });
+  }
+}
+
+function subscribeToSession(type: string, id: string, handler: SessionHandler): () => void {
+  if (!dispatcherRegistered) {
+    WalletService.getMessageBroker().onMessage(dispatchSessionMessage);
+    dispatcherRegistered = true;
```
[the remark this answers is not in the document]
kajoseph: This is beyond the scope of this PR. Initialize is only called once on startup. The messagebroker being overwritten is not likely, but it would be good to guard against it.

packages/bitcore-wallet-service/src/lib/routes/tss.ts
```diff
@@ -37,17 +37,34 @@ export class TssRouter {
...
+        //   "session not found" or "not a participant" can still be returned with a proper
+        //   HTTP status instead of silently becoming an empty 200 (see below).
+        const session = await TssKeyGen.getSessionForCopayer({ id, copayerId });
+
+        // Keep the connection alive while waiting for the change stream to return a result.
+        // Headers must be finalized before writing the first heartbeat byte.
+        // Flush ensures the heartbeat is sent immediately to keep the connection alive.
+        res.writeHead(200, { 'Content-Type': 'application/json' });
+        interval = setInterval(() => { res.write('\n'); res.flush(); }, 1000);
+        req.on('close', () => clearInterval(interval));
```
MichaelAJay: This is to clean up the heartbeat when the client disconnects?
For Node 22, `req`'s `close` is emitted when the request completes. `res.close` is emitted when the response is completed or its underlying connection was terminated prematurely.

Consider `res.once('close',...)`.

Same thing on `'/v1/tss/sign/:id/:round'`.

https://nodejs.org/docs/latest-v22.x/api/http.html#event-close_3
https://nodejs.org/docs/latest-v22.x/api/http.html#event-close_2

packages/bitcore-wallet-service/src/lib/tss.ts
```diff
@@ -103,7 +105,8 @@ async function listenForSessionComplete<T extends TssKeyGenModel | TssSigGenMode
   let timer: NodeJS.Timeout;
   const sessionUpdate = Promise.race([
     new Promise<T>(r => events.once('session', r)),
-    new Promise<T>(r => timer = setTimeout(() => { unsubscribe(); r(session); }, maxWaitTime))
+    new Promise<T>(r => timer = setTimeout(() => { unsubscribe(); r(session); }, maxWaitTime)),
+    new Promise<T>((_, j) => abortSignal?.addEventListener('abort', () => { unsubscribe(); j(new Error('Aborted')); }))
```
MichaelAJay, line 109: 1) Need to remove the `abortSignal.abort` event listener in `finally`
2) Need to handle this rejectable side Promise at creation - if abort signal is received before `await sessionUpdate` (including if that's never executed), then there's nothing to catch it & it will result in an uncaught error. I'd probably just not reject at all here.

MichaelAJay: Additionally, the timeout & abort can't finish a stalled `fetchSession`

packages/bitcore-wallet-service/src/lib/routes/tss.ts
```diff
@@ -55,11 +55,12 @@ export class TssRouter {
         // Keep the connection alive while waiting for the change stream to return a result.
         // Headers must be finalized before writing the first heartbeat byte.
         // Flush ensures the heartbeat is sent immediately to keep the connection alive.
+        const abort = new AbortController();
```
MichaelAJay, line 58: move this & `res.on('close')` at the top of the method - then check for `abort.signal.aborted` & return early right after `await getSessionForCopayer`.

packages/bitcore-wallet-service/src/lib/tss.ts
```diff
@@ -1,36 +1,183 @@
...
+    copayerId: string;
+  }): Promise<TssKeyGenModel> {
+    const { id, copayerId } = params;
+    const storage = WalletService.getStorage();
+    const session = await storage.fetchTssKeyGenSession({ id });
+    if (!session) {
+      throw Errors.TSS_SESSION_NOT_FOUND;
+    }
+
+    const partyId = session.participants.indexOf(copayerId);
```
MichaelAJay, line 149: For an existence check, `includes` is more semantically appropriate
Could just be:
`if (!session.participants.includes(copayerId)) throw ...`

packages/bitcore-wallet-service/src/lib/tss.ts
kajoseph: 1. The AbortController is contained w/in the scope of the API request. Once the request ends, it gets garbage collected. Cleaning it up here would require a significant "gumming up" the code which I don't consider to be worth it and I'm fine letting the GC handle this one.
2. Actually, it would get caught further up the call stack - in this case, in the API handler. The reason I went with a rejection instead a `r(session)` is because this is a client-driven abort, so there's no reason to go the happy-route of checking if the session is complete. Rather, rejecting here short circuits straight to the catch in the API handler.

What does "stalled fetchSession" mean, specifically? I know what it means in the abstract, but in this case it's just a DB call. If it somehow gets bogged down by a COLSCAN or something heinous, we have bigger problems.

packages/bitcore-wallet-service/src/lib/routes/tss.ts
kajoseph: Not needed. The abort will result in the catch below being hit. But even if it doesn't (future changes?), the client has closed the connection, so whatever happens after this point is largely moot. If the criteria for abort changes, I wouldn't presume to know what that criteria might be in the future or how to properly handle it.

packages/bitcore-cli/src/tss.ts
```diff
@@ -57,27 +66,36 @@ export async function sign(args: {
...
+        if (err.message?.startsWith('TSS_ROUND_ALREADY_DONE')) {
+          const sig = await tssSign.getSignatureFromServer();
+          if (!sig) {
+            throw new Error('It looks like the TSS signature session was interrupted. Try deleting this proposal and creating a new one.');
+          }
+          return {
+            signature: transformISignature(sig),
+            publicKey: sig.pubKey
+          };
+        } else if (err instanceof Errors.CONNECTION_ERROR) {
```
MichaelAJay, line 91: I think this will read better if you swap this with the `if` clause. Reduces the noise around the loop. Tomato/tomato

packages/bitcore-wallet-service/src/lib/tss.ts
```diff
@@ -1,36 +1,183 @@
...
+      logger.error('Error fetching updated TSS session: %o - %o', session.id, err);
+    }
+  };
+  const unsubscribe = subscribeToSession(messageType, session.id, sessionUpdateHandler);
+  
+  // Listen for session update events over the message broker service
+  let timer: NodeJS.Timeout;
+  const sessionUpdate = Promise.race([
+    new Promise<T>(r => events.once('session', r)),
+    new Promise<T>(r => timer = setTimeout(() => { unsubscribe(); r(session); }, maxWaitTime)),
```
MichaelAJay, line 108: Should we refresh the session once on timeout?

packages/bitcore-wallet-service/src/lib/routes/tss.ts
```diff
@@ -104,13 +122,32 @@ export class TssRouter {
...
+        // Validate access and fetch session before committing to a streaming response, so that errors like
+        //   "session not found" or "not a participant" can still be returned with a proper
+        //   HTTP status instead of silently becoming an empty 200 (see below).
+        const session = await TssSign.getSessionForCopayer({ id, copayerId });
+
+        // Keep the connection alive while waiting for the change stream to return a result.
+        // Headers must be finalized before writing the first heartbeat byte.
+        // Flush ensures the heartbeat is sent immediately to keep the connection alive.
+        const abort = new AbortController();
+        res.writeHead(200, { 'Content-Type': 'application/json' });
```
MichaelAJay, line 140: If anything downstream of writing 200 throws, then returnError short circuits with 200 and I believe an empty body.

bwc doRequest is going to let that 200 through, then tsssign's `subscribe` does this:
```javascript
        const { body } = await this.#request.get(`/v1/tss/sign/${this.id}/${prevRound}`) as RequestResponse;

        if (body.participants?.length) {
```
You'll get a TypeError there since body is undefined.

packages/bitcore-cli/src/tss.ts
```diff
@@ -57,27 +66,36 @@ export async function sign(args: {
...
+          throw err;
         }
-        return {
-          signature: transformISignature(sig),
-          publicKey: sig.pubKey
-        };
       }
-      throw err;
-    }
+    } while (isTransientError);
```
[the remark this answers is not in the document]
MichaelAJay: @kajoseph - this is correct. Once we get into this state, then not only will `Errors.CONNECTION_ERROR` continue the loop indefinitely - but even a successfully-executed try block is going to not exit the loop. At the very least, you need to reassign `isTransientError = false` at the end of the try block, after anything that's potentially throwable. But setting it false at the top of the loop body may be better yet. Then the only way that `isTransientError` will be true at loop-eval time will be as a result of that assignment in the else if.

MichaelAJay:
@kajoseph tangent to your work - bwc Request constructor doesn't set timeout - nothing does.
`Property 'timeout' has no initializer and is not definitely assigned in the constructor.`

packages/bitcore-wallet-service/src/lib/routes/tss.ts
MichaelAJay: I was thinking about the client closing while getSessionForCopayer() is pending, before the close listener is registered. In that case no abort occurs - the close event is missed. Registering it before the await and checking abort.signal.aborted afterward prevents us from continuing setup for an already-closed response. It’s primarily closing that race/resource-cleanup gap. And `clearInterval(undefined)` is a no-op, so there's no problem there.

It's a small window - but it's a window. There's a useful heuristic here that I think the codebase at large could benefit from:
```
Attach an event handler before the first async boundary during which an event can occur
```

packages/bitcore-wallet-service/src/lib/tss.ts
MichaelAJay: The API handler can only catch the thrown error from Promise.race (assuming the abort signal's reject wins) once it's `await`ed. But it will still reject on abort (assuming abort wins) - and there's nothing to catch it because it's not connected to the outer async chain - it's independent until it's anchored in on `await`.

packages/bitcore-wallet-service/src/lib/tss.ts
kajoseph: I would say no. At worst, it returns a stale session and the client retries. As it is today, the client retries every second, and the default timeout is 20-30s. So after 20-30s, what's another 1s in latency? And it keeps the code cleaner.

kajoseph:
> @kajoseph tangent to your work - bwc Request constructor doesn't set timeout - nothing does. `Property 'timeout' has no initializer and is not definitely assigned in the constructor.`

It doesn't need to. The heartbeat byte stream works around HTTP header constraints. There doesn't need to be any client change for the long poll here to work. The timeout is passed as `maxWaitTime` query param - which the client currently does not set. I'm leaving that work for a subsequent PR _if_ it's needed. Currently, letting the server define it is better.

packages/bitcore-cli/src/tss.ts
kajoseph: I'm fine with it as is. I expect this whole clause will be extended with other error cases soon. There may be a cleaner, more elegant way of handling these errors, but I'm not sure what they are yet.

packages/bitcore-wallet-service/src/lib/routes/tss.ts
kajoseph: Body is an empty string. It doesn't throw a TypeError and the client retries as usual

packages/bitcore-wallet-service/src/lib/routes/tss.ts
```diff
@@ -52,13 +55,15 @@ export class TssRouter {
         //   HTTP status instead of silently becoming an empty 200 (see below).
         const session = await TssKeyGen.getSessionForCopayer({ id, copayerId });
 
+        if (abort.signal.aborted || res.closed) {
```
MichaelAJay, line 58: Isn't the client gone here? Should this just be a return? At any rate, this will result in a 500, which, if we ever measured, would skew actual error results.

packages/bitcore-cli/src/tss.ts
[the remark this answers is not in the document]
kajoseph: Nothing's changed with the util's behavior - this is merely to reduce the logging jitter of transient connection errors which happens quite often when someone's internet connection is unstable, during a BWS deploy, etc. This comment was a case of AI identifying a pattern but not understanding the purpose.

packages/bitcore-wallet-service/src/lib/routes/tss.ts
MichaelAJay: Confirmed. But do we really want this to be a 200 in this case? It will obfuscate fully whatever useful information the server might otherwise return to the client.

kajoseph: We don't _want_ a 200 but there isn't really a way around it without significantly changing the protocol being used. Once the heartbeat byte starts to keep the connection alive, the headers have been sent. A websocket would be a better protocol for this type of communication, but that's a significant overhaul to all of this and a good candidate for a v2. This PR is meant to just be a drop-in, works today, lower-effort solution.

MichaelAJay (approved):

nitsujlangston (approved):

PR #4244, opened 2026-09-10
Fix/sol rpc get account info

MichaelAJay:
### Description
Previously, SolRpc specifically handled one type of error (trying to use getAccountInfo for ATAs). This problem could actually be mitigated by specifying the encoding as base64. This change is necessitated by the need for an account's space.

Tests were updated and added to verify behavior.

### Changelog

* Add `{ encoding: 'base64' }` to Solana RPC getAccountInfo call
* Move getAccountInfo out of try/catch - no longer needed since issue is mitigated by above
* Add tests

### Testing Notes

<hr style="border-width:3px">

### Checklist

### DIFF
--- a/packages/crypto-rpc/lib/sol/SolRpc.js
+++ b/packages/crypto-rpc/lib/sol/SolRpc.js
@@ -977,7 +977,7 @@ export class SolRpc {
       effectiveMaxDepth = 0;
     }
     const atas = await this.getTokenAccountsByOwner({ address, skipExistenceCheck: true, maxDepth: effectiveMaxDepth });
-    return { lamports, atas, space: accountInfoResponse.value.space };
+    return { lamports, atas, space: accountInfoResponse.value?.space };
   }
 
   /**
--- a/packages/crypto-rpc/lib/sol/SolRpc.js
+++ b/packages/crypto-rpc/lib/sol/SolRpc.js
@@ -991,7 +991,7 @@ export class SolRpc {
   async getTokenAccountsByOwner({ address, skipExistenceCheck = false, maxDepth = 0 }) {
     // Only explicit skipExistenceCheck: true should bypass
     if (skipExistenceCheck !== true) {
-      const accountInfoResponse = await this.rpc.getAccountInfo(address).send();
+      const accountInfoResponse = await this.rpc.getAccountInfo(address, { encoding: 'base64' }).send();
       if (!accountInfoResponse.value) {
         throw new Error(SOL_ERROR_MESSAGES.SOL_ACCT_NOT_FOUND);
       }
--- a/packages/crypto-rpc/lib/sol/SolRpc.js
+++ b/packages/crypto-rpc/lib/sol/SolRpc.js
@@ -980,7 +980,12 @@ export class SolRpc {
       effectiveMaxDepth = 0;
     }
     const atas = await this.getTokenAccountsByOwner({ address, skipExistenceCheck: true, maxDepth: effectiveMaxDepth });
-    return { lamports, atas, space: accountInfoResponse.value ? Number(accountInfoResponse.value.space) : undefined };
+    return { 
+      lamports,
+      atas,
+      owner: accountInfoResponse.value?.owner,
+      space: accountInfoResponse.value ? Number(accountInfoResponse.value.space) : undefined
+    };
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/crypto-rpc/lib/sol/SolRpc.js
```diff
@@ -963,28 +963,21 @@ export class SolRpc {
    * @returns 
    */
   async getAccountInfo({ address, maxDepth }) {
```
leolambo, line 965: bitpay calls getAccountInfo through ChainInterface in isAddressPayable (lib/utils.js:511) and relies on the catch at :549 to return false for ATA addresses. With the throw gone an ATA passes validation as a SOL destination, and SOL sent to an ATA is stuck unless the account gets closed. We need a guard on the bitpay side first, something like reject when space > 0, or keep some way to flag ATAs here.

packages/crypto-rpc/lib/sol/SolRpc.js
```diff
@@ -963,28 +963,21 @@ export class SolRpc {
...
+    let effectiveMaxDepth;
+    if (maxDepth === -1) {
+      effectiveMaxDepth = Infinity;
+    } else if (typeof maxDepth === 'number' && maxDepth >= 1) {
+      effectiveMaxDepth = maxDepth;
+    } else {
+      effectiveMaxDepth = 0;
     }
+    const atas = await this.getTokenAccountsByOwner({ address, skipExistenceCheck: true, maxDepth: effectiveMaxDepth });
+    return { lamports, atas, space: accountInfoResponse.value?.space };
```
leolambo: space comes back as a bigint but everything else in this return is Number coerced, including lamports on the same line. Two real problems: JSON.stringify throws on bigint, and 0n is falsy, so any `if (!space)` check rejects normal wallets since 0 is the typical wallet space. Suggest `space: accountInfoResponse.value ? Number(accountInfoResponse.value.space) : undefined`.

packages/crypto-rpc/lib/sol/SolRpc.js
```diff
@@ -963,28 +963,21 @@ export class SolRpc {
...
-      return { lamports, atas };
-    } catch (err) {
-      const errMsg = err.message.toLowerCase();
-      if (SolKit.isSolanaError(err) && errMsg.includes('json-rpc') && errMsg.includes('should be less than 128 bytes')) {
-        // This message can occur when getAccountInfo is called with an SPL address instead of a SOL address
-        throw new Error(SOL_ERROR_MESSAGES.ATA_ADD_SENT_INSTEAD_OF_SOL_ADD);
-      }
-      throw err;
+    const accountInfoResponse = await this.rpc
+      .getAccountInfo(address, { encoding: 'base64' })
```
leolambo: Nothing here reads .data, but base64 drops the old base58 size limit that was keeping these responses small. A big account now sends its whole data payload on every call. Adding `dataSlice: { offset: 0, length: 0 }` keeps the fix and the response stays tiny. Same applies to the call at :994.

packages/crypto-rpc/lib/sol/SolRpc.js
```diff
@@ -963,28 +963,21 @@ export class SolRpc {
...
-      } else {
-        effectiveMaxDepth = 0;
-      }
-      const atas = await this.getTokenAccountsByOwner({ address, skipExistenceCheck: true, maxDepth: effectiveMaxDepth });
-      return { lamports, atas };
-    } catch (err) {
-      const errMsg = err.message.toLowerCase();
-      if (SolKit.isSolanaError(err) && errMsg.includes('json-rpc') && errMsg.includes('should be less than 128 bytes')) {
-        // This message can occur when getAccountInfo is called with an SPL address instead of a SOL address
-        throw new Error(SOL_ERROR_MESSAGES.ATA_ADD_SENT_INSTEAD_OF_SOL_ADD);
```
leolambo, line 984: ATA_ADD_SENT_INSTEAD_OF_SOL_ADD has no references left after this PR, the throw and the test that used it are both gone. error_messages.js:9 can be deleted, unless we end up keeping the throw per the other comment.

packages/crypto-rpc/lib/sol/SolRpc.js
MichaelAJay: Nice catch. I'll make that change in the bitpay MR that pulls this crypto-rpc change in - I'll have you review that too so we'll make sure that gets through. That method is actually what precipitated this change in the first place.

kajoseph (approved):

PR #4250, opened 2026-09-16
[BCN] Add wallet stats collection service

leolambo:
### Description
Adds a `WalletStatsService` that records weekly wallet statistics per chain and network: wallet counts, balances, activity windows, and duplicate detection, stored in two new collections (`walletstats`, `walletstatswallets`). These numbers are currently produced by long-running operational scripts that iterate every wallet through the chain state providers; a full run takes days and the EVM pass is prone to hanging on stalled provider requests. With snapshots accumulating on a schedule, wallet statistics become a database query.

### Changelog

* `ExternalApiStream` page requests now carry a 90s timeout and a default page cap. A stalled or endlessly-paginating external provider surfaces as a stream error within the timeout. This applies to all wallet tx streaming, and fixes the hang the EVM stats script kept hitting in prod.
* New `WalletStatsService` snapshots wallet counts, balances, activity windows (14d/30d/90d/6m/12m) and duplicate-wallet counts weekly. UTXO chains come from indexed aggregations over `coins`. EVM chains use per-address balance and nonce reads plus one bounded, spam-filtered Moralis check that catches wallets whose only activity is receiving tokens.
* Snapshots are watermark-driven and idempotent. Missed weeks get recorded in `meta.gaps`, re-runs upsert on a unique `{chain, network, date}` key, and chains that are neither UTXO nor EVM (XRP, SOL) are skipped with a warning.
* Per-wallet facts (creation date, balance, nonce, last activity, dup flag) are kept in `walletstatswallets` so cohort-style questions can be answered from stored data.
* Runs as its own worker (`npm run walletstats`, with `--EXIT` for cron use) or inside the main process. Disabled by default; merging this changes nothing until `services.walletStats.disabled: false` is set.

### Example config

```json
"services": {
  "walletStats": {
    "disabled": false,
    "snapshotDayUTC": 1,
    "snapshotHourUTC": 2,
    "sleepMs": 50,
    "every": 10
  }
}
```

`snapshotDayUTC`/`snapshotHourUTC` set the weekly collection time (default Monday 02:00 UTC). `sleepMs`/`every` throttle the EVM per-wallet loop. Everything is optional except `disabled: false`.

### Testing Notes
On a node with synced data (regtest works fine):

1. Add to `bitcore.config.json` under `bitcoreNode.services`: `"walletStats": { "disabled": false }`
2. Run a one-shot collection: `node build/src/workers/walletStats.js --CHAIN BTC --NETWORK regtest --EXIT true`
[...]

PR #4252, opened 2026-09-17
[BCN] Fix cache expiry race that discards refreshed entries

leolambo:
### Description
`CacheStorage.getGlobal` and `getForWallet` fire an unawaited `remove` keyed on `_id` when they read an expired entry. `setGlobal`'s upsert refreshes the same document in place, so a delete that lands late removes the entry that was just written. Around each expiry a hot key can end up missing on every request, and each miss re-runs the underlying lookup (RPC fee estimates today, plus any other `getGlobalOrRefresh` consumer). Surfaced while reviewing cache usage in the wallet stats work (#4250).

### Changelog

* Await the expiry delete and match on the entry's `created` stamp, so the delete only removes the version that was actually read as expired. A concurrently refreshed entry keeps its `_id` but gets a new `created`, and the stale delete no-ops against it.
* Add Cache Model integration tests pinning the expiry-delete contract.

### Testing Notes
Covered by the new cases in the Cache Model integration suite. Fresh entries and normal expiry behave the same as before.

<hr style="border-width:3px">

### Checklist


packages/bitcore-node/src/models/cache.ts
```diff
@@ -75,7 +75,8 @@ export class CacheModel extends BaseModel<ICache<any>> {
       return found.value as T;
     } else {
       // cache miss
-      this.collection.remove({ _id: found._id });
+      // cache miss - only delete the version seen expiring; a refresh reuses the _id
```
kajoseph, line 78: Rm comment above.

kajoseph (commented):
Nice catch! While we're at it, might as well change the deprecated `.remove` to `.deleteOne`

kajoseph (approved):

PR #4210, opened 2026-08-03
BWS: Transak use forwarded deviceIp as x-user-ip for web-context 

shadi-bp:
### Description
Production web Buy flows for Transak fail with "Invalid or expired session" when the widget opens. Transak's mandatory security changes pin each widget session to the `x-user-ip` it was created with, but web requests are proxied through the bitpay backend, so BWS was sending that server's IP instead of the customer's. Sandbox is unaffected because Transak relaxes IP checks there.

### Changelog

* Add `transakGetUserIp` helper: web context uses the proxy-forwarded `deviceIp` (same pattern as the MoonPay fix in 2f16b1299), app context keeps using the request IP.
* Use it for `x-user-ip` in `transakGetAccessToken` and `transakGetSignedPaymentUrl`.
* Fall back to the request IP when `deviceIp` is not forwarded, so BWS can deploy before the bitpay server change.
* Canonicalize IPv4-mapped IPv6 addresses (`::ffff:1.2.3.4` → `1.2.3.4`).
* Strip `deviceIp` from the `widgetParams` sent to Transak.

### Testing Notes
Run `npx mocha --require ts-node/register test/integration/externalservices/transak.test.ts` in `packages/bitcore-wallet-service` — includes new tests for deviceIp forwarding, canonicalization, non-web behavior, the missing-deviceIp fallback, and widgetParams exclusion. Full production verification requires the paired bitpay server change that forwards `deviceIp`, deployed to a Transak-whitelisted host.

<hr style="border-width:3px">

### Checklist
* [x] I have read [CONTRIBUTING.md](https://github.com/bitpay/bitcore/tree/master/CONTRIBUTING.md) and verified that this PR follows the guidelines and requirements outlined in it.

PR #4253, opened 2026-09-21
Feat: forward customer deviceIp to Simplex for web Buy requests

mahume:
## Description
Since September 2, 2026, no Simplex Buy orders placed through the web widget (`bitpaywidget`) have been approved. Simplex confirmed every `/quote` and `/partner/data` request carries a static `client_ip` / `account_details.signup_login.ip` — BWS's own server IP — regardless of the shopper's real location. Simplex recently added an IP-consistency fraud check that compares this to the requester's real IP and blocks the order on mismatch.

Root cause: web Buy is server-to-server — the shopper's browser talks to bitpay.com, and BitPay's backend (frontend proxy wallet) is what calls BWS. Both `simplexGetQuote` and `simplexPaymentRequest` derive the IP via `Utils.getIpFromReq(req)`, which for the web flow returns BWS's own server IP, not the shopper's.

This is the same bug class already hit and fixed for MoonPay (`4e698ca5b`) and Transak (#4210).

This PR adds a `simplexGetUserIp(req)` helper, mirroring `transakGetUserIp`: prefers `req.body.deviceIp` when `context === 'web'`, falls back to `Utils.getIpFromReq(req)` otherwise (unchanged behavior for the app/native context), and canonicalizes IPv4-mapped IPv6. It replaces the direct `Utils.getIpFromReq(req)` calls in both `simplexGetQuote` (→ `client_ip`) and `simplexPaymentRequest` (→ `account_details.signup_login.ip`), and strips `deviceIp` before the body is forwarded to Simplex.

**Pairs with** the `bitpay/bitpay` change that forwards `deviceIp` (`lib/integrations/cryptoWidget/simplex.js`); safe to deploy in either order since BWS currently ignores unknown body fields — Simplex orders aren't unblocked until both ship.

## Testing notes for reviewer
- Existing `simplexGetQuote`/`simplexPaymentRequest` specs (web and non-web) are unaffected — they don't set `req.body.deviceIp`, so behavior falls through to the existing `Utils.getIpFromReq(req)` path unchanged.
- Added coverage mirroring the Transak `deviceIp` tests: forwarding `deviceIp` for web context, IPv4-mapped IPv6 canonicalization, not leaking `deviceIp` into the outgoing Simplex body, ignoring a stray `deviceIp` for non-web context, and falling back to the request IP for web when `deviceIp` isn't forwarded.

## Issue
Bug: B1-3726


Gamboster (approved):
LGTM

mahume:
Associated bitpay repo bug ticket change is on ship.

kajoseph (approved):

